PR #1, opened 2020-05-11
Define a Source class, but add fetch and release semantics.

jpivarski:
Uproot3's Sources were assumed to be like a memmap, but most users access data remotely. It was sorely missing methods for prefetching byte ranges that can run in parallel with interpretation.

We couldn't just add those to the old version; they'd have to be used consistently. Now we can use them consistently.


jpivarski:
@chrisburr This is the modification to Uproot `Sources` that will make it possible to fetch all the desired baskets in a multipart request. The cost is that it creates a new distinction between `Source` (from which data can be drawn) and `Chunk` (data ready to be read by a `Cursor`). I'll have to work that into every part of Uproot—to start each object by asking for a set of `Chunks` which fill in the background while I interpret them. (The whole library needs to be refactored, but I had to do that for Awkward 1.0 anyway.)

chrisburr:
This is wonderful, thank you very much.

It looks like `uproot4` isn't in a usable enough state yet that I can try benchmarking. Can you ping me once it's possible to read some kind of large ROOT file? Flat TTrees are my default, but I can try something else if that better aligns with your roadmap for uproot4.

For the sake of tying issues together: https://github.com/scikit-hep/uproot/issues/393

eduardo-rodrigues:
Hi @jpivarski, I noticed while looking at the development of uproot4 that you have a copy-and-paste typo "BSD 3-Clause License; see https://github.com/jpivarski/awkward-1.0/blob/master/LICENSE" at the beginning of various files ...

jpivarski:
> It looks like `uproot4` isn't in a usable enough state yet that I can try benchmarking. Can you ping me once it's possible to read some kind of large ROOT file? Flat TTrees are my default, but I can try something else if that better aligns with your roadmap for uproot4.

There's not even anything ROOT-related in here yet. I just wanted to get the physical layer figured out before building on top of it because it implies a difference in how everything else is written. Uproots 1‒3 were written with the idea in mind that I have a memory-mapped file, which means any byte position can be accessed at any time, and trivially in parallel (no state). Then the other interfaces were made to look like memory-mapped files, with bad consequences for performance. One of the things that I didn't fully recognize when I started Uproot was that most users would be accessing remote files most of the time.

It would be possible to do performance tests on randomly chosen ranges of a file. The test would consist of creating a `Source` with some `num_workers`, asking for `chunks` and then iterating through those chunks, asking for `raw_data` from each one (because they're not fully downloaded until you can access the `raw_data`). For realistic byte ranges, you could use Uproot 3 to look at a branch's `_fBasketSeek` (byte positions where its baskets start; only up to `numbaskets` are meaningful) and its `_fBasketBytes` (number of bytes in each basket). This would allow you to create realistic `ranges` (list of `(start, stop)` tuples, where `stop` is `start + num_bytes`).

That would test everything involved in fetching data in a hopefully efficient way; the part it leaves out is interpreting the data, which is subdominant in remote file access over regular networks.

If I'm making a fundamental mistake in these file access methods, then I don't want to design around it. I've been thinking about how I can adequately test that. Do you have a suggestion for a publicly available test file? Would a test through a household ISP mean anything, given that most of these files would be used on university or lab computers or on the GRID?

jpivarski:
> Hi @jpivarski, I noticed while looking at the development of uproot4 that you have a copy-and-paste typo "BSD 3-Clause License; see https://github.com/jpivarski/awkward-1.0/blob/master/LICENSE" at the beginning of various files ...

Good catch! Thanks—I'll take care of that as soon as I can.

In fact, the Awkward ones shouldn't point to `jpivarski` as the owner anymore. It's been a long time since I moved that to `scikit-hep`.

chrisburr:
> If I'm making a fundamental mistake in these file access methods, then I don't want to design around it. I've been thinking about how I can adequately test that. Do you have a suggestion for a publicly available test file?

I default to using these three open data files for XRootD:
```bash
$ xrdfs root://eospublic.cern.ch/ ls -l /eos/opendata/lhcb/AntimatterMatters2017/data
-r-- 2017-03-07 13:53:05   666484974 /eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetDown.root
-r-- 2017-03-07 13:53:08   444723234 /eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetUp.root
-r-- 2017-03-07 13:53:08     2272072 /eos/opendata/lhcb/AntimatterMatters2017/data/PhaseSpaceSimulation.root
```

I've just copied them to CERN S3 so they can be accessed using standard HTTP(S) as well:
* https://scikit-hep-test-data.s3.cern.ch/B2HHH_MagnetDown.root
* https://scikit-hep-test-data.s3.cern.ch/B2HHH_MagnetUp.root
* https://scikit-hep-test-data.s3.cern.ch/PhaseSpaceSimulation.root

> Would a test through a household ISP mean anything, given that most of these files would be used on university or lab computers or on the GRID?

It's definitely useful as the quality of institute connections varies a lot in my experience and it's more likely to expose issues with latency, especially given you'll be accessing them from the US. I think a good implementation should only be limited by bandwidth when reading large enough files.

If you put together a test I can easily try it at a few different institutes as well as pushing a test job to every LHCb grid site.

jpivarski:
Great, I'll do that! I'll make a script that uses three Uproot 4 accessors (HTTPSource, HTTPMultipartSource, and XRootDSource), synchronous and with a variable number of asynchronous workers, to request all of the byte ranges that are associated with TBaskets in these files, though they won't interpret any of them. Knowing how this fares on a home ISP in the U.S. and on LHCb grid sites would make a good test.

Thanks!

PR #3, opened 2020-05-15
This is another way that multipart HTTP(S) can fail, forcing us to fall back on one-request-per-thread.

jpivarski:


jpivarski:
My first example of how a multipart range request can fail (by sending a response code other than 206) came from github.io. This one came from Amazon S3 (through HTTPS). The [S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html) says that it's not supported:

Screenshot from 2020-05-14 20-01-12

but the specific way that it fails is by responding with response code 206 and _not_ sending multipart data; instead it sends _only one part_. So the heuristic for checking is if the actual length matches one of the ranges, we give up. (Note that even a multipart request with only one part would have a length that doesn't match because it has to have the separator and header.)

There are probably other ways that it can fail.

I can't find a lot of documentation on multipart GET. [This guy on StackOverflow](https://stackoverflow.com/q/55665466/1623645) is even asking, "What are they good for?" and gets NO RESPONSE. That's not a good sign.

Issue #2, opened 2020-05-15
File access methods performance studies


jpivarski:
Data provided by @chrisburr; discussion below
------------------------------

> If I'm making a fundamental mistake in these file access methods, then I don't want to design around it. I've been thinking about how I can adequately test that. Do you have a suggestion for a publicly available test file?

I default to using these three open data files for XRootD:
```bash
$ xrdfs root://eospublic.cern.ch/ ls -l /eos/opendata/lhcb/AntimatterMatters2017/data
-r-- 2017-03-07 13:53:05   666484974 /eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetDown.root
-r-- 2017-03-07 13:53:08   444723234 /eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetUp.root
-r-- 2017-03-07 13:53:08     2272072 /eos/opendata/lhcb/AntimatterMatters2017/data/PhaseSpaceSimulation.root
```

I've just copied them to CERN S3 so they can be accessed using standard HTTP(S) as well:
* https://scikit-hep-test-data.s3.cern.ch/B2HHH_MagnetDown.root
* https://scikit-hep-test-data.s3.cern.ch/B2HHH_MagnetUp.root
* https://scikit-hep-test-data.s3.cern.ch/PhaseSpaceSimulation.root

> Would a test through a household ISP mean anything, given that most of these files would be used on university or lab computers or on the GRID?

It's definitely useful as the quality of institute connections varies a lot in my experience and it's more likely to expose issues with latency, especially given you'll be accessing them from the US. I think a good implementation should only be limited by bandwidth when reading large enough files.

If you put together a test I can easily try it at a few different institutes as well as pushing a test job to every LHCb grid site.

_Originally posted by @chrisburr in https://github.com/scikit-hep/uproot4/pull/1#issuecomment-627928856_

jpivarski:
Using original Uproot, I could find basket byte ranges for some interesting branches in B2HHH_MagnetDown.root (the biggest file).

```python
import uproot

file = uproot.open("root://eospublic.cern.ch//eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetDown.root")
tree = file["DecayTree"]

basic = tree["H1_PX"]
uneven = tree["B_FlightDistance"]
manysmall = tree["H1_ProbK"]
small = tree["H1_Charge"]
smaller = tree["H2_isMuon"]

[...]

PR #4, opened 2020-05-15
Add XRootDVectorReadSource

chrisburr:
Adds an XRootD source that uses vector reads. This should probably be the only XRootD source.

By default it takes the number of chunks in a request by querying the server configuration:
```
time_profile(uproot4.source.xrootd.XRootDVectorReadSource(
    "root://eospublic.cern.ch//eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetDown.root"
), basic_ranges)
```
However for now I've exposed this as a parameter, if only to make the performance studies easier:
```
time_profile(uproot4.source.xrootd.XRootDVectorReadSource(
    "root://eospublic.cern.ch//eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetDown.root",
    max_num_elements=5
), basic_ranges)
```

Related to #2.

### DIFF
--- a/tests/test_0001-source-class.py
+++ b/tests/test_0001-source-class.py
@@ -44,6 +44,14 @@ def test_file(tmpdir):
                 b"@@@@@",
             ]
 
+
+def test_file(tmpdir):
+    filename = os.path.join(str(tmpdir), "tmp.raw")
+
+    with open(filename, "wb") as tmp:
+        tmp.write(b"******    ...+++++++!!!!!@@@@@")
+
+    for num_workers in [0, 1, 2]:
         with pytest.raises(Exception):
             uproot4.source.file.FileSource(
                 filename + "-does-not-exist", num_workers=num_workers
@@ -68,6 +76,13 @@ def test_memmap(tmpdir):
             b"@@@@@",
         ]
 
+
+def test_memmap(tmpdir):
+    filename = os.path.join(str(tmpdir), "tmp.raw")
+
+    with open(filename, "wb") as tmp:
+        tmp.write(b"******    ...+++++++!!!!!@@@@@")
+
     with pytest.raises(Exception):
         uproot4.source.file.FileSource(filename + "-does-not-exist")
 
@@ -86,6 +101,8 @@ def test_http():
         chunks = tmp.chunks([(0, 100), (50, 55), (200, 400)])
         assert [x.raw_data.tostring() for x in chunks] == [one, two, three]
 
+
+def test_http():
     source = uproot4.source.http.HTTPMultipartSource(
         "https://wonky.cern/does-not-exist", timeout=0.1
     )
@@ -106,6 +123,9 @@ def test_no_multipart():
             assert len(three) == 200
             assert one[:4] == b"root"
 
+
+def test_no_multipart_fail():
+    for num_workers in [0, 1, 2]:
         source = uproot4.source.http.HTTPSource(
             "https://wonky.cern/does-not-exist", num_workers=num_workers, timeout=0.1
         )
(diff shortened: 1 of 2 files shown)


jpivarski:
This is definitely faster than the multithreaded version: `XRootDSource` typically returned after 200‒500 seconds while `XRootDVectorReadSource` returned after 60‒200 seconds. From the waterfall of times, I think that the multithreaded version is more affected by stragglers. Actually, it's the fact that the multithreaded version is partly synchronized by the fact that the number of workers is much smaller than the number of chunks... The vector_read is qualitatively better because the server gets the whole request up-front...

I un-reversed the order of chunks (`while True: ranges.pop()` → `for start, stop in ranges`). I'm still figuring out how the interpretation code is going to use this—which is my primary intent from these performance tests—but it would be nice if I can expect the outputs to come out in roughly the order of request (which the `time_profile` test code explicitly assumes). Without that assumption, the interpretation of baskets (mostly decompression) would _have_ to be multithreaded and triggered by chunks finishing (which is an interface change from what I have now, so thinking about this is useful!).

Maybe optimization comes in the form of distributing chunk sizes well among the `all_request_ranges`? Would it make sense to distribute `ranges` among the `all_request_ranges` such that the total number of bytes in each set is more uniform? The fact that these B2HHH.root tests all have small baskets followed by large baskets is a reminder that we often see patterns like that (because ROOT's basket-filling algorithm tunes itself as it goes forward).

Do you want to make more changes to this before I commit it? I will probably remove the original `XRootDSource` and maybe rename this one, if vector_read is stable, available everywhere, and doesn't have memory leaking issues.

jpivarski:
I would merge this if you have no other additions. Let me know!

After this PR, there are a few things that will happen:
   * The names will be changed so that this is `XRootDSource` and the other one, if it continues to exist at all, will have a highly specialized name. Unless there are problems with the vector_read version or we need the other one to exist as a fallback, the vector_read version will be the only XRootD source, as you suggest.
   * The `chunks` method (on all Sources) will return a queue of completed chunks so that I don't have to rely on them being finished in order and can start decompressing/interpreting them in the order that they have been read.
   * I will also have to add a simpler interface for single-chunk access, since there are many occasions in which we only want one contiguous block of bytes and we need it synchronously before we can look for the next one (e.g. an item fetched from a TDirectory or all of the streamers). That will probably be `chunk` vs `chunks`.

chrisburr:
I'll reply tomorrow but go ahead and merge if you're ready 👍

jpivarski:
Okay! I figured it was late there and I'd catch you on the next cycle, but thanks!

lukasheinrich:
hi @chrisburr -- how does this handle authentication to an xrootd server?

chrisburr:
@lukasheinrich Primarily environment variables (`XRD_.*`/`X509_.*`) the same way as the command line client/C++ library/ROOT does.

jpivarski:
If pyxrootd works by forking a subprocess, we can perhaps control those environment variables programmatically from Python. (It's my understanding that setting `os.environ` only affects subprocesses.)

chrisburr:
Thanks for the clean up, this looks much better now.

> Maybe optimization comes in the form of distributing chunk sizes well among the all_request_ranges?

I'd guess the optimisation comes from most storage being configured to work better when requesting larger chunks of data to minimise the overhead of each request. This matches the most common operations: copies of the entire file or vector reads from ROOT. This might be especially important considering the file is physically on a different server and might even be split across many servers (e.g. ECHO at RAL).

This also makes me realise I should run my grid tests with the file at different storage elements as there can be a lot of quirks between different setups.

> Would it make sense to distribute ranges among the all_request_ranges such that the total number of bytes in each set is more uniform?

This feels like a premature optimisation, it could also be better to make requests which are spatially closer in the file. I think the requests should be optimised to return data in the most useful groups for uproot, e.g. small chunks of metadata in one request and sequential baskets in another.

> If pyxrootd works by forking a subprocess, we can perhaps control those environment variables programmatically from Python.

I think this is a bad idea. Everywhere I'm familiar with it supposed to "just work" (e.g. lxplus, local kerberos, local grid proxies) or these variables are intended to be set up without users knowing about it (e.g. grid jobs setting `X509_`). Keeping code independent of authentication is a feature that helps make collaborative working and long term preservation easier.

> It's my understanding that setting os.environ only affects subprocesses.

On most systems the changes to `os.environ` affect everything except things that have already done what they were going to with the environment. From the Python docs:

> When putenv() is supported, assignments to items in os.environ are automatically translated into corresponding calls to putenv(); however, calls to putenv() don’t update os.environ, so it is actually preferable to assign to items of os.environ.

PR #13, opened 2020-05-28
Add RNTuple anchor, with test.

jpivarski:


jpivarski:
@oshadura This provides access to RNTuple in Uproot4. The test shows the "right way" to use it:
* context manager for the file (`f` is actually a TDirectory that points back to the TFile through `f.file`)
* get the ROOT::Experimental::RNTuple through `f["Staff"]`
* get its members through `obj.member("fWhatever")`
* read a single chunk of raw bytes (immediately), such as the data corresponding to the RNTuple header, with `f.file.source.chunk`
* put in a request (might run in the background) for multiple chunks with `f.file.source.chunks` (for XRootD, this is a VectorRead; for HTTP, this is a multipart-GET)
* print out bytes for debugging with `cursor.debug(chunk, length)`; shows their decimal values (more useful than hex, I've found), string interpretation, and optionally numerical interpretation (pass a dtype and offset)
* actually interpret the header using `cursor.fields(struct)`, `cursor.string()`, etc.

The `Cursor` is a pointer to some position of the file, the `Source` is an object that delivers data (local file, XRootD, HTTP), and a `Chunk` is a range of bytes taken from the `Source`. `Cursors` only operate on a `Chunk`, not the `Source` (that's a change from old Uproot, allowing for more lazy file access). `Chunks` may be filled in the background (depending on the type of source), but when a `Cursor` asks for something from the `Chunk`, it waits until such data are available.

When you get to the point of reading a bunch of chunks for RNTuple pages, note that you can [pass a notification queue](https://github.com/scikit-hep/uproot4/blob/52cc61a277773d22098090f2e5a2e5a39f684340/tests/test_0006-notify-when-downloaded.py#L196-L211) to `Source.chunks` so that you can initiate page-decompression as soon as the page is ready. I'll be doing the same for TTree baskets, so you may be able to follow my example.

If the data you're pointing to could be compressed, you'll probably want to follow [TKey.get_uncompressed_chunk_cursor](https://github.com/scikit-hep/uproot4/blob/52cc61a277773d22098090f2e5a2e5a39f684340/uproot4/reading.py#L660-L684). If you know that what you have is compressed, but haven't yet read the 9-byte compression header, call [uproot4.compression.decompress](https://github.com/scikit-hep/uproot4/blob/52cc61a277773d22098090f2e5a2e5a39f684340/uproot4/compression.py#L146) to turn a `Chunk` of compressed bytes into a `Chunk` of uncompressed bytes. Note that the returned `Chunk` starts at `Cursor(0)`, rather than the seek point in the file (like the `Cursor` you pass to this function), because there isn't a one-to-one relationship between uncompressed byte positions and compressed byte positions. The `context` is unused but included as a formality; it can be an empty dict. You do need to know the number of `compressed_bytes` and `uncompressed_bytes` so that the output array can be allocated, but I think you know this information from RNTuple's equivalent of `TKey::fNbytes` and `TKey::fObjlen`.

Do you want me to give you write access to the uproot4 repo in case you need to make any upstream changes to get your things to work?

oshadura:
Thanks Jim and thanks for an explanation ! (about access it will nice, but in case if you prefer I can jus t open PR and ping you there)

PR #18, opened 2020-06-02
Fill out the array-fetching interface.

jpivarski:


.github/workflows/build-test.yml
```diff
@@ -50,13 +56,15 @@ jobs:
       - name: "flake8"
         shell: "bash -l {0}"
```
henryiii, line 57: The `shell: "bash -l {0}"` lines can now be dropped (likewise elsewhere), it was only needed for setup-miniconda.

jpivarski: Thanks!

PR #19, opened 2020-06-03
Fixing bugs for HATS.

jpivarski:


henryiii:
You can set fail-fast to false to allow all the jobs to run, rather that quitting as soon as one fails.

henryiii:
Looks like conda may be installed on Linux only.

henryiii:
Conda is pre-installed in windows too, just not macOS: https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md

henryiii:
You'll probably need `shell: bash -l {0}`, and it will *really* slow down your tests to preinstall conda each time. How about just testing conda on linux and windows, and then having a pip-based test elsewhere?

henryiii:
Keep in mind in multiple OS testing, Windows defaults to powershell, you need `shell: bash` to get the bash shell on Windows. (All OS's have powershell-core, so you can go all powershell, too)

henryiii:
`conda init bash` is the wrong thing to use for the setup-miniconda action - that needs the `shell: bash -l {0}` instead. Also conda init bash requires that you restart your shell, IIRC.

jpivarski:
Thanks for the help, @henryiii—I haven't been ignoring you; I just noticed all of these messages where you warned me of just about everything I stumbled over.

henryiii:
Conda is pre-installed on all OS's, including macOS. I'm discussing with GitHub Actions team, will have an update here soon.

jpivarski:
> You can set fail-fast to false to allow all the jobs to run, rather that quitting as soon as one fails.

The fail-fast is new to me in GitHub Actions (if it was possible in Travis or Azure, I didn't know about it), but I like it. Usually, when a commit is wrong, it's wrong in all platforms because it's a coding bug. (GHA's flake8 catches more errors than mine, and I need to figure out how to get that version! Oh, 3.7 vs 3.8... maybe that's it.)

Occasionally, I really do need it to not fail-fast, such as a recent search on scikit-hep/uproot#493 for the `struct` syntax for that platform-dependent `long` type, and I'll turn off fail-fast if I'm in one of those situations.

henryiii:
They've changed the defaults and names, but otherwise, GHA is pretty much identical to Azure. And it was also possible to enable fail-fast in Travis, but don't remember the name; CI systems like fail-fast, after all. :)

GitHub CI will fail on a multiline command if a command in the middle fails, which is *fantastic*, whereas Azure will fail only if the final command fails, which is awful - they said it was too late to change the default in Azure so they only changed it when they developed GHA.

jpivarski:
The Uproot HATS is done, so I'll close off this PR and give it a version number.

PR #20, opened 2020-06-04
[DO NOT MERGE] Adding test of conda envs

henryiii:
Testing built-in conda.


jpivarski:
Should this one be deleted now?

henryiii:
I think so - it's pretty easy to setup if we need more info, but most of the relevant errors have been copied to the GHA issue.

PR #21, opened 2020-06-04
Simplify Conda activate

henryiii:
This is the recommended way to activate conda, and does use the built-in conda automatically (for the last few weeks or months). You can also cache a more complex environment now (not done in this cleanup).

It would probably be better to add a (test) environment.yml for testing instead of manual conda and pip installs.


jpivarski:
I used to have a requirements-test.txt, but I needed XRootD to depend on Python version and platform. I suppose I could make it depend on Python version in the requirements-test.txt. Or maybe introduce requirements-test.txt and requirements-test-extra.txt, where the latter has only XRootD.

PR #23, opened 2020-06-17
Develop more interpretations.

jpivarski:


jpivarski:
All of the old Uproot tests have been ported over except those in

   * test_tree.py
   * test_issues.py

and some of the ones that were ported are skipped.

PR #58, opened 2020-07-23
Detach live, readable file objects from Model instances, replacing them with static DetachedFile objects.

jpivarski:


jpivarski:
To clarify that last claim, it can be written and read back because of [PEP 562](https://www.python.org/dev/peps/pep-0562/), which means Python 3.7+ only.

Since it can't be demonstrated in a unit test, I'll write it here. The writing process does

```python
import uproot4, skhep_testdata, pickle
f = uproot4.open(skhep_testdata.data_path("uproot-hepdata-example.root"))
original = f["hpx"]
pickle.dump(original, open("tmp.pkl", "wb"))
```

and the reading process does

```python
import pickle
reconstituted = pickle.load(open("tmp.pkl", "rb"))
reconstituted.to_numpy()
...
```

The class of `original` is dynamically generated from the TStreamerInfo in the ROOT file. Pickle doesn't like classes that don't exist in a module, but we put it in `uproot4.dynamic` for pickle to find when writing.

When reading, we no longer have the dynamically generated class, but when pickle looks for it in `uproot4.dynamic`, it finds a (new, empty) class object, which it uses to fill the instance. Models have a custom `__getstate__`/`__setstate__` that updates the empty shell `uproot4.dynamic` produced with the `member_names`, `class_code`, `class_streamer`, etc. that it needs, as well as all of the class's behaviors.

So the `reconstituted` is a _work-alike_ of the original object. It lacks the methods needed for it to be read from a file (though they're in the `class_code`, we could re-evaluate that), but the object does not need to be read from a file again.

henryiii:
`__slots__` interfere with pickle?

Also, why can't that be a unit test?

jpivarski:
> Also, why can't that be a unit test?

Which can't be a unit test?

henryiii:
I just saw:
> Since it can't be demonstrated in a unit test

jpivarski:
Writing to a pickle file in one process and reading in another can't be demonstrated in a unit test because unit tests, at least the way pytest does it, are all in a single process. I could save the written file into the GitHub repo and read that as a test, but that wouldn't be a fair test unless I found a ROOT class that no other tests read in. For it to be a fair test, we have to have never seen this class before (and I'm using TH1F for it, which isn't uncommon).

PR #59, opened 2020-07-24
XRootD improvements

chrisburr:
Apologies this PR grew a little as I wrote it, the commits should be self contained and easy to review individually. If you'd rather not have any of them (especially 52aa7f4 and a68a6bd) I'll rebase that commit out of the history.

In summary:
* Adds an extra to `setup.py` so `pip install -e .[testing]` allows the tests to be ran fully
* Makes some minor corrections to the test dependencies and what is used in `importorskip`
* Replaces `pyxrootd` with `XRootD` as `pyxrootd` is more of a private implementation detail that should be prefixed with an underscore
* Avoids a deadlock on exit (https://github.com/scikit-hep/uproot/issues/504) by using an `atexit` handler to cleanup open files (this should be made XRootD version dependent once fixed upstream (https://github.com/xrootd/xrootd/pull/1260)
* Adds a test which leaks an open file to make sure pytest deadlocks if https://github.com/scikit-hep/uproot/issues/504 is an issue
* Improves the XRootD server configuration detection to handle redirectors and locally stored files (https://github.com/scikit-hep/uproot4/issues/57)

Closes https://github.com/scikit-hep/uproot4/issues/57
Closes https://github.com/scikit-hep/uproot/issues/504

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ The full list is
    * `pandas`: only if `library="pd"`.
    * `cupy`: only if `library="cp"` (reads arrays onto GPUs).
    * `dask[array]` and `dask[dataframe]`: experimental, for lazy arrays with `library="da"`.
-   * `pyxrootd`: only if reading files with `root://` URLs.
+   * `xrootd`: only if reading files with `root://` URLs.
    * `lz4` and `xxhash`: only if reading ROOT files that have been LZ4-compressed.
    * `zstandard`: only if reading ROOT files that have been ZSTD-compressed.
    * `backports.lzma`: only if reading ROOT files that have been LZMA-compressed (in Python 2).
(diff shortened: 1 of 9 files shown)


chrisburr:
> It's my understanding that the first XRootD version that includes your PR should not register the atexit handler.

Yes though it doesn't matter if both are registered, it will only have the effect of making the Python process exit slightly more slowly (but not noticeably so). If the fix is backported it might have to be a split check like: `if (version < 5.0.0 and version > 4.12.4) or (version > 5.0.2)`. I'll take care of adding this once we know what versions should be there.

> Ordinarily, I squash and merge PRs (because I think of a PR as a single unit to add to a repo), and even though these commits are nicely curated, I'd like to do the same here

Absolutely! The commits only serve to make this page easier to understand, if anyone is ever tracing a bug back through the git history they should come here to read the comments and associated issues.

nikoladze:
This is great! Not only does it fix both of my issues - also by looking at these commits i learned that i can query the server i'm talking to which is very useful for testing. Thanks @chrisburr!

PR #62, opened 2020-08-03
optimize handling of header bytes

bendavid:
Faster handling of the data masking required when header bytes are present.
Speeds up the reading of large std::vectors especially with multiple cores because the GIL usage is reduced.


bendavid:
No more changes coming here for the moment. (Further significant improvements would require Numba I think.)

Issue #64, opened 2020-08-10
Cursor.skip_over can only be used on an object with non-null `num_bytes`


raymondEhlers:
Using uproot4 v'0.0.16', when trying to read a tree that was copied from another file (may be coincidental - just including all details), I receive the following traceback:

```
>>> import uproot4 as uproot
>>> f = uproot.open("AnalysisResults.root")
>>> f.keys()
['AliAnalysisTask;3']
>>> f["AliAnalysisTask"]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/reading.py", line 1371, in __getitem__
    return self.key(where).get()
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/reading.py", line 870, in get
    out = cls.read(chunk, cursor, context, self._file, self)
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/model.py", line 478, in read
    versioned_cls.read(chunk, cursor, context, file, parent, concrete=concrete),
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/model.py", line 123, in read
    self.read_members(chunk, cursor, context)
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/models/TTree.py", line 665, in read_members
    chunk, cursor, context, self._file, self._concrete
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/model.py", line 123, in read
    self.read_members(chunk, cursor, context)
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/models/TObjArray.py", line 42, in read_members
    chunk, cursor, context, self._file, self._parent
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/deserialization.py", line 280, in read_object_any
    obj = cls.read(chunk, cursor, context, file, parent)
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/model.py", line 478, in read
    versioned_cls.read(chunk, cursor, context, file, parent, concrete=concrete),
  File "/alf/data/rehlers/substructure/.venv/lib/python3.7/site-packages/uproot4/model.py", line 123, in read
    self.read_members(chunk, cursor, context)
[...]

PR #65, opened 2020-08-10
Fixing #64; skip_over should be conditional.

jpivarski:


henryiii:
The problem seems to be "archspec" not being installable on windows, not that there's a problem with the GH Action.  This looks like a conda-forge issue possibly. Try using the action, but removing the conda-forge channel (you can add it manually to the install command)

jpivarski:
> The problem seems to be "archspec" not being installable on windows, not that there's a problem with the GH Action. This looks like a conda-forge issue possibly. Try using the action, but removing the conda-forge channel (you can add it manually to the install command)

Thanks—that was a big help. I just wanted to spot-fix one little issue and then a workflow thing that used to work took up an hour. (I think this is the gritting teeth one: `:E` Or else maybe three-toothed vampire?)

Anyway, your help was invaluable! I was thinking that the "goanpeca/setup-miniconda" marketplace item had just gotten out of date or something and wanted it to work any way at all, even by manual installation (which is what I'd do to a VM).

PR #67, opened 2020-08-10
Add function to retrieve common entry offsets among branches

nsmith-:
This used to be called "clusters" in uproot0 but is not strictly the
ROOT definition of clusters. Users may wish to partition reading
a TTree according to these offsets to prevent having to read and
decompress the same basket in multiple partitions.


nsmith-:
I am adding some tests and found `uproot-hepdata-example.root` has several baskets and the only common offset is the whole file. Now I am looking for another file with well-aligned baskets, any hints?

jpivarski:
foriter.root has some small, oddly shaped baskets. The purpose was to check the correctness of stitching together partial baskets when an iteration size doesn't match it.

jpivarski:
The physics samples like HZZ.root and Zmumu.root might have the typical AutoFlush pattern (small baskets followed by big baskets as the algorithm learns how much memory it has).

nsmith-:
This is ready

PR #71, opened 2020-08-11
Fix a small bug in AsGrouped form

nsmith-:


jpivarski:
Here, I must have been assuming (hoping?) that the RecordForm constructor would be like the RecordArray constructor, making it independent of dict order (Python 2.7 and 3.5-).

But I'll merge it if you're ready.

nsmith-:
ah I see why, ok maybe better to add that kind of form constructor instead?

jpivarski:
If you're willing. I should let you know that it's pybind11 work, not Python, so it will involve some compilation time. I can point out the line where it goes.

jpivarski:
This is RecordForm's only constructor so far: it takes a `std::map<std::string, ak::FormPtr>` as its only required argument.

https://github.com/scikit-hep/awkward-1.0/blob/f491da0adcd04fb6b07eb5616aa146c2ef737e32/src/python/forms.cpp#L555-L575

The `std::map` is presented in Python as dict, which loses order in Pythons 2.7 and 3.5. If there were a second constructor that took two lists (`std::vector<std::string>` and `std::vector<ak::FormPtr>`), then the keys and values could be passed to it in a way that preserves order.

The C++ RecordForm itself takes the list of `ak::FormPtr` (`contents`) separately from the list of keys (`util::RecordLookupPtr recordlookup`):

https://github.com/scikit-hep/awkward-1.0/blob/f491da0adcd04fb6b07eb5616aa146c2ef737e32/include/awkward/array/RecordArray.h#L22-L26

so there wouldn't be anything to do after the pybind11 interface (i.e. not an exploding-in-size project).

jpivarski:
The RecordForm constructor described above has been added and is being published in 0.2.30. If that works, this PR may be unnecessary and should be closed.

jpivarski:
I'm going to close this because I think it should work now. If it doesn't, then you can open it up again.

PR #80, opened 2020-08-26
Simplify file path/object splitter

tamasgal:
This addresses https://github.com/scikit-hep/uproot4/issues/79 which shows that `uproot4` is not able to handle filenames containing colons.

I played around a bit but I could not find a suitable solution without the prior knowledge that a filename ends with `.root` due to the ambiguity of the colon used as object path separator. Therefore this implementation assumes that filename terminates with `.root` and makes the splitting fairly straight forward. Given that this will only fail when a user specifies a file without the `.root` extension **and** also an object path in addition, the collateral damage might be very low.

The risk can be further reduced by first checking for the occurrence of `.root` and adapt the behaviour accordingly.

Another possible solution would be to force the user to use spaces around the object path separator (` : `), but I guess this is quite established already.

Let me know what you think.

### DIFF
--- a/uproot4/_util.py
+++ b/uproot4/_util.py
@@ -144,6 +144,7 @@ def regularize_path(path):
     return path
 
 
+_file_path_object_path_matcher = re.compile(r"^(.+\.root)[ ]*:[ ]*(.+)$")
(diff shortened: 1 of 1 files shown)


jpivarski:
Thanks for the suggestion, but this is now being dealt with in #81.

Issue #90, opened 2020-09-08
Reading large doubly jagged arrays iteratively/lazily?


tamasgal:
Edit: sorry, some copy&paste error destroyed the intro :wink: now fixed

So basically what we need is to process many branches of doubly jagged arrays in files which do not fit into memory.

Here is an example with a small file, showing the structure of such a branch (with uproot3 which falls back to `ObjectArray` and uproot4 which gives a nice awkward1 array):

```
import uproot
import uproot4
from km3net_testdata import data_path
print(uproot.__version__)  # 3.12.0
print(uproot4.__version__)  # 0.0.20

filename = data_path("offline/km3net_offline.root")
f = uproot.open(filename)
f4 = uproot4.open(filename)

f["E/Evt/trks/trks.rec_stages"].array()
# <ObjectArray [[[1, 3, 5, 4], [1, 3, 5], [1, 3], ...

f4["E/Evt/trks/trks.rec_stages"].array()
# <Array [[[1, 3, 5, 4], [1, ... 1], [1], [1]]] type='10 * var * var * int64'>
```

I uploaded a larger file here: http://131.188.167.67:8889/doubly_jagged.root for which `f["E/Evt/trks/trks.rec_stages"].array()` takes extremely long.

So I tried to utilise `uproot4.lazy()` and played around with `uproot4.iterate()` but I have not figured out how to read the data iteratively. Here are my naive approaches (based on some

```python
trks = uproot4.lazy({"doubly_jagged.root": "E/Evt/trks"})
trks[0, "trks.rec_stages"]
```

which yiels:

```
ValueError: generated array does not conform to expected form:

{
    "class": "ListOffsetArray64",
    "offsets": "i64",
    "content": {
        "class": "ListOffsetArray64",
        "offsets": "i64",
        "content": "int32"
    }
}

but generated:

{
    "class": "ListOffsetArray64",
    "offsets": "i64",
    "content": {
        "class": "ListOffsetArray64",
        "offsets": "i64",
        "content": "int64"
    }
}

(https://github.com/scikit-hep/awkward-1.0/blob/0.2.33/src/libawkward/virtual/ArrayGenerator.cpp#L46)
```

I am not sure why it confuses between `int32` and `int64`, maybe due to empty arrays? No clue...

ping @zinebaly

jpivarski:
[...]

PR #96, opened 2020-09-10
Set up interface between Uproot and Awkward so that Awkward can be used to optimize object-reading.

jpivarski:
As discussed in issue #90.

Corresponds to scikit-hep/awkward-1.0#448.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,3 +57,4 @@ The full list is
    * `backports.lzma`: only if reading ROOT files that have been LZMA-compressed (in Python 2).
    * `boost-histogram`: only if converting histograms to Boost with `.to_boost()`.
    * `hist`: only if converting histograms to hist with `.to_hist()`.
+
--- a/uproot4/behaviors/TBranch.py
+++ b/uproot4/behaviors/TBranch.py
@@ -645,10 +645,15 @@ def lazy(
             names = []
             for key in common_keys:
                 branch = obj[key]
+
                 interpretation = branchid_interpretation[branch.cache_key]
                 form = interpretation.awkward_form(obj.file, index_format="i64")
                 if isinstance(interpretation, uproot4.interpretation.objects.AsObjects):
-                    form = uproot4._util.awkward_form_of_iter(awkward1, form)
+                    if not uproot4.interpretation.objects.awkward_can_optimize(
+                        interpretation, form
+                    ):
+                        form = uproot4._util.awkward_form_of_iter(awkward1, form)
+
                 generator = awkward1.layout.ArrayGenerator(
                     branch.array,
                     (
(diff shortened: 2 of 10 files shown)


jpivarski:
@tamasgal The good news is that, with this PR and scikit-hep/awkward-1.0#448, the following code takes 3.6 seconds instead of 194 seconds (53× faster):

```python
import uproot4
branch = uproot4.open("issue-90.root:E/Evt/trks/trks.fitinf")
for i in range(branch.num_baskets):
    print(repr(branch.basket(i).array()))
```

In the above, we're reading each TBasket individually. The bad news is that it still takes a long time to concatenate TBaskets because [ak.concatenate](https://awkward-array.readthedocs.io/en/latest/_auto/ak.concatenate.html) has only been (internally) implemented for pairs: concatenating _n_ arrays means creating and throwing away _n - 2_ temporary arrays of quadratically increasing size. That was fine for examples of concatenating 2 or 3 arrays at a time, but when it comes to concatenating data from this file's 461 TBaskets, it's a problem.

Clearly, [ak.concatenate](https://awkward-array.readthedocs.io/en/latest/_auto/ak.concatenate.html) needs to be fixed anyway. There might already be an issue open about it. (It's been on my mind for a while...) Anyway, I'll tackle that next. We want the conclusion of the above story to be that the whole array is produced in 3.7 seconds!

tamasgal:
Awesome! Many thanks, Jim, that's really a huge leap.

jpivarski:
For cases that aren't covered by the new interpret-by-Awkward mechanism, how about parallelizing the pure Python interpretation?

   * **parallel interpretation in 'basket_array':** 259 seconds. 12 cores utilized at about 5-10% each. I'll blame the GIL. Reversing the order (putting the big TBaskets first) doesn't help—it's not about stragglers.
   * **sequential interpretation in 'basket_array', concatenate NumPy dtype=O arrays:** 190 seconds.
   * **sequential interpretation in 'final_array' (old way):** 197 seconds. Moving the interpretation from `final_array` into `basket_array` doesn't hurt. Entries that would be trimmed (because they're at the ends of the first and last basket) are now unnecessarily interpreted, but that's probably better than not having the possibility to parallelize.
   * **interpret-by-Awkward in 'basket_array':** 2.5 seconds.

So the baseline of 197 seconds wasn't unnecessarily harsh. It really is an 80× speedup.

jpivarski:
Making use of this will require Awkward 0.2.37 (but it won't _break_ with earlier versions), which is being deployed now. I'll do one last test after that deployment so that GHA pulls the new Awkward from PyPI.

tamasgal:
Thanks, that's really nice and helps a lot!

PR #97, opened 2020-09-10
Update TH* behaviors to match uproot3 expectations

### LINKED ISSUE #93: Values read from TH2* are mixed up
It appears that something is going wrong with reading TH2* histograms compared to ROOT or uproot3. When reading a TH2, the second axis values seem to be reversed, as well as the over and underflow next to each other. I can reproduce it (see below), and it seems like it should be straightforward, but somehow I haven't yet found the source.

```python
from pathlib import Path
filename = Path("reproducer.root")
if not filename.exists():
    import ROOT
    # Fill to create an identity, but with different number of bins to ensure that we don't accidentally get it right.
    h = ROOT.TH2D("test", "test", 10, 0, 10, 12, 0, 12)
    for i in range(0, 10):
        h.Fill(i + 0.5, i + 0.5)

    f_out = ROOT.TFile("reproducer.root", "RECREATE")
    h.Write()
    f_out.Close()

import uproot4
import uproot as uproot3
import numpy as np

hist_3 = uproot3.open("reproducer.root")[b"test"]
hist_4 = uproot4.open("reproducer.root")["test"]

# Ignore overflow for now since uproot3.values ignores overflow
np.testing.assert_allclose(hist_3.values, hist_4.values()[1:-1, 1:-1])
```

this gives:

```
AssertionError:
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 19 / 120 (15.8%)
Max absolute difference: 1.
Max relative difference: 1.
 x: array([[1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0.],...
 y: array([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0.],...
```

Just printing all of the values gives some info. Expected from uproot3 (note: excluding flow bins):

```python
>>> print(hist_3.values)
[[1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0.]]
```

Uproot4 (note: including flow bins and the reverse):

```python
>>> print(hist_4.values()[:, ::-1])
[[1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]]
```

Any help is appreciated!

raymondEhlers:
Fixes #93


raymondEhlers:
When testing locally, I'm seeing that `test_numpy_2d` from `test_0046-histograms-bh-hist.py` fails. However, when I create my test histogram in boost_histogram (based on my reproducer from #93) and compare it, my changes here agree with it. Uproot3 does too (as expected - I crosschecked the code from there). So I'm not yet sure what's going on with that

jpivarski:
The histogram bin values in the test were derived from a sample execution of the function, so when you fixed their order, it revealed the test was wrong. In this case, the test wasn't pinning the code to truth, but to an earlier state, so that we'd catch any unexpected changes.

raymondEhlers:
Ah, okay, I was wondering if that was the case! (hence the PR now rather than further investigation at the moment) Thanks!

jpivarski:
Whenever you feel that it's done, let me know and I'll merge it. It looks good to me as it is.

(I replaced `array.T` with `numpy.transpose(array)` because it makes the code easier to maintain. Other than that, I have no comments or requested changes.)

raymondEhlers:
I think it's fine to merge now. There are probably other possible improvements here, but I tested a few variations with my reproducer and they seem to work okay, so it should be good enough for now. Thanks!

jpivarski:
Thank you!

PR #113, opened 2020-09-19
ci: cleanup a bit

henryiii:
Cleaning up a little. setup-minconda has moved. GHA has added a new feature that allows the default shell to be set.

You still have way too many quotes on things that don't need them. :P

(But you are okay to put the contents of `run:` in without quotes?...)

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -21,8 +21,6 @@ jobs:
             python-version: "3.6"
           - platform: "windows-latest"
             python-version: "3.7"
-        # platform: ["windows-latest"]
-        # python-version: ["3.5", "3.6", "3.7", "3.8"]
 
     runs-on: "${{ matrix.platform }}"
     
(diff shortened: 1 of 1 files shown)


jpivarski:
Yeah, because the rules for YAML string escapes are complicated and I got burned by one of them. (I don't remember what it was, but there are things like `yes` meaning `true` unless it's quoted.) I wanted to make sure that everything was being interpreted as I thought it was because I was having troubles.

Multiline YAML strings are a safer encoding—only a deindent closes the quote.

But thanks for updating the setup! Do you foresee any other changes or can I merge it like this?

henryiii:
Not all your run's are multiline, though. Generally ": " is the most common issue for an unquoted string. `uses`, branches, urls, etc are always safe. Totally up to you, though :)

Nope, as long as it passed (it did), it's safe to merge. If you wanted to move a little of the 

```bash
conda config --add channels conda-forge
conda update -n base -c defaults conda
```

up into the [action options](https://github.com/conda-incubator/setup-miniconda/blob/master/action.yml):

```yaml
channels: conda-forge
```

Then you can just `conda list` / `conda info` to get some logging information, but you shouldn't need further configuration, just install lines. Conda is set to update itself whenever it installs packages.

henryiii:
Applied the minor cleanup I mentioned above, seems to be fine.

jpivarski:
Okay, thanks. I'll merge it now.

PR #127, opened 2020-10-19
feat: show similar keys first on KeyInFile error

### LINKED ISSUE #119: KeyInFileError to show near matches?
`KeyInFileError` currently shows a truncating "Known keys" list. This cool regarding the general format of what to ask for instead, but since we quite commonly have files with a large number of branches one still needs to go back to some kind of `[key for key in f.keys() if 'bla' in key]` to find the correct name.

It would be really cool if this would instead show closest matches first. I guess this could get computationally pricey, i.e. slow, so maybe it would only get computed if the number of keys was not too crazy?

andrzejnovak:
Fixes #119. 

Seems to work pretty well.
image

### DIFF
--- a/uproot4/_util.py
+++ b/uproot4/_util.py
@@ -635,3 +635,48 @@ def awkward_form_of_iter(awkward1, form):
         )
     else:
         raise RuntimeError("unrecognized form: {0}".format(type(form)))
+
+
+def damerau_levenshtein(a, b, ratio=False):
+    # Modified Damerau-Levenshtein distance. Adds a middling penalty
+    # for capitalization.
+    # https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance
+    M = [[0] * (len(b) + 1) for i in range(len(a) + 1)]
+
+    for i in range(len(a) + 1):
+        M[i][0] = i
+    for j in range(len(b) + 1):
+        M[0][j] = j
+
+    for i in range(1, len(a) + 1):
+        for j in range(1, len(b) + 1):
+            if a[i - 1] == b[j - 1]:  # Same char
+                cost = 0
+            elif a[i - 1].lower() == b[j - 1].lower():  # Same if lowered
+                cost = 0.5
+            else:  # Different char
+                cost = 2
+            M[i][j] = min(
+                M[i - 1][j] + 1,  # Addition
+                M[i][j - 1] + 1,  # Removal
+                M[i - 1][j - 1] + cost,  # Substitution
+            )
+
+            # Transposition
+            if (
+                i > 1
+                and j > 1
+                and a[i - 1].lower() == b[j - 2].lower()
+                and a[i - 2].lower() == b[j - 2].lower()
+            ):
+                if a[i - 1] == b[j - 2] and a[i - 2] == b[j - 1]:
+                    # Transpose only
+                    M[i][j] = min(M[i][j], M[i - 2][j - 2] + 1)
+                else:
+                    # Traspose and capitalization
+                    M[i][j] = min(M[i][j], M[i - 2][j - 2] + 1.5)
+
+    if not ratio:
+        return M[len(a)][len(b)]
+    else:
+        return (len(a) + len(b)) - M[len(a)][len(b)] / (len(a) + len(b))
--- a/uproot4/exceptions.py
+++ b/uproot4/exceptions.py
@@ -1,0 +1,76 @@
+from uproot4._util import damerau_levenshtein
+
+
+class KeyInFileError(KeyError):
+    """
+    Exception raised by attempts to find ROOT objects in ``TDirectories``
+    or ``TBranches`` in :py:class:`~uproot4.behaviors.TBranch.HasBranches`, which
+    both have a Python ``Mapping`` interface (square bracket syntax to extract
+    items).
+
+    This exception descends from Python's ``KeyError``, so it can be used in
+    the normal way by interfaces that expect a missing item in a ``Mapping``
+    to raise ``KeyError``, but it provides more information, depending on
+    availability:
+
+    * ``because``: an explanatory message
+    * ``cycle``: the ROOT cycle number requested, if any
+    * ``keys``: a list or partial list of keys that *are* in the object, in case
+      of misspelling
+    * ``file_path``: a path (or URL) to the file
+    * ``object_path``: a path to the object within the ROOT file.
+    """
+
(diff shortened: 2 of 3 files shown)


uproot4/exceptions.py
```diff
@@ -0,0 +1,80 @@
...
+        else:
+            because = " because " + self.because
+
+        with_keys = ""
+        if self.keys is not None:
+            to_show = None
+            distance = []
+            for key in self.keys:
+                distance.append(damerau_levenshtein(self.key, key))
+            self.keys = np.array(self.keys)[np.argsort(distance)]
```
jpivarski, line 48: An information-seeking method like `__str__` should not change the state of an object (assign to `self.keys`). I can't imagine any specific errors that this is likely to cause, but it's always something you don't think about. It's good to minimize mutability as much as possible, leaving everything but a few well-advertised attributes immutable, especially for something we like to think of an an inert object, such as an exception instance. You could either solve this by sorting in every `__str__` (performance is not very important here) or by moving the sorting to the constructor.

Also, be aware that this leaves `keys` as a NumPy array, where a Python list would be expected—the contents are strings and are likely to be a small set of them. Things like that end up being "gotchas."

A simple way to solve this would be to do the sort with a [list sorting key](https://docs.python.org/3/howto/sorting.html) instead of a NumPy-like argsort. Actually, since `damerau_levenshtein` takes two arguments, it would make more sense to use a [cmp parameter](https://docs.python.org/3/howto/sorting.html#the-old-way-using-the-cmp-parameter), which is not preferred for some reason, but a `cmp` function can always be turned into a `key` function.

uproot4/_util.py
```diff
@@ -635,3 +635,56 @@ def awkward_form_of_iter(awkward1, form):
...
+    a : str
+    b : str
+    ratio : bool, optional, default False
+        Return ratio (len(a)+len(b) - DLH)/(len(a)+len(b))
+
+    Returns
+    -------
+    float
+        Modified Damerau Levenshtein distance
+    """
```
jpivarski, line 655: Docstrings are not required for hidden functions like this. I also think it's a little dangerous, since their implementations could change and wrong documentation is worse than no documentation (though people disagree on this point). Anyway, this is written in [NumPy style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) and we use [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

Since it's an internal function, I think a comment would be as good as a docstring, since people should only be using these if they are looking at the code, anyway, so that's up to you. Perhaps the most useful commentary would be the URL where the algorithm is described in detail—that's the sort of thing we'd need if we have to revisit this.

uproot4/_util.py
```diff
@@ -635,3 +635,56 @@ def awkward_form_of_iter(awkward1, form):
...
+            if a[i - 1] == b[j - 1]:  # Same char
+                cost = 0
+            elif a[i - 1].lower() == b[j - 1].lower():  # Same if lowered
+                cost = 0.5
+            else:  # Different char
+                cost = 2
+            M[i][j] = min(M[i - 1][j] + 1,  # Addition
+                          M[i][j - 1] + 1,  # Removal
+                          M[i - 1][j - 1] + cost  # Substitution
+                          )
```
jpivarski, line 674: Be sure to run `black` on this. I don't enforce it in the CI because I think automatically changing contributor's code is more confusing than helpful, and the automatically-run `flake8` usually catches formatting issues, at least enough to be a good reminder to run `black`.

jpivarski (commented):
Nice! Just a few minor comments:

uproot4/_util.py
andrzejnovak: Ah cool, I didn't realize you switched to black

jpivarski:
Let me know when you're done (no more changes expected), and if everything's in good shape, I'll merge it right away. Thanks!

andrzejnovak:
@jpivarski should be it

uproot4/exceptions.py
```diff
@@ -0,0 +1,79 @@
...
+        else:
+            because = " because " + self.because
+
+        with_keys = ""
+        if self.keys is not None:
+            to_show = None
+            distance = []
+            for key in self.keys:
+                distance.append(damerau_levenshtein(self.key, key))
+            sorted_keys = [x for _, x in sorted(zip(distance, self.keys), key=lambda pair: pair[0])]
```
jpivarski, line 47: Isn't it possible to replace the four lines above with

```python
sorted_keys = sorted(self.keys, key=lambda x: damerau_levenshtein(self.key, x))
```

I haven't explicitly tried it, but it's supposed to work something like that—much simpler (and therefore easier to maintain) than what you have. I might have misled you when I was talking about `cmp` vs `key`; I hadn't noticed that the relevant `key` function here has only one parameter because `self.key` (the bad one that the user tried) is a constant across all `self.keys` (the good alternatives; any one of which is what they meant). Only one argument of `damerau_levenshtein` varies.

I don't see any other issues (now that I'm looking closely).

Oh, I don't know if there are any tests that sample this because it's an exception. I'll take your word for it if, after the above fix, you get the right results in a KeyInFileError. Thanks!

jpivarski (changes requested):
Sorry, after looking closely to the changes in the sorting part, I have one requested change.

uproot4/exceptions.py
andrzejnovak: You're right. That works.

andrzejnovak: In principle, there could be a test to capture the error and check that it's always the same for some test file, but that sounds a bit extreme to me.

jpivarski (approved):

uproot4/exceptions.py
jpivarski: A test that checks the string representation of a key error would be extremely brittle. One little change in formatting would cause such a test to fail, and that would be a "false positive" in the sense that nothing actually went wrong, and yet we have to fix tests. A good test suite is as much about what you _don't_ test as what you do test.

Anyway, this looks good. Thanks!

jpivarski:
I'm using this now on a MiniAOD file, which has 2299 branches, and it takes a human-perceptible amount of time, like a few seconds. Fortunately, it shows the "`raise uproot.KeyInFileError(`" text while it computes all those Damerau-Levenshtein distances, so at least I know why it's pausing.

(And searching through those 2299 branches by hand to find my misspelling would be much worse! So thanks again for this!)

PR #171, opened 2020-10-31
Quickfix .lazy() when called on missing files. Fixes #120

### LINKED ISSUE #120: Calling lazy() on a filename that doesn't exist produces confusing error
This is what happens if you call `lazy()` on a path that doesn't exist:

```python
>>> import uproot4
>>> uproot4.lazy('nonexistent_file.root')
  File "/home/mproffit/miniconda3/lib/python3.8/site-packages/uproot4/behaviors/TBranch.py", line 585, in lazy
    raise ValueError(
ValueError: allow_missing=True and no TTrees found in

    
```

Obviously this doesn't really indicate the actual problem--that the file doesn't exist. It also strangely claims that `allow_missing=True`, which is not correct since the default is `False`. It says the same thing even if you explicitly set `allow_missing=False`:

```python
>>> uproot4.lazy('nonexistent_file.root', allow_missing=False)
  File "/home/mproffit/miniconda3/lib/python3.8/site-packages/uproot4/behaviors/TBranch.py", line 585, in lazy
    raise ValueError(
ValueError: allow_missing=True and no TTrees found in

    
```

tamasgal:
The exception from `uproot4._util._file_not_found` was simply not raised.

### DIFF
--- a/tests/test_0045-lazy-arrays-1.py
+++ b/tests/test_0045-lazy-arrays-1.py
@@ -148,3 +148,10 @@ def test_awkward_pluralization():
     )
     array = uproot4.lazy({files: "sample"})
     assert awkward1.to_list(array[:5, "i4"]) == [-15, -14, -13, -12, -11]
+
+
+def test_lazy_called_on_nonexistent_file():
+    filename = "nonexistent_file.root"
+    with pytest.raises(FileNotFoundError) as excinfo:
+        uproot4.lazy(filename)
+    assert filename in str(excinfo.value)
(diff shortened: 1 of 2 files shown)


jpivarski:
Wow, this looks like a silly mistake. Does it fix #120?

tamasgal:
Oh, I forgot to mention that, yes of course. That was the main intention, to fix #120 😆

tamasgal:
Actually I put it in the branch name, at least ;)

jpivarski:
**Great, thanks for finding this!** I must have forgotten that this (internal) function returns an exception, rather than raising it. I think the reason for that was for the stack trace to end in a relevant spot, rather than the relevant spot being second-to-last in the stack trace.

tamasgal:
I see! : )

Issue #121, opened 2020-10-02
HTTP protocol doesn't follow redirect


nsmith-:
I cannot open 
```python
uproot4.open("https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root")
```
yet I can open the place it `302` redirects to:
```python
uproot4.open("https://raw.githubusercontent.com/CoffeaTeam/coffea/master/tests/samples/nano_dy.root")
```

jpivarski:
Adding my comment from Slack, for the record:

This is a missed case:

https://github.com/scikit-hep/uproot4/blob/d8a4b7203952bb7bd85cbf6d2ed481d610c83e67/uproot4/source/http.py#L163-L174

doesn't check for r`esponse.status == 302` (redirect) and use the connection to try again at `response.headers["location"]`.

Same for:

https://github.com/scikit-hep/uproot4/blob/d8a4b7203952bb7bd85cbf6d2ed481d610c83e67/uproot4/source/http.py#L244

which only returns True or False based on whether the status is 206 or not. This should be an issue asking to support redirects.

On the second issue of not being able to access the redirected file: I wasn't able to the first time, then I was every subsequent time. I verify that I can read the data in it. I don't know if this was a one-time GitHub thing or if it always needs to "warm up," or what.

maxnoe:
When you are at it, please don't forget the other redirect status codes:

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_redirection

PR #177, opened 2020-11-03
Fix multipart HTTP reading

### LINKED ISSUE #176: Cannot open ROOT over HTTPS
I cannot open a ROOT file over HTTPS with uproot4 while it works with uproot

```python
import uproot4
_ = uproot4.open('https://cern.ch/starterkit/data/advanced-python-2019/dalitzdata.root')
```
It fails with
```
OSError: remote server does not support HTTP range requests
for URL https://cern.ch/starterkit/data/advanced-python-2019/dalitzdata.root
```

Do you have any ideas what may goes wrong?

Full StackTrace:
```
OSError                                   Traceback (most recent call last)
<ipython-input-3-2fc8249c41b5> in <module>
----> 1 _ = uproot4.open('https://cern.ch/starterkit/data/advanced-python-2019/dalitzdata.root')

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/reading.py in open(path, object_cache, array_cache, custom_classes, **options)
    134         )
    135 
--> 136     file = ReadOnlyFile(
    137         file_path,
    138         object_cache=object_cache,

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/reading.py in __init__(self, file_path, object_cache, array_cache, custom_classes, **options)
    565             self._fUUID_version,
    566             self._fUUID,
--> 567         ) = uproot4.source.cursor.Cursor(0).fields(
    568             self._begin_chunk, _file_header_fields_small, {}
    569         )

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/source/cursor.py in fields(self, chunk, format, context, move)
    193         if move:
    194             self._index = stop
--> 195         return format.unpack(chunk.get(start, stop, self, context))
    196 
    197     def field(self, chunk, format, context, move=True):

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/source/chunk.py in get(self, start, stop, cursor, context)
    360         already.
    361         """
--> 362         self.wait()
    363 
    364         if (start, stop) in self:

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/source/chunk.py in wait(self)
    314         """
    315         if self._raw_data is None:
--> 316             self._raw_data = numpy.frombuffer(self._future.result(), dtype=self._dtype)
    317             if len(self._raw_data) != self._stop - self._start:
    318                 raise OSError(

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/source/futures.py in result(self, timeout)
    124             return self._result
    125         else:
--> 126             delayed_raise(*self._excinfo)
    127 
    128     def _run(self):

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/source/futures.py in delayed_raise(exception_class, exception_value, traceback)
     44         exec("raise exception_class, exception_value, traceback")
     45     else:
---> 46         raise exception_value.with_traceback(traceback)
     47 
     48 

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/source/futures.py in _run(self, resource)
    281     def _run(self, resource):
    282         try:
--> 283             self._result = self._task(resource)
    284         except Exception:
    285             self._excinfo = sys.exc_info()

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/source/http.py in task(resource)
    199 
    200         def task(resource):
--> 201             return resource.get(connection, start, stop)
    202 
    203         return uproot4.source.futures.ResourceFuture(task)

~/anaconda3/envs/starterkit38/lib/python3.8/site-packages/uproot4/source/http.py in get(self, connection, start, stop)
    167         if response.status != 206:
    168             connection.close()
--> 169             raise OSError(
    170                 """remote server does not support HTTP range requests
    171 for URL {0}""".format(

OSError: remote server does not support HTTP range requests
for URL https://cern.ch/starterkit/data/advanced-python-2019/dalitzdata.root

```

chrisburr:
Fixes #176 with dd75037.

I started looking at #121 but it becomes messy with the current architecture. I think the fundamental problem is that `uproot4/source/http.py` uses too low level an interface (`http.client`). The [Python docs also suggest this](https://docs.python.org/3/library/http.client.html):

> It is normally not used directly — the module `urllib.request` uses it to handle URLs that use HTTP and HTTPS.
> **See also:** The Requests package is recommended for a higher-level HTTP client interface. 

For now I've left an expected-to-fail test for #121 (2fce1d1) and slightly improved HTTP error code handling to make failures easier to understand (be91e45) but I can rebase these out if you'd rather not have these changes.

### DIFF
--- a/uproot4/source/http.py
+++ b/uproot4/source/http.py
@@ -284,8 +284,12 @@ for URL {0}""".format(
 
         return uproot4.source.futures.ResourceFuture(task)
 
-    _content_type = re.compile(r"^multipart/byteranges; boundary=([^;=]+)$", re.IGNORECASE)
-    _content_range_size = re.compile(b"Content-Range: bytes ([0-9]+-[0-9]+)/([0-9]+)", re.IGNORECASE)
+    _content_type = re.compile(
+        r"^multipart/byteranges; boundary=([^;=]+)$", re.IGNORECASE
+    )
+    _content_range_size = re.compile(
+        b"Content-Range: bytes ([0-9]+-[0-9]+)/([0-9]+)", re.IGNORECASE
+    )
     _content_range = re.compile(b"Content-Range: bytes ([0-9]+-[0-9]+)", re.IGNORECASE)
 
     def is_multipart_supported(self, ranges, response):
@@ -333,7 +337,9 @@ for URL {0}""".format(
             multipart_boundary = None
 
         for i in uproot4._util.range(len(futures)):
-            range_string, size, multipart_boundary = self.next_header(response, multipart_boundary)
+            range_string, size, multipart_boundary = self.next_header(
+                response, multipart_boundary
+            )
             if range_string is None:
                 raise OSError(
                     """found {0} of {1} expected headers in HTTP multipart
@@ -388,12 +394,13 @@ for URL {3}""".format(
                 elif len(line.strip()) == 0:
                     continue
                 elif line.startswith(b"--"):
-                    multipart_boundary = line[len(b"--"):].strip().decode()
+                    multipart_boundary = line[len(b"--") :].strip().decode()
                     break
 
                 raise NotImplementedError(
-                    "Failed to parse Content-Type header {}"
-                    .format(response.getheader("Content-Type"))
+                    "Failed to parse Content-Type header {}".format(
+                        response.getheader("Content-Type")
+                    )
                 )
         else:
             line = response.fp.readline()
--- a/uproot4/source/http.py
+++ b/uproot4/source/http.py
@@ -315,15 +315,14 @@ for URL {0}""".format(
         to handle the multipart GET response.
         """
         match = self._content_type.match(response.getheader("Content-Type"))
-        if not match:
-            raise NotImplementedError(
-                "Failed to parse Content-Type header {}"
-                .format(response.getheader("Content-Type"))
-            )
-        multipart_boundary = match.groups()[0]
+        if match:
+            multipart_boundary = match.groups()[0]
+        else:
+            # Infer it from the first response
+            multipart_boundary = None
 
         for i in uproot4._util.range(len(futures)):
-            range_string, size = self.next_header(response, multipart_boundary)
+            range_string, size, multipart_boundary = self.next_header(response, multipart_boundary)
             if range_string is None:
                 raise OSError(
                     """found {0} of {1} expected headers in HTTP multipart
@@ -368,11 +367,29 @@ for URL {3}""".format(
         Helper function for :py:meth:`~uproot4.source.http.HTTPResource.multifuture`
         to return the next header from the ``response``.
         """
-        line = response.fp.readline()
-        while line.strip() != "--{}".format(multipart_boundary).encode():
+        if multipart_boundary is None:
+            # This seems to be purely for testing against example.com where the
+            # response appears to be invalid
+            while True:
+                line = response.fp.readline()
+                if len(line) == 0:
+                    pass
+                elif len(line.strip()) == 0:
+                    continue
+                elif line.startswith(b"--"):
+                    multipart_boundary = line[len(b"--"):].strip().decode()
+                    break
+
+                raise NotImplementedError(
+                    "Failed to parse Content-Type header {}"
+                    .format(response.getheader("Content-Type"))
+                )
+        else:
             line = response.fp.readline()
-            if not line:
-                raise NotImplementedError("Failed to find multipart boundary")
+            while line.strip() != "--{}".format(multipart_boundary).encode():
+                line = response.fp.readline()
+                if not line:
+                    raise NotImplementedError("Failed to find multipart boundary")
 
         line = response.fp.readline()
         range_string, size = None, None
@@ -389,7 +406,7 @@ for URL {3}""".format(
             line = response.fp.readline()
             if len(line.strip()) == 0:
                 break
-        return range_string, size
+        return range_string, size, multipart_boundary
 
     @staticmethod
     def partfuture(results, start, stop):
--- a/uproot4/source/http.py
+++ b/uproot4/source/http.py
@@ -71,6 +71,38 @@ def make_connection(parsed_url, timeout):
         )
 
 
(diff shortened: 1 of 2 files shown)


chrisburr:
So far as I can tell https://example.com returns an invalid response:
```bash
$ curl -L http://www.example.com -i -H "Range: bytes=0-10, 15-20"
HTTP/1.1 206 Partial Content
Accept-Ranges: bytes
Age: 419053
Cache-Control: max-age=604800
Content-Type: text/html; charset=UTF-8
Date: Tue, 03 Nov 2020 10:00:47 GMT
Etag: "3147526947"
Expires: Tue, 10 Nov 2020 10:00:47 GMT
Last-Modified: Thu, 17 Oct 2019 07:18:26 GMT
Server: ECS (nyb/1D04)
Vary: Accept-Encoding
X-Cache: HIT
Content-Length: 225


--3d6b6a416f9b5
Content-Type: text/html; charset=UTF-8
Content-Range: bytes 0-10/1256

<!doctype h
--3d6b6a416f9b5
Content-Type: text/html; charset=UTF-8
Content-Range: bytes 15-20/1256


<html
--3d6b6a416f9b5--
```

as the `Content-Type` header [should be](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#Multipart_ranges) `Content-Type: multipart/byteranges; boundary=3d6b6a416f9b5`. I guess there could be other servers which do this so I've added a workaround in a8beb25.

jpivarski:
For #121, I was just going to add HTTP response numbers. I know that you brought up this point about the requests library being higher level, but I wanted to find out "what we're buying" with the dependency. Redirects is one thing, which doesn't seem so bad, but if you're willing to jump in and help like this and `http.client` is a terrible working environment, I can be persuaded to swallow the dependency just to be more friendly to contributors like you.

I can't believe I missed the message about `urllib.request`; having a higher-level interface already in the standard library is a no-brainer. That comes with zero dependency cost.

I wouldn't try to evolve this toward a completely different level of interface. HTTPSource is a small enough class that it can be rewritten to fit a new interface. I've already written it (and XRootDSource) from scratch twice: once at the start of Uproot 4, to establish a remote-file interface that minimizes round trips (to write the rest of Uproot around), and a second time after I identified a subset of the remote-file interface that I actually used (also, to clean up some spaghetti).

I started a wiki where I'll document the [physical layer interface](https://github.com/scikit-hep/uproot4/wiki/Physical-layer-interface); what Uproot expects of a class like HTTPSource or XRootDSource. This document will be useful in the long run, to add additional interfaces in the future (Ceph? S3?), but if you want to make a clean break and write HTTPSource using `urllib.request` or the third-party requests library, I'll support that. I'll try to get that document written in the next few hours, in either case.

@chrisburr, what do you think? If this is more than you want to get into right now, I'll be patching up #121 by handling all the redirect codes, but the possibility of redoing HTTPSource as a requests-based class is always available. (The requests library would become an optional dependency for PyPI, with a loader in [extras.py](https://github.com/scikit-hep/uproot4/blob/master/uproot4/extras.py), but a strict dependency in conda-forge, like awkward1.)

jpivarski:
I've finished the above-mentioned documention at https://github.com/scikit-hep/uproot4/wiki/Physical-layer-interface

In the meantime, I'll mark #121 as "in progress" (in the [November projects](https://github.com/scikit-hep/uproot4/projects/6)), and I'll wait for your response about how you want to do this. If you want to rewrite HTTPSource with `urllib.request` or the requests library, I'll support that. If not, I'll take over and just add the redirect codes.

jpivarski:
@chrisburr Let me know what you want to do with this: (1) merge in as-is, (2) revamp HTTPSource as requests-based (either `urllib.requests` or the third-party `requests`), or (3) neither: I would address #121 by just adding more redirect codes.

uproot4/source/http.py
```diff
@@ -71,32 +71,51 @@ def make_connection(parsed_url, timeout):
...
+    elif str(response.status)[0] != "2":
         connection.close()
         raise OSError(
-            """HTTP response was {0}, rather than 200, in attempt to get file size
-in file {1}""".format(
-                response.status, file_path
-            )
+            "HTTP response was {}, rather than 2xx, ".format(response.status)
+            + "in attempt to get file size in file {}".format(location)
         )
```
jpivarski, line 101: I'll be sure to apply this policy universally: all the HTTP codes in the 200's are good, all others are bad. But I'll be doing it by checking the integer against the 200's range, not by checking that the first character in the string is "2". I don't know if it's possible to get an error code like 20 or 2000, but it seems like better practice to protect against that.

jpivarski:
@chrisburr Thanks for looking into this. I decided to finish it off, but after resolving a merge conflict, the tests failed. The helper function that checks for redirects is a good idea in principle, but I couldn't follow what was going wrong. (It wasn't just the helper function; the propagation of the "boundary" string also complicated things—it might have been residual merge errors.)

Starting from the main branch, I copied your tests and manually reimplemented what you developed here into PR #201. I didn't do a helper function because the calling structure for what to do in case of a redirect is different in the places where it appears. (Also, I put in a fix for redirects.)

Here are the main things that I gleaned from your PR:

   * Headers, including those that separate parts of a multipart response, are case-insensitive. That's what was breaking the LHCb StarterKit example.
   * Although the "`Content-Type: multipart/byteranges`" is a part of the specification, and it provides a potentially useful string that could be used to cross-check the multipart boundaries, it isn't strictly necessary. We know exactly how many bytes are in each part and this information is reiterated in the "`Content-Range:`" (which _is_ strictly necessary, since the parts might come in a different order—we need to know which one we're on). I've left the parsing of the multiparts minimal.
   * Any response in the 300's is a redirect. As for the 200's, I'm still going to require exact status codes.

Meanwhile, it has been [failing intermittently because of XRootD](https://github.com/scikit-hep/uproot4/runs/1433794743?check_suite_focus=true), though that's a separate thing. I noticed that XRootD went through a major version update (4 → 5). I'll keep an eye on that.

PR #183, opened 2020-11-06
Pass through HTTP(S) query parameter if they exist.

jpivarski:


jpivarski:
@BenGalewsky and @gordonwatts I don't have an example that will test this.

It was hard enough finding an example that I could use to verify that I understood the HTTPSConnection interface correctly—I used this:

```python
connection.request("GET", "/search/code?q=addClass+in:file+language:js+repo:jquery/jquery", headers={"user-agent": "Python"})
```

to be sure that the query is added as part of the second argument, rather than being another argument (without the question mark, encoded differently, etc.). I also don't see anywhere else in the codebase where `ParseResult.path` is used in a context other than local files. If XRootD URLs take query parameters (I doubt it), I didn't fix that, either.

Could you do a simple test against this branch to confirm that you can access the server that gives you ROOT files that depend on query parameters? (Or just reply here with some sample URLs, if they're publicly available and I can test them. Furthermore, if they can be guaranteed to exist, I could add them to the tests.)

gordonwatts:
Ah, thank you so much for tackling this so quickly! We'll see if we can run a test.

BenGalewsky:
It works!
Thank you

chrisburr:
> If XRootD URLs take query parameters (I doubt it), I didn't fix that, either.

They do and reading from at least one site fails in weird ways without them.

jpivarski:
Apparently, [uproot4/source/xrootd.py](https://github.com/scikit-hep/uproot4/blob/master/uproot4/source/xrootd.py) uses the whole `file_path` without decomposing it into a `parsed_url`, so I guess the query parameters go along for the ride.

PR #189, opened 2020-11-09
fix: warning about GHA security

henryiii:
We can instead use dependabot.yml to keep actions up via PRs every day or week when new changes are detected. In that case, we'd use v2.0.0 here.

This was released about 30 minutes ago, see https://github.com/conda-incubator/setup-miniconda/issues/84


jpivarski:
Thanks!

PR #196, opened 2020-11-19
Use the common histogram interface.

jpivarski:
Generalizing https://github.com/henryiii/uhi/blob/master/uhi/typing/plottable.py .

### DIFF
--- a/tests/test_0046-histograms-bh-hist.py
+++ b/tests/test_0046-histograms-bh-hist.py
@@ -2183,7 +2183,7 @@ def test_numpy_profile():
     with uproot4.open(skhep_testdata.data_path("uproot-hepdata-example.root")) as f:
         obj = f["hprof"]
 
-        assert obj.to_numpy(flow=True, errors=True)[1].tolist() == [
+        assert obj.axis().edges(flow=True).tolist() == [
             -numpy.inf,
             -4.0,
             -3.92,
@@ -2289,7 +2289,7 @@ def test_numpy_profile():
             numpy.inf,
         ]
 
-        assert obj.to_numpy(flow=True, errors=True)[0][0].tolist() == [
+        assert obj.values_errors(flow=True)[0].tolist() == [
             17.99833583831787,
             17.05295467376709,
             16.96826426188151,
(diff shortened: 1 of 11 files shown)


jpivarski:
@henryiii This is intended to generalize the histogram protocol.

These are tests, demonstrating everything with the histograms I have in scikit-hep-testdata:

https://github.com/scikit-hep/uproot4/blob/jpivarski/use-the-common-histogram-interface/tests/test_0167-use-the-common-histogram-interface.py

The TAxis behaviors are defined here, implementing this protocol and the Sequence protocol (because I think that was desired):

https://github.com/scikit-hep/uproot4/blob/jpivarski/use-the-common-histogram-interface/uproot4/behaviors/TAxis.py

Histograms are defined in files for each class, but an abstract class is defined here in the file for TH1, as a way of getting docstrings to all the others. (Uproot 4 does not follow ROOT's class hierarchy, giving us more freedom in behavior implementations, like not having TH2 inherit from TH1, etc.)

https://github.com/scikit-hep/uproot4/blob/jpivarski/use-the-common-histogram-interface/uproot4/behaviors/TH1.py

This abstract class is mostly overwritten by the one for profiles, since the docstrings have to be very different.

https://github.com/scikit-hep/uproot4/blob/jpivarski/use-the-common-histogram-interface/uproot4/behaviors/TProfile.py

There's a proliferation of variance methods: `errors`, `variances`, `values_errors`, and `values_variances`. This is because regular histograms use `values` as part of the calculation for `variances`, which can be used to derive `errors`, and profiles use `values` as part of the calculation of `errors`, which can be used to derive `variances`. The methods that return two arrays, values and something else, are primary; the others are conveniences derived from them.

There are two **FIXMEs** in the the profile implementation, one for filling `counts` (I'm unclear about what it means) and the other for building a Boost histogram representing the profile. I'm 100% certain of the calculation from ROOT's C++ members to values and variances, but I'm unsure how to fill Boost. To start with, shouldn't a profile be "WeightedMeanVariance Storage," not "WeightedMean Storage"?

PR #199, opened 2020-11-20
Add fallback for wrong class version.

jpivarski:


jpivarski:
The reason a test was removed is because it relied on the previous brokenness of its instance versions, which would have been corrected by reading its streamers. Now, though, we switch to the right version preemptively, without relying on a deserialization error to kick us back to the streamers. There are now fewer cases that read streamers, but the test was hard-coded to expect it to have read streamers. I don't know of a case to replace it with.

PR #202, opened 2020-11-24
Start writing the Uproot 3 to 4 cheat sheet.

jpivarski:


jpivarski:
Have to stop for now. Will finish this after the name change.

PR #207, opened 2020-11-30
feat: support TProfile 1D conversion

henryiii:
Needs tests, 2D and 3D TProfile.

Using internal DType info because https://github.com/scikit-hep/boost-histogram/pull/475 isn't ready yet.


jpivarski:
Thanks, @henryiii! In general, I try to find data samples "in the wild," which means that it may be some time before we have an instance of a TProfile2D or TProfile3D. When this does come up, we'll have examples to generalize.

PR #219, opened 2020-12-05
Catch up to the latest histogram protocol.

jpivarski:
The only change needed to catch up to

https://github.com/henryiii/uhi/blob/9fcfdc6158c686e58dfd31728b0aef4bd9572abe/uhi/typing/plottable.py

was to rename `interpretation` as `kind` and make its output all upper-case.

@henryiii @HDembinski, this should be ready for the non-release-candidate release, which I'll be starting now (first Awkward Array, then Uproot; should be about half an hour).

### DIFF
--- a/uproot/behaviors/TH1.py
+++ b/uproot/behaviors/TH1.py
@@ -100,12 +100,12 @@ class Histogram(object):
         raise NotImplementedError(repr(self))
 
     @property
-    def interpretation(self):
+    def kind(self):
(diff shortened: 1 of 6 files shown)


HDembinski:
@jpivarski I think it is fine. We are currently only debating whether an ordinary unweighted histogram should something .variances() or None, which does not really affect the interface, only how it should behave in the various cases.

PR #226, opened 2020-12-11
Fix warnings associated with NumPy deprecations.

douglasdavis:
I'm playing with cutting edge NumPy and I got some warnings in unit tests from deprecations triggered in uproot. [NumPy v1.20 deprecates](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated) using `numpy.builtin` over `builtin` (e.g. `numpy.object` instead of `object`). The one change in this PR that I'm not sure about is `numpy.long --> numpy.compat.long`. This is necessary if uproot4 is supporting Python 2, but if Python 3 is required the change can be `numpy.long --> int`.

### DIFF
--- a/uproot/_util.py
+++ b/uproot/_util.py
@@ -47,7 +47,7 @@ def isint(x):
     including bool).
     """
     return isinstance(x, (int, numbers.Integral, numpy.integer)) and not isinstance(
-        x, (numpy.bool, numpy.bool_)
+        x, (bool, numpy.bool_)
     )
 
 
@@ -57,7 +57,7 @@ def isnum(x):
     including bool).
     """
     return isinstance(x, (int, float, numbers.Real, numpy.number)) and not isinstance(
-        x, (numpy.bool, numpy.bool_)
+        x, (bool, numpy.bool_)
     )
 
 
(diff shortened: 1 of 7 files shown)


jpivarski:
Nice! I never knew whether to use `object` or `numpy.object` (or `"O"`): this makes it clear, and I'll do the same in Awkward Array. It also clarifies the distinction between `numpy.bool` and `numpy.bool_`.

Although Uproot doesn't formally target Python 2.7, I occasionally verify that it's not broken. The main reason for not targeting 2.7 is the difficulty of getting the testing framework to work. This policy of not officially supporting it, but making sure that it nevertheless works is a balance between helping people who have constrained, legacy systems and not being obliged to do so. Therefore, I'd rather replace `numpy.long` → `numpy.compat.long`. It's easy and it makes the intention more clear, anyway.

jpivarski:
Oh, let me know when you're done and I'll merge it. Thanks!

douglasdavis:
Jim Pivarski <notifications@github.com> writes:

> This policy of not officially supporting it, but making sure that it
> nevertheless works is a balance between helping people who have
> constrained, legacy systems and not being obliged to do so.

Sounds good!

> Oh, let me know when you're done and I'll merge it. Thanks!

I did a quick search of the repository for the deprecations listed in
the NumPy release notes and everything I could find has been fixed. If I
missed any perhaps they'll surface when some more folks start testing
with the upcoming NumPy release :)

cheers,
Doug

jpivarski:
Great—then it sounds like you're done. I'll merge it!

(The merging step requires a "lock" because it's a concurrency problem—I don't know if you're still working on your local branch and haven't pushed changes. That's why I ask!)

PR #231, opened 2020-12-17
refactor: modernize setup

henryiii:
This is an attempt to modernize the setup. This _might_ fix the issue with https://github.com/conda-forge/uproot-feedstock/pull/73 - and is generally much cleaner and nicer being declarative.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ classifiers =
 
 
 [options]
-python_requires = >=2.7, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
+python_requires = >=2.6, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
 packages = find:
 install_requires =
     numpy >=1.13.3
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ classifiers =
 python_requires = >=2.6, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
 packages = find:
 install_requires =
-    numpy >=1.13.3
+    numpy
 
 [options.packages.find]
 exclude =
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,52 @@
+[metadata]
+name = uproot
+author = Jim Pivarski
+author_email = pivarski@princeton.edu
+maintainer = Jim Pivarski
+maintainer_email = pivarski@princeton.edu
+description = ROOT I/O in pure Python and NumPy.
+long_description = file: README.md
+long_description_content_type = text/markdown
+url = https://github.com/scikit-hep/uproot4
+download_url = https://github.com/scikit-hep/uproot4/releases
+license = BSD 3-clause
+
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Intended Audience :: Developers
+    Intended Audience :: Information Technology
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: BSD License
+    Operating System :: MacOS
+    Operating System :: POSIX
+    Operating System :: Unix
+    Programming Language :: Python
+    Programming Language :: Python :: 2.6
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Topic :: Scientific/Engineering
+    Topic :: Scientific/Engineering :: Information Analysis
+    Topic :: Scientific/Engineering :: Mathematics
+    Topic :: Scientific/Engineering :: Physics
+    Topic :: Software Development
+    Topic :: Utilities
+
+
+[options]
+python_requires = >=2.7, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
+packages = find:
+install_requires =
+    numpy >=1.13.3
+
+[options.packages.find]
+exclude =
+    tests
+
+
 [flake8]
 ignore = E203, W503, E501, E266, N
 max-complexity = 100
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,6 @@
 
 import os.path
 from setuptools import setup
-from setuptools import find_packages
 
 
 def get_version():
@@ -17,55 +16,7 @@ extras = {
 }
 extras["all"] = sum(extras.values(), [])
 
-install_requires = open("requirements.txt").read().strip().split("\n")
-
-setup(name = "uproot",
-      packages = find_packages(exclude = ["tests"]),
-      scripts = [],
-      version = get_version(),
-      author = "Jim Pivarski",
-      author_email = "pivarski@princeton.edu",
-      maintainer = "Jim Pivarski",
-      maintainer_email = "pivarski@princeton.edu",
-      description = "ROOT I/O in pure Python and NumPy.",
-      long_description = open("README.md").read(),
-      long_description_content_type = "text/markdown",
-      url = "https://github.com/scikit-hep/uproot4",
-      download_url = "https://github.com/scikit-hep/uproot4/releases",
-      license = "BSD 3-clause",
-      test_suite = "tests",
-      python_requires = ">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
-      install_requires = ["numpy"],
-      tests_require = extras["test"],
-      extras_require = extras,
-
-      classifiers = [
-#         "Development Status :: 1 - Planning",
-#         "Development Status :: 2 - Pre-Alpha",
-#         "Development Status :: 3 - Alpha",
-#         "Development Status :: 4 - Beta",
-          "Development Status :: 5 - Production/Stable",
-#         "Development Status :: 6 - Mature",
-#         "Development Status :: 7 - Inactive",
-          "Intended Audience :: Developers",
-          "Intended Audience :: Information Technology",
-          "Intended Audience :: Science/Research",
-          "License :: OSI Approved :: BSD License",
-          "Operating System :: MacOS",
-          "Operating System :: POSIX",
-          "Operating System :: Unix",
-          "Programming Language :: Python",
-          "Programming Language :: Python :: 2.6",
-          "Programming Language :: Python :: 2.7",
-          "Programming Language :: Python :: 3.5",
-          "Programming Language :: Python :: 3.6",
-          "Programming Language :: Python :: 3.7",
-          "Programming Language :: Python :: 3.8",
-          "Programming Language :: Python :: 3.9",
-          "Topic :: Scientific/Engineering",
-          "Topic :: Scientific/Engineering :: Information Analysis",
-          "Topic :: Scientific/Engineering :: Mathematics",
-          "Topic :: Scientific/Engineering :: Physics",
-          "Topic :: Software Development",
-          "Topic :: Utilities",
-          ])
+setup(
+    version = get_version(),
+    extras_require = extras,
+)
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,8 @@ from setuptools import setup
 
 def get_version():
     g = {}
-    exec(open(os.path.join("uproot", "version.py")).read(), g)
+    with open(os.path.join("uproot", "version.py")) as f:
+        exec(f.read(), g)
     return g["__version__"]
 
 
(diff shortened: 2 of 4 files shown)


henryiii:
I don't think you build in your tests, but test against an uninstalled uproot. So this isn't really tested. I also can't directly patch the feedstock to apply this, so will try a version applied against 4.0.0 tomorrow.

jpivarski:
I've been thinking about setup.cfg vs setup.py for a while now. Uproot is a plain Python package without strange features, so there are no issues with this declarative approach, and it's much better for metadata.

I'm not sure what you mean about it "not being tested." It does test out of the local directory, not a pip-installed version, which verifies that the code does what it's supposed to do. Do you mean that we're not testing that all the files we intend to include in the package are included in the package? If Uproot had any `data_files`, then I can see how that would be an issue.

One thing about changes to Uproot's build procedure: I don't want to put all the source code into a `src` directory because that interferes with incremental development. Changing setup.py to setup.cfg is great, changing the CI tests to test a pip-installed version is fine, and you'll notice that I'm using black and flake8 now (not automatically changing the code, but flake8 is the gate-keeper and it usually doesn't pass if black hasn't been run). However, I want to keep the directory structure as `uproot`, rather than `src`, so that development/debugging has no intervening steps.

Thanks for your help, though! I appreciate it.

henryiii:
The problem with not having src is that it makes it hard to even test against an installed version, as it picks up the local directory when you run pytest on your tests. If it ever makes a difference - such as if the packaging is broken, then you can't tell in your tests. But I totally get it, I work on several packages this way, other builds systems like poetry and flit are structured this way, so would not push for a `/src` change unless someone really wanted it (or if you have binary components, maybe I'd push for it in that case). I believe it is a better structure due to the issue above, and it pushes you to always develop with virtual environments and `pip install -e .` rather than relying on directory based imports.

You can still keep the logic / funny stuff in setup.py for packages with strange features - it helps highlight the parts that are normal by pulling them out into setup.cfg, and then setup.py is _just_ the special stuff with logic. I left the version and extras there; the version extraction code in setuptools is a little picky on older versions, and special extras like "all" are really a bit nicer if you don't have to be so repetitive.

For really simple packages, I'm actually beginning to be tempted to remove setup.py and only have setup.cfg - people keep thinking `python setup.py ...` is valid API to put into linux repository packaging and such, and it's not, it's an implementation detail, and doesn't activate all the correct machinery for building/installing in some cases. Removing setup.py would teach them a lesson. :)

henryiii:
I mean that if I totally broke `pip install uproot` I could not tell that from the tests. I prefer to consider that all packages to require installing, and the fact Python happens to import directories in the working directory as packages (except for Windows + Python 3.8+, which will not import a DLL this way anymore) to be a trick that might be okay to use locally, but not in tests or production. 99% of your users get your package through pip - that should be the "correct" way to get it even from GitHub/local, and therefore the one tested. Tools like `install_requires` are _really_ handy to be able to depend on!

henryiii:
I'm going to target v4.0.0 for a while, ignore that, I'm just trying to build a patch for conda-forge, and will come back to main once it starts working. That will test the install, because conda does a pip install. :)

henryiii:
Nevermind, GitHub says I can target a tag, but I can't. Admittedly, I have no idea what a "merge" would have been, but it would have been useful for the .patch file. I'll do it locally then.

setup.cfg
```diff
@@ -1,3 +1,52 @@
...
+    Development Status :: 5 - Production/Stable
+    Intended Audience :: Developers
+    Intended Audience :: Information Technology
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: BSD License
+    Operating System :: MacOS
+    Operating System :: POSIX
+    Operating System :: Unix
+    Programming Language :: Python
+    Programming Language :: Python :: 2.6
```
henryiii, line 24: The `python_requires` rules 2.6 out, do we want to list it here? If you ask for the versions, it does show up: PyPI - Python Version (or, alternatively, we could change `python_requires`?)

setup.py
```diff
@@ -17,55 +16,7 @@ def get_version():
...
-      description = "ROOT I/O in pure Python and NumPy.",
-      long_description = open("README.md").read(),
-      long_description_content_type = "text/markdown",
-      url = "https://github.com/scikit-hep/uproot4",
-      download_url = "https://github.com/scikit-hep/uproot4/releases",
-      license = "BSD 3-clause",
-      test_suite = "tests",
-      python_requires = ">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
-      install_requires = ["numpy"],
-      tests_require = extras["test"],
```
henryiii, line 39: I've dropped these - `test_suite` and `tests_require` have been deprecated for a long time. `python setup.py test` requires `setup.py` and is an implementation detail, so using it to run tests was one of the first things to start getting the ax.

setup.py
```diff
@@ -17,55 +16,7 @@ def get_version():
 }
 extras["all"] = sum(extras.values(), [])
 
-install_requires = open("requirements.txt").read().strip().split("\n")
```
henryiii, line 20: I really hate one-line open's, open should always be with `with` - and CPython will spit warnings about this if you enable all warnings. Nice to get rid of this one - but there's still one above in get_version.

setup.py
```diff
@@ -17,55 +16,7 @@ def get_version():
...
-          "Topic :: Scientific/Engineering",
-          "Topic :: Scientific/Engineering :: Information Analysis",
-          "Topic :: Scientific/Engineering :: Mathematics",
-          "Topic :: Scientific/Engineering :: Physics",
-          "Topic :: Software Development",
-          "Topic :: Utilities",
-          ])
+setup(
+    version = get_version(),
+    extras_require = extras,
```
henryiii, line 21: This is pretty easy to move to setup.cfg, but many packages leave it here, so that you can combine extras like "all" and "dev" from existing extras.

setup.py
```diff
@@ -17,55 +16,7 @@ def get_version():
...
-          "Programming Language :: Python :: 3.9",
-          "Topic :: Scientific/Engineering",
-          "Topic :: Scientific/Engineering :: Information Analysis",
-          "Topic :: Scientific/Engineering :: Mathematics",
-          "Topic :: Scientific/Engineering :: Physics",
-          "Topic :: Software Development",
-          "Topic :: Utilities",
-          ])
+setup(
+    version = get_version(),
```
henryiii, line 20: Up to you if you want to try to move this to the cfg. Getting the version is a bit tricky - newer setuptools will try to use the AST, but before then, you had to import or put it in a text (data) file. On my own packages, I use setuptools_scm instead most of the time. Shouldn't be a problem to leave it here.

setup.cfg
```diff
@@ -1,3 +1,52 @@
...
+    Topic :: Scientific/Engineering :: Physics
+    Topic :: Software Development
+    Topic :: Utilities
+
+
+[options]
+python_requires = >=2.7, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
+packages = find:
+install_requires =
+    numpy >=1.13.3
```
henryiii, line 43: Added a min version here, but if it's not true, feel free to remove.

henryiii:
Yes! This fixed the encoding error in conda-forge. :)

Added a few notes on things that changed slightly or might be things you might be interested in changing.

jpivarski:
It really does work for Python 2.6. Of course, I couldn't promise that—it had to go under stealth mode... But okay, fine. 2.7.  `:)`

It's waaaay more important that it works in conda!

henryiii:
I'm quite okay with it working in 2.6 - I just think these two numbers should match. `>=2.6` is fine. :)

Though if we say it works in 2.6, it should have a test. I'm pretty sure GHA provides even 3.0 (?!).

henryiii:
And the `python_requires` is the much more important number - it's the one that stops pip from installing if it doesn't match, and causes it to look for an older version. Though IIRC there might not be a version of pip that supports both 2.6 and `python_requires`, but still, on principle, it should match the trove classifiers!

Just tell me what you want me to set it to or change it. :)

henryiii:
> It's waaaay more important that it works in conda!

Actually, I believe pip install uproot is broken on all Windows, not just conda. Your tests never pip install so you don't catch it, even though you test on Windows. Will open a quick PR adding the pip install to see.

jpivarski:
> I'm quite okay with it working in 2.6 - I just think these two numbers should match. `>=2.6` is fine. :)
> Though if we say it works in 2.6, it should have a test. I'm pretty sure GHA provides even 3.0 (?!).

That's why I didn't want to _say_ it, but let it work if it's going to work for somebody. I even shied away from testing 2.7 because I couldn't get it to work in GHA.

> Actually, I believe pip install uproot is broken on all Windows, not just conda.

If so, then you got me. I don't know how that could have happened, though.

setup.py
```diff
@@ -17,55 +16,7 @@ def get_version():
...
-      maintainer = "Jim Pivarski",
-      maintainer_email = "pivarski@princeton.edu",
-      description = "ROOT I/O in pure Python and NumPy.",
-      long_description = open("README.md").read(),
-      long_description_content_type = "text/markdown",
-      url = "https://github.com/scikit-hep/uproot4",
-      download_url = "https://github.com/scikit-hep/uproot4/releases",
-      license = "BSD 3-clause",
-      test_suite = "tests",
-      python_requires = ">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
```
henryiii, line 37: Wow, I totally read this as 2.7! I must have copied and pasted from the scikit-hep developer pages and have been looking at that! So yours was consistent and correct. I thought they were mismatched.

setup.cfg
```diff
@@ -1,3 +1,52 @@
...
+    Topic :: Scientific/Engineering
+    Topic :: Scientific/Engineering :: Information Analysis
+    Topic :: Scientific/Engineering :: Mathematics
+    Topic :: Scientific/Engineering :: Physics
+    Topic :: Software Development
+    Topic :: Utilities
+
+
+[options]
+python_requires = >=2.7, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
```
henryiii, line 40:
```suggestion
python_requires = >=2.6, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
```

henryiii:
> If so, then you got me.

I added pip install to the tests in #233 and Windows breaks, as promised. Adding the same commit to this PR passes.

> I don't know how that could have happened, though.

Anything untested must be assumed to be broken. You never tested pip installs, so at some point you added a character to the readme that broke on the default codec on Windows, breaking your simple `open()`. The declarative config and `file: README.md` does the open correctly and assumes UTF-8, safely reading the readme.

jpivarski:
Python is platform-independent... except for Windows. To think that reading text from a file without specifying a code breaks it—okay, so your point is well-taken that if it's not doing a `pip install .`, it's not being tested.

I'm having trouble following all of the changes, but I see that it is passing now, and it's doing a `pip install .`. The minimum NumPy would be much lower than 1.13.1, which is significant for people who get NumPy from the system distribution.

About old Pythons, including 2.7, I wanted to only promise that it works for 3.5+ (as it says on the README), but not prevent it from being installed on arbitrarily old systems (where "arbitrarily old" is 2.6).

I'll have to follow up later; a meeting is starting now.

setup.cfg
henryiii:
```suggestion
    numpy
```

setup.py
henryiii: I've gone ahead and removed it too.

setup.cfg
jpivarski: The python_requires now only constrains 2.6, so we can leave this.

setup.py
jpivarski: Thanks.

setup.py
jpivarski: Since we run `pytest` externally to any setup, this no longer has the semantic meaning that it once had. So sure, we can drop it.

setup.py
jpivarski: Let's leave it as one of the last things in setup.py. setuptools_scm can be a later improvement.

jpivarski (approved):
I've checked over all of the changes. I'm in favor! When you give me the thumbs-up, I'll merge it.

henryiii:
Sorry about the 2.6 / 2.7 confusion! I was totally confused because I changed it to 2.7 without realizing it, then was unhappy about the mismatch. 🤦

And, as requested, here it is: 👍

PR #233, opened 2020-12-18
tests: pip install

henryiii:
See #231. I'm cherry picking the addition there to verify it is working.


henryiii:
Yes, this is why I'm such a pill about doing things the proper way and pip installing, etc. Pip install being broken should never pass tests!

henryiii:
Closing this because it's contained in #231, and was opened to show the current situation was broken without the fix contained there.

PR #237, opened 2021-01-07
Hide ReferenceError in XRootD atexit.

jpivarski:
From @ryuwd on https://gitter.im/Scikit-HEP/uproot:

> I am using `uproot4.iterate` to read files over xrootd - I am currently seeing this error at exit after the script has completed all its tasks:
> 
> ```
>     ERROR    | Error in atexit._run_exitfuncs:
>     ERROR    | Traceback (most recent call last):
>     ERROR    |   File "/XicpStToXicpPiPi/.snakemake/conda/3c5bbdd0/lib/python3.8/site-packages/uproot4/extras.py", line 112, in cleanup_open_files
>     ERROR    | if isinstance(obj, XRootD.client.file.File) and obj.is_open():
>     ERROR    | ReferenceError
>     ERROR    | :
>     ERROR    | weakly-referenced object no longer exists
>     ERROR    | Error in atexit._run_exitfuncs:
> ```

My response:

> Uproot has one `atexit` handler, which runs when the script is done, and apparently XRootD has weak references that are no longer valid at that time. You're not forgetting to do something (requiring users to check off formalities is not "Pythonic"). However, there isn't much to do about this error except ignore it. This PR does that: scikit-hep/uproot4#237 Could you check to see if it fixes your error message? (I don't have anything that reproduces it.)

I should point out that this PR does a little more than just ignore the ReferenceError: it ignores it on a per-file basis, so that the exception in one file doesn't cause another file to not get closed. (The `try/except/else` is inside the `for` loop, not outside.)

### DIFF
--- a/uproot/extras.py
+++ b/uproot/extras.py
@@ -110,8 +110,13 @@ def XRootD_client():
         import gc
 
         for obj in gc.get_objects():
-            if isinstance(obj, XRootD.client.file.File) and obj.is_open():
-                obj.close()
+            try:
+                isopen = isinstance(obj, XRootD.client.file.File) and obj.is_open()
+            except ReferenceError:
+                pass
+            else:
+                if isopen:
+                    obj.close()
 
     return XRootD.client
 


ryuwd (approved):
Can confirm this patch removes all of the tracebacks! 

In fixing those, I have noticed another similar-looking traceback in the `XRootD` module:
```
    ERROR    | Error in atexit._run_exitfuncs:
    ERROR    | Traceback (most recent call last):
    ERROR    |   File "/XicpStToXicpPiPi/.snakemake/conda/3c5bbdd0/lib/python3.8/site-packages/XRootD/client/finalize.py", line 45, in finalize
    ERROR    | if isinstance(obj, File) and obj.is_open():
    ERROR    | ReferenceError
    ERROR    | :
    ERROR    | weakly-referenced object no longer exists
```
N.B. this was also present before I applied this patch, and anyway this code is not within `uproot4`! Thanks again.

jpivarski:
Yeah—all this PR does is ignore XRootD's error.

I've used weak references once (so that Awkward Array's virtual array caches don't have circular references through C++, which Python's garbage collector can't detect), and I have to say, they're really hard to get right. Also, `atexit` is a strange time in the Python process when a lot of things you expect to be present are not. It's not surprising that they're having trouble with it, but if they need it (because they go through C++, like Awkward Array does), then they'll have to work it out.

The good news is that it happens after all of your code is done, so it's something of a cosmetic problem, unless the failure to close some files prevents other files from being closed, and that somehow has consequences beyond the lifetime of the process.

PR #239, opened 2021-01-08
Fix TH3's 'to_numpy' (https://stackoverflow.com/q/65632670/1623645).

jpivarski:
As soon as @giammi56/[giammi56](https://stackoverflow.com/users/14010009/giammi56) confirms that this fixes the above-mentioned StackOverflow issue, I'll merge this PR. (And if the tests pass.)

Thanks!


jpivarski:
I haven't heard back, but to be sure this doesn't get forgotten, I'm going to merge it now. (It works and is definitely an improvement.)

PR #240, opened 2021-01-08
feat: Add TGraphAsymmErrors

kratsg:
here's a test script to work with

```python
import uproot
import os.path

fname = "HEPData-ins1755298-v3-Expected_limit_1lbb.root"

if not os.path.isfile(fname):
    with open(fname, 'wb') as f:
        import requests
        response = requests.get("https://www.hepdata.net/download/table/ins1755298/Expected%20limit%201lbb/3/root")
        response.raise_for_status()
        f.write(response.content)

f = uproot.open(fname)
graph = f['Expected limit 1lbb/Graph1D_y1']
graph.values()
graph.errors()
```

### DIFF
--- a/uproot/behaviors/TGraphAsymmErrors.py
+++ b/uproot/behaviors/TGraphAsymmErrors.py
@@ -8,9 +8,8 @@ from __future__ import absolute_import
 
 import numpy
 
-import uproot
-
 # dict_keys(['@fUniqueID', '@fBits', 'fName', 'fTitle', 'fLineColor', 'fLineStyle', 'fLineWidth', 'fFillColor', 'fFillStyle', 'fMarkerColor', 'fMarkerStyle', 'fMarkerSize', 'fNpoints', 'fX', 'fY', 'fFunctions', 'fHistogram', 'fMinimum', 'fMaximum', 'fEXlow', 'fEXhigh', 'fEYlow', 'fEYhigh'])
+
 
 class TGraphAsymmErrors(object):
     """
@@ -41,7 +40,9 @@ class TGraphAsymmErrors(object):
             _exhigh, _eyhigh = self.member("fEXhigh"), self.member("fEYhigh")
 
             _low = numpy.asarray([_exlow, _eylow], dtype=_exlow.dtype.newbyteorder("="))
-            _high = numpy.asarray([_exhigh, _eyhigh], dtype=_exhigh.dtype.newbyteorder("="))
+            _high = numpy.asarray(
+                [_exhigh, _eyhigh], dtype=_exhigh.dtype.newbyteorder("=")
+            )
             errors = (_low, _high)
             self._errors = errors
 
--- a/uproot/behaviors/TGraphAsymmErrors.py
+++ b/uproot/behaviors/TGraphAsymmErrors.py
@@ -8,12 +8,15 @@ from __future__ import absolute_import
 
 import numpy
 
-# dict_keys(['@fUniqueID', '@fBits', 'fName', 'fTitle', 'fLineColor', 'fLineStyle', 'fLineWidth', 'fFillColor', 'fFillStyle', 'fMarkerColor', 'fMarkerStyle', 'fMarkerSize', 'fNpoints', 'fX', 'fY', 'fFunctions', 'fHistogram', 'fMinimum', 'fMaximum', 'fEXlow', 'fEXhigh', 'fEYlow', 'fEYhigh'])
+# ['@fUniqueID', '@fBits', 'fName', 'fTitle', 'fLineColor', 'fLineStyle',
+#  'fLineWidth', 'fFillColor', 'fFillStyle', 'fMarkerColor', 'fMarkerStyle',
+#  'fMarkerSize', 'fNpoints', 'fX', 'fY', 'fFunctions', 'fHistogram',
+#  'fMinimum', 'fMaximum', 'fEXlow', 'fEXhigh', 'fEYlow', 'fEYhigh']
 
 
 class TGraphAsymmErrors(object):
     """
-    Behaviors for TGraphAsymmErrors, not including ``TGraph``
+    Behaviors for TGraphAsymmErrors: get values and errors as NumPy arrays.
     """
 
     def _normalize_array(self, key):
--- a/uproot/behaviors/TGraphAsymmErrors.py
+++ b/uproot/behaviors/TGraphAsymmErrors.py
@@ -20,48 +20,43 @@ class TGraphAsymmErrors(object):
         values = self.member(key)
         return numpy.asarray(values, dtype=values.dtype.newbyteorder("="))
 
+    def _reduce_errors(self, which, lowkey, highkey):
+        if which == "low":
+            return self._normalize_array(lowkey)
+        elif which == "high":
+            return self._normalize_array(highkey)
+        elif which == "mean":
+            return (
+                self._normalize_array(highkey) + self._normalize_array(lowkey)
+            ) / 2.0
+        elif which == "diff":
+            return self._normalize_array(highkey) - self._normalize_array(lowkey)
+
     def values(self, axis="both"):
         if axis in [0, -2, "x"]:
             return self._normalize_array("fX")
         elif axis in [1, -1, "y"]:
             return self._normalize_array("fY")
         elif axis == "both":
-            return (self._normalize_array("fX"), self._normalize_array("fY"))
+            return (self.values("x"), self.values("y"))
         else:
             raise ValueError(
                 "axis must be 0 (-2), 1 (-1) or 'x', 'y' or 'both' for a TGraphAsymmErrors"
             )
 
-    def errors(self, axis="both", which="both"):
-        """
-        ( array[XLow, YLow], array[XHigh, YHigh] )
-        """
-        axes = []
+    def errors(self, which, axis="both"):
+        if which not in ["low", "high", "mean", "diff"]:
+            raise ValueError(
+                "which must be 'low', 'high', 'mean', or 'diff' for a TGraphAsymmErrors"
+            )
+
         if axis in [0, -2, "x"]:
-            axes = ["X"]
+            return self._reduce_errors(which, "fEXlow", "fEXhigh")
         elif axis in [1, -1, "y"]:
-            axes = ["Y"]
+            return self._reduce_errors(which, "fEYlow", "fEYhigh")
         elif axis == "both":
-            axes = ["X", "Y"]
+            return (self.errors(which, "x"), self.errors(which, "y"))
         else:
             raise ValueError(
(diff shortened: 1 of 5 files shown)


uproot/behaviors/TGraphAsymmErrors.py
```diff
@@ -0,0 +1,48 @@
...
+import uproot
+
+# dict_keys(['@fUniqueID', '@fBits', 'fName', 'fTitle', 'fLineColor', 'fLineStyle', 'fLineWidth', 'fFillColor', 'fFillStyle', 'fMarkerColor', 'fMarkerStyle', 'fMarkerSize', 'fNpoints', 'fX', 'fY', 'fFunctions', 'fHistogram', 'fMinimum', 'fMaximum', 'fEXlow', 'fEXhigh', 'fEYlow', 'fEYhigh'])
+
+class TGraphAsymmErrors(object):
+    """
+    Behaviors for TGraphAsymmErrors, not including ``TGraph``
+    """
+
+    def values(self):
```
jpivarski, line 20: I like the names `values` and `errors` (as methods, not properties), since this matches TH*.

I just deleted a lot of review because I'm thinking that they shouldn't be returning tuples that users need to unpack. For instance, if someone really wants just the low error bars on _y_, they'd have to say things like

```python
hist.errors()[1][0]
```

I know that `self.member("fEYlow")` is also an option, but most people won't discover that when there's a documented `errors` method available.

The TH* behaviors have an `axis` keyword that works like NumPy's (`axis=0` and `axis=-2` are the _x_ axis; `axis=1` and `axis=-1` are the _y_ axis) with ROOT-like extensions (`axis="x"` and `axis="y"`). If it also had a `dir` argument, each thing could be extracted without having to know the output tuple format:

```python
hist.values("x")
hist.values("y")
hist.errors("x", "low")
hist.errors("x", "high")
hist.errors("y", "low")
hist.errors("y", "high")
```

and then we could get this packed format by introducing a meaning for `"both"` on `axis` and `dir`:

```python
hist.values("both")
hist.errors("both", "both")
```

Part of the issue with returning both is that I would have guessed that the interface returns objects of shape `(N, 2)`, rather than `(2, N)`. That is, you have it returning the equivalent of `numpy.vstack([_x, _y])` (and `vstack` is more vectorized than iterating over `_x` and `_y` in `numpy.asarray([_x, _y])`), whereas I would have expected `numpy.dstack([_x, _y])[0]`:

```
>>> numpy.vstack([_x, _y])
array([[ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9],
       [ 5,  6,  7,  8,  9, 10, 11, 12, 13, 14]], dtype=int32)
>>> numpy.dstack([_x, _y])[0]
array([[ 0,  5],
       [ 1,  6],
       [ 2,  7],
       [ 3,  8],
       [ 4,  9],
       [ 5, 10],
       [ 6, 11],
       [ 7, 12],
       [ 8, 13],
       [ 9, 14]], dtype=int32)
```

I think of the first axis as being the one ranging over a large dataset. (That's how Pandas would index it, too.) There are some NumPy interfaces that return a _tuple_ of 2 arrays, but not an array of shape `(2, N)`. So I just think the `(N, 2)` form (`dstack`) or 2-tuple (no "stack") would be more expected.

With `errors(axis="both", dir="both")`, should it be a 3D vector, in which the second and third dimension both have length 2? Or a 2D vector, in which the second dimension has length 4? (I think more structure is better, so 3D.) However it's set up, there's going to be some guessing or searching documentation to determine whether the two axes are paired more closely than the two directions or vice-versa (though most people would guess that _x_ comes before _y_ and _low_ comes before _high_).

Getting six 1D arrays is more typing, but it's the only unambiguous case. As a user, I'd want to have that ability, so that I don't have to look up documentation.  `:)`

uproot/behaviors/TGraphAsymmErrors.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        array[X, Y]
+        """
+        if hasattr(self, "_values"):
+            values = self._values
```
jpivarski, line 25: Adding arguments to these functions breaks the cache. I think it should just be dropped (also in `errors`). There isn't any processing here that is computationally intensive enough to justify a cache.

Note: if you have a histogram object, it has already been fully read from disk, just due to how ROOT is serialized. These calls to `member` do not invoke a data-read: they're just accessing data in a dict.

Even if these functions return a computed `dstack` or `vstack`, the array would have to contain at least thousands of items before the time to calculate `dstack`/`vstack` is noticeable compared to the time needed for Python to step through the code. It's likely that checking for a cached value costs more than recomputing it.

jpivarski (commented):
Great, thanks! After reading this over a second time, I have some suggestions for the API, described below.

jpivarski:
> here's a test script to work with

Oh, and if that URL is stable, it can be included as a test in Uproot's test suite (numbered with "0240" to reference this PR). It just needs to be marked as "network" ([like this](https://github.com/scikit-hep/uproot4/blob/434f51f0e3f3d48fd890d071edafabbba35729a4/tests/test_0001-source-class.py#L105)) so that it can be turned off if someone doesn't have network access.

uproot/behaviors/TGraphAsymmErrors.py
kratsg: Thanks for the heavily detailed review here. I've implemented everything in the latest set of changes. I do agree that we need to try and reduce any amount of ambiguity and the best way to do this is to return (in as few calls as makes sense) simple 1-dimensional arrays where possible.

kratsg:
> Oh, and if that URL is stable, it can be included as a test in Uproot's test suite (numbered with "0240" to reference this PR). It just needs to be marked as "network" ([like this](https://github.com/scikit-hep/uproot4/blob/434f51f0e3f3d48fd890d071edafabbba35729a4/tests/test_0001-source-class.py#L105)) so that it can be turned off if someone doesn't have network access.

I've gone ahead and added a full-fledged network-based test for this. We definitely should do something similar for pyhf for network-related tests (if we don't already @matthewfeickert ).

kratsg:
@jpivarski maybe I'll leave this here for now. I thought you would've had python `requests` as part of the test suite. HEPData doesn't allow streaming or chunking ROOT files (I tried) and you just need to download the file entirely which is how I wrote the test... do you want to add `requests` as a test dependency? 🤷‍♂️

jpivarski:
I suppose `requests` could be a test dependency. I had been avoiding it in the main codebase because I wanted to find out what we're paying for with that dependency. (One thing is the automated handling of redirects, but that was easy enough to add to the built-in Python HTTP client.)

The primary place for test files is `scikit-hep-testdata`, which get installed as local files, but the process of adding a file to that repo and pushing a new version of it all before you can test a PR here is a bit arduous. We should probably find a way to distribute local files for tests in a way that doesn't involve PyPI. But that's a longer term project.

As far as I know, Uproot has been the only project using `scikit-hep-testdata`, but if you have similar needs in pyhf, we could maybe band together to make a common test file source (accessible as both HTTP and local, maybe even XRootD, without file size limits and without the steps involved in publishing versions, since the files rarely/never change, they just get added to).

I have a big collection of files on my computer named "uproot4-lazy", which are files associated to Uproot 4 issues that I have been too lazy to add to `scikit-hep-testdata`. They're each about a MB or less.

kratsg:
> As far as I know, Uproot has been the only project using `scikit-hep-testdata`, but if you have similar needs in pyhf, we could maybe band together to make a common test file source (accessible as both HTTP and local, maybe even XRootD, without file size limits and without the steps involved in publishing versions, since the files rarely/never change, they just get added to).

We certainly do have similar needs in pyhf. I think we're going to start seeing this needed for ServiceX as well and various other IRIS-HEP projects. I wonder if it's possible to talk with the Chicago folks here about having something like faxbox available for use. I know we've used CVMFS in the past to unpack and store various docker images so that grid sites only need CVMFS in order to access a subset of docker images (if they are registered there). One concern i've had generally is that this decouples the files from the codes that use it, and it's much harder to figure out who depends on a file when it needs to get updated or changed, or what that policy would be.

This file isn't abundantly large and I could get a pure-python only variation of the setUp fixture here in order to make this work (based on how I factorized the tests) so it's not the end of the world here. That said, i've always thought test dependencies were mostly not as restrictive since only the CI and developers would need to grab them. I guess you're trying to improve the developer experience here.

jpivarski:
> That said, i've always thought test dependencies were mostly not as restrictive since only the CI and developers would need to grab them.

That's what I'm saying: there's a much lower bar to including a dependency for the tests, and I'm definitely willing to make `requests` a test dependency.

henryiii:
Requests is the forth most popular package https://pythonwheels.com , I certainly wouldn't worry about it as a dependency, especially for tests. :)

eduardo-rodrigues:
> As far as I know, Uproot has been the only project using scikit-hep-testdata, but if you have similar needs in pyhf, we could maybe band together to make a common test file source (accessible as both HTTP and local, maybe even XRootD, without file size limits and without the steps involved in publishing versions, since the files rarely/never change, they just get added to).

For completeness - also `pylhe` uses it. For now I believe it only contains in there a single test file, though. So that's 3 org packages using the test package, which is nice.

kratsg:
> For completeness - also `pylhe` uses it. For now I believe it only contains in there a single test file, though. So that's 3 org packages using the test package, which is nice.

Indeed. My only issue is that this file is not a good candidate for a common test utility for multiple packages. `uproot` has the slightly special case that while it's great for it to test lots of "common" ROOT files, it also needs to handle edge cases as well as minimal cases which are not of as-interest for other packages using the `scikit-hep-testdata`. There's a similar issue with `pyhf` because we only want `HistFactory` objects in these files which isn't that useful for something like `uproot` at the moment.

I tend to agree with @jpivarski that it feels a lot more cumbersome to add to the `scikit-hep-testdata` repo for something like this.

eduardo-rodrigues:
Yes. Do keep in mind that the test files can also be handy for tutorials, sometimes. And some (a very small subset) *have* been used in tutorials and hands-on. For example the pylhe test file could also be useful for pyhepmc ...

I'm not saying the situation of having to make a release is perfect.

henryiii:
Ideally, we should haven an online repo of files, and scikit-hep-testdata would just be an interface to that online repository, save for a few, tutorial-ready files locally.

henryiii:
It could be modeled after importlib.resouces.files in Python 3.9, where it returns an object that behaves like a Path opens the remote resource.

kratsg:
fyi @jpivarski this is ready to go.

jpivarski:
> fyi @jpivarski this is ready to go.

Thanks! I just added some formatting and documentation and will merge it when it passes. (I also manually tested Python 2.7.)

jpivarski:
@all-contributors please add @kratsg  for code

Issue #122, opened 2020-10-02
XRootD source doesn't fall back to non-vector read


nsmith-:
In this CMS open data example, I cannot load a branch's array due to a NotImplementedError
```python
>>> import uproot4
>>> file = uproot4.open("root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/ZZTo4mu.root")
>>> file["Events/Muon_eta"].array(entry_stop=244800)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ncsmith/src/uproot4/uproot4/behaviors/TBranch.py", line 2017, in array
    _ranges_or_baskets_to_arrays(
  File "/Users/ncsmith/src/uproot4/uproot4/behaviors/TBranch.py", line 3180, in _ranges_or_baskets_to_arrays
    hasbranches._file.source.chunks(ranges, notifications=notifications)
  File "/Users/ncsmith/src/uproot4/uproot4/source/xrootd.py", line 263, in chunks
    raise NotImplementedError(
NotImplementedError: TODO: Probably need to fall back to a non-vector read
```
Can this be implemented?

jpivarski:
Copying from Slack again:

This "probably" is because I don't understand the "max element size" logic here; it was written by @chrisburr.
However, you can test a non-vector read by setting `xrootd_handler=uproot4.MultithreadedXRootDSource`. This alternative XRootD source doesn't attempt a vector-read and makes separate XRootD connections in Python threads. The number of those Python threads is set by `num_workers` (by default, 1).

HTTP sources do a fallback from multi-part GET to many threads each doing one GET; this would be the same thing for XRootD. I just wasn't sure if the "max element size" failure was reason enough to do the fallback and wanted to get real use-case experience first.

nsmith-:
Another workaround would be using the http endpoint `https://eospublichttp.cern.ch/eos/root-eos/cms_opendata_2012_nanoaod/ZZTo4mu.root` but there they don't have a host certificate that is trusted by default. Can I access the certificate verify option in the HTTP source somehow?

jpivarski:
That would be another requested feature. It would go into

https://github.com/scikit-hep/uproot4/blob/7e42e2f71d9fb20dbfb12d177403b3b0839d0cc2/uproot4/source/http.py#L40

as an argument.

PR #243, opened 2021-01-14
Split vector reads into smaller chunk when single requests are larger than `max_element_size`

nikoladze:
Since i also ran into #122 i would like to help fixing it. I believe better than falling back to non-vector reads in case a single request is larger than the maximum the server allows would be chunking the request into smaller parts as i think [ROOT does it](https://root.cern/doc/master/TNetXNGFile_8cxx_source.html#l00469). The relevant code is probably that part:

```cpp
       // If the length is bigger than max readv size, split into smaller chunks
       if (length[i] > fReadvIorMax) {
          Int_t nsplit = length[i] / fReadvIorMax;
          Int_t rem    = length[i] % fReadvIorMax;
          Int_t j;
  
          // Add as many max-size chunks as are divisible
          for (j = 0; j < nsplit; ++j) {
             offset = position[i] + (j * fReadvIorMax);
             chunks.push_back(ChunkInfo(offset, fReadvIorMax, cursor));
             cursor += fReadvIorMax;
          }
  
          // Add the remainder
          offset = position[i] + (j * fReadvIorMax);
          chunks.push_back(ChunkInfo(offset, rem, cursor));
          cursor += rem;
```

I tried to implement that into `XRootDResource`, but the problem is now that the returned chunks are expected to be the same ranges that were requested. So this would need an additional kind of "merging" step before returning the chunks.
But i don't really know how to do that best - at that stage the chunks are not necessarily "materialized" yet and might just contain futures - so either one would need to wait for the result and then merge them, creating a new chunk object with merged `raw_data` or modify the task that is submitted to not only fetch the chunks, but also merge them if nescessary ...

Any ideas what could be good ways to deal with this?

### DIFF
--- a/tests/test_0001-source-class.py
+++ b/tests/test_0001-source-class.py
@@ -329,6 +329,34 @@ def test_xrootd_vectorread_max_element_split():
 
 @pytest.mark.network
 @pytest.mark.xrootd
+def test_xrootd_vectorread_max_element_split_consistency():
+    pytest.importorskip("XRootD")
+    filename = "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root"
+    def get_chunk(Source, **kwargs):
+        with Source(
+            filename, **kwargs
+        ) as source:
+            notifications = queue.Queue()
+            max_element_size = 2097136
+            chunks = source.chunks([(0, max_element_size + 1)], notifications)
+            one, = [tobytes(chunk.raw_data) for chunk in chunks]
+            return one
+    chunk1 = get_chunk(
+        uproot.source.xrootd.XRootDSource,
+        timeout=10,
+        max_num_elements=None
+    )
+    chunk2 = get_chunk(
+        uproot.source.xrootd.MultithreadedXRootDSource,
+        timeout=10,
+        num_workers=1
+    )
+    assert chunk1 == chunk2
+
+
+
+@pytest.mark.network
+@pytest.mark.xrootd
 def test_xrootd_vectorread_fail():
     pytest.importorskip("XRootD")
     with pytest.raises(Exception) as err:
--- a/tests/test_0001-source-class.py
+++ b/tests/test_0001-source-class.py
@@ -313,6 +313,21 @@ def test_xrootd_vectorread():
 
 @pytest.mark.network
 @pytest.mark.xrootd
+def test_xrootd_vectorread_max_element_split():
+    pytest.importorskip("XRootD")
+    with uproot.source.xrootd.XRootDSource(
+        "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root",
+        timeout=10,
+        max_num_elements=None,
+    ) as source:
+        notifications = queue.Queue()
+        max_element_size = 2097136
+        chunks = source.chunks([(0, max_element_size + 1)], notifications)
+        assert len(tobytes(chunks[0].raw_data)) == max_element_size + 1
+
+
+@pytest.mark.network
+@pytest.mark.xrootd
 def test_xrootd_vectorread_fail():
     pytest.importorskip("XRootD")
     with pytest.raises(Exception) as err:
--- a/tests/test_0001-source-class.py
+++ b/tests/test_0001-source-class.py
@@ -323,7 +323,8 @@ def test_xrootd_vectorread_max_element_split():
         notifications = queue.Queue()
         max_element_size = 2097136
         chunks = source.chunks([(0, max_element_size + 1)], notifications)
-        assert len(tobytes(chunks[0].raw_data)) == max_element_size + 1
+        one, = [tobytes(chunk.raw_data) for chunk in chunks]
+        assert len(one) == max_element_size + 1
 
 
 @pytest.mark.network
--- a/tests/test_0001-source-class.py
+++ b/tests/test_0001-source-class.py
@@ -301,6 +301,7 @@ def test_xrootd_vectorread():
         "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root",
         timeout=10,
         max_num_elements=None,
+        num_workers=1
     ) as source:
         notifications = queue.Queue()
         chunks = source.chunks([(0, 100), (50, 55), (200, 400)], notifications)
@@ -319,6 +320,7 @@ def test_xrootd_vectorread_max_element_split():
         "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root",
         timeout=10,
         max_num_elements=None,
+        num_workers=1
     ) as source:
         notifications = queue.Queue()
         max_element_size = 2097136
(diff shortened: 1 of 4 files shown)


jpivarski:
Thanks! I'm glad that you're taking this on—you likely know a lot more about XRootD than I do.

As for merging Chunks, the thing to do would be for the Chunk's `future` to be a function that initiates the next step in the operation. The HTTP service used to do this when it was handling a case that gets the length of the file _and then_ gets the data, though I've removed that because that optimization of fetching the last bytes in the file didn't provide a benefit to outweigh its complexity. I also tried replacing the `future` object from the relevant Chunks, which is perhaps more complicated than writing the `future` function in such a way that it could do either operation.

Downstream of this, Uproot only uses the single-Chunk interface (for everything except TBaskets) and the `notifications` queue of the multi-Chunk interface (for TBaskets), so it can deal with Chunks in any order and only starts working on a Chunk when its `future` function finishes.

   * [This wiki](https://github.com/scikit-hep/uproot4/wiki/Physical-layer-interface) explains that in detail.
   * Early tests, [like this one](https://github.com/scikit-hep/uproot4/blob/main/tests/test_0001-source-class.py), access the physical layer in isolation, which is probably how you want to do it when debugging.

If having the `future` of each Chunk deal with the exceptional case by putting extra logic in that function is too complex, then you might want to instead consider a two-phase approach: in the `XRootDSource.chunks` method, do one request that will determine how it _ought_ to be chunked, then do another that requests XRootD-friendly chunks and if a single Uproot Chunk requires more than one of these XRootD-friendly chunks, then it has to wait on them all.

Another possibility: it might actually simplify things to have two ResourceExecutors, one that the Uproot sets up to wait for its Chunks, and another that the XRootDSource sets up to wait for XRootD-friendly chunks that feed into the Uproot Chunks.

These are just some initial hints. I don't understand the problem well enough to make a definitive suggestion, but you can contact me on Gitter ([Scikit-HEP/uproot](https://gitter.im/Scikit-HEP/uproot)) to talk it over in real time.

nikoladze:
Thanks @jpivarski again for the detailed explanation! That wiki page was also very helpful.

I tried it now with a variant similar to the suggestion of having 2 separate Executors. The chunks in the `XRootDSource` are actually retrieved using the internal threads from the xrootd library and receive a python callback that calls `_run` for the futures to trigger the notifications. I added a `ResourceThreadPoolExecutor` to `XRootDSource` that executes potential tasks for merging sub-chunks. Those tasks receive the futures of the sub-chunks, wait for them and merge the buffers. Maybe i should also add a `num_workers` argument since this now probably only uses one thread.

@jpivarski  you mentioned that uproot also sets up a ResourceExecutor (meaning outside of the Source?). Can i access this from within the Source and use that instead of the Executor i added to `XRootDSource`?

jpivarski:
> @jpivarski you mentioned that uproot also sets up a ResourceExecutor (meaning outside of the Source?). Can i access this from within the Source and use that instead of the Executor i added to `XRootDSource`?

All I meant was that Uproot maintains a single ResourceExecutor (passed in by the user or a default of `array`/`arrays`/etc.), but if it helps for XRootDSource to have a second ResourceExecutor to manage splitting and grouping of XRootD-friendly chunks into the Chunks that Uproot actually asked for, then do so.

There isn't another ResourceExecutor beyond Uproot's.

nikoladze:
I think this works now. The `XRootDSource` now has a `ResourceExecutor` to which it submits futures that either wait for the xrootd chunks if they are the same as requested or waits for several of them and merges them if they had to be split up. First i tried to only submit futures in case of merging and use the futures that are given to the xrootd `vector_read` function in case of no merging, but doing so i made the mistake to call `_set_notify` after passing the callback to xrootd which subsequently calls `_run`. That lead to deadlocks in case the thread that reads the queue was faster (took me quite a long time to debug this and i only noticed it by running a larger scale test with 1000 files on 100 workers). So now all chunks in the notifications queue come from futures submitted to the `ResourceExecutor` of `XRootDSource` (and not from the ones passed to `vector_read`).

The `ResourceExecutor` of `XRootDSource` actually doesn't need a resource, since it only forwards or merges chunks that are retrieved via the main xrootd File (and xrootd has it's own threads ...). So i passed `None` as a Resource in the constructor - is this fine? I still wanted to have the rest of the `ResourceExecutor` and `ResourceFuture` methods like `_set_notify`.

jpivarski:
If XRootDSource's `ResourceThreadPoolExcutor` doesn't actually need a resource, then it might as well be a plain [ThreadPoolExecutor](https://github.com/scikit-hep/uproot4/blob/06a17bea7ffb3b9cb17553435daa85ad5b94d421/uproot/source/futures.py#L178-L250). The difference between these is that the "resource" one passes a resource as the first argument to the future, so it sounds like you can use the simple one.

Uproot's `ThreadPoolExecutor` is functionally identical to a subset of Python 3's `ThreadPoolExecutor`, but I explicitly defined that subset so that it can be used in Python 2 (and to make the generalization to `ResourceThreadPoolExcutor` explicit by putting them side-by-side).

jpivarski (approved):
Awesome! I don't see any issues here, and a `mergefuture` function is exactly what I'd expect.

See above the comment I made about simplifying `ResourceThreadPoolExecutor` → `ThreadPoolExecutor`, since you don't need each worker in this second executor to maintain a unique (stateful) resource.

I see that you have a physical-layer test, and I've read your comment that you've scaled it up and found & fixed a deadlock bug in the large-scale test.

Without this branch, I'm able to open both `"root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root"` and `"root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root"` and read some `"Muon_pt"` data (first 10 events). Shouldn't one of these fail without the fix? If this is only something that happens at large scale, have you done a full test, all the way to physics data, that a file that couldn't be read before can be read now?

nikoladze:
The example from @nsmith- in #122 fails without the fix and works now:

```pycon
>>> import uproot
>>> file = uproot.open("root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/ZZTo4mu.root")
>>> array1 = file["Events/Muon_eta"].array(entry_stop=244800)
```

we can also cross check this against `MultithreadedXRootDSource`

```pycon
>>> file2 = uproot.open(
...     "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/ZZTo4mu.root",
...     xrootd_handler=uproot.MultithreadedXRootDSource
... )
>>> array2 = file2["Events/Muon_eta"].array(entry_stop=244800)
>>> import awkward as ak
>>> assert ak.all(array1 == array2)
```

I chose `ResourceThreadPoolExecutor`/`ResourceFuture` over `ThreadPoolExecutor`/`Future` because it has this mechanism of adding the chunk to the notifications queue (via `_notify`, `_set_notify`), which `ThreadPoolExecutor`/`Future` lack.

jpivarski:
Thanks for the explicit check!

> I chose `ResourceThreadPoolExecutor`/`ResourceFuture` over `ThreadPoolExecutor`/`Future` because it has this mechanism of adding the chunk to the notifications queue (via `_notify`, `_set_notify`), which `ThreadPoolExecutor`/`Future` lack.

That makes sense. Okay, I guess this is ready!

PR #254, opened 2021-01-27
Prevent the xrootd server readv_iov_max limit from being exceeded

### LINKED ISSUE #253: XRootD: The number of 'read ahead' elements requested in a readv call can exceed the server's limit
Uproot 4.0.2rc1

I'm sorry that I can't provide a root file / xrd server to reproduce this.

In
https://github.com/scikit-hep/uproot4/blob/3e53f480475883397a3a4e2dea199cdc00fd6477/uproot/source/xrootd.py#L326-L341

if the `request_ranges` list length exceeds `readv_iov_max`, the server will report the error "Read vector is too long" i.e.
```cpp
   if (rdVecNum > XrdProto::maxRvecsz)
      return Response.Send(kXR_ArgTooLong, "Read vector is too long");
``` 
where `rdVecNum = rdVecLen / sizeof(readahead_list);`,  which corresponds to the number of elements in the `request_ranges` list sent to the server.

I think that the code at https://github.com/scikit-hep/uproot4/blob/3e53f480475883397a3a4e2dea199cdc00fd6477/uproot/source/xrootd.py#L294-L298

may be the cause, since it does the max element size check. Changing
```
     if len(all_request_ranges[-1]) > self._max_num_elements: 
```
to
```
     if len(all_request_ranges[-1]) >= self._max_num_elements: 
```
seems to fix the issue in my test. Will open a PR.

ryuwd:
Fixes #253 

This change just prevents the length of any sublist in `all_request_ranges` from exceeding the `readv_iov_max` limit.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -11,10 +11,8 @@ jobs:
     strategy:
       matrix:
         platform: ["windows-latest", "macos-latest", "ubuntu-latest"]
-        python-version: ["3.5", "3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9"]
         exclude:
-          - platform: "macos-latest"
-            python-version: "3.5"
           - platform: "macos-latest"
             python-version: "3.6"
           - platform: "windows-latest"
@@ -54,7 +52,7 @@ jobs:
           conda list
 
       - name: "Install XRootD"
-        if: "matrix.python-version != 3.5 && runner.os != 'macOS'  &&  runner.os != 'Windows'"
+        if: "runner.os != 'macOS'  &&  runner.os != 'Windows'"
         run: |
           conda env list
           conda install xrootd
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![PyPI version](https://badge.fury.io/py/uproot.svg)](https://pypi.org/project/uproot)
 [![Conda-Forge](https://img.shields.io/conda/vn/conda-forge/uproot)](https://github.com/conda-forge/uproot-feedstock)
-[![Python 3.5‒3.9](https://img.shields.io/badge/python-3.5%E2%80%923.9-blue)](https://www.python.org)
+[![Python 3.6‒3.9](https://img.shields.io/badge/python-3.6%E2%80%923.9-blue)](https://www.python.org)
 [![BSD-3 Clause License](https://img.shields.io/badge/license-BSD%203--Clause-blue.svg)](https://opensource.org/licenses/BSD-3-Clause)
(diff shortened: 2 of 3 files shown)


ryuwd:
The CI failure is caused by a HTTP 500 error from a remote server in one of the tests https://github.com/scikit-hep/uproot4/pull/254/checks?check_run_id=1777200916#step:9:441

jpivarski:
That HTTP issue was unrelated to the XRootD thing you fixed, so I restarted the jobs.

jpivarski:
Now it's failing because pip is using syntax that's not legal in Python 3.5: "`sys.stderr.write(f"ERROR: {exc}")`". I wonder why it's pulling in a version of pip that's incompatible with the Python that's running.

ryuwd:
Thanks! Yes, I saw that, odd. Something wrong with `conda-incubator/setup-miniconda` ?

henryiii:
Python 3.5 hasn't been supported by conda in a long time. Normally I'd say this sounds like SLC6; `pip install -U pip` from an old pip that doesn't know about the Requires-Python metadata slot will upgrade to pip 21.0 and break.

henryiii:
Ahh, I think the problem is pip on conda is a no-arch package, meaning it's not tied to Python version. It's listed as >=3, not >=3.6. If I can manage to get anyone to care about 3.5, I might be able to get it fixed. But not sure I can get anyone to care.

jpivarski:
I don't know if it's conda, but maybe. I know that a lot of things have been dropping Python 3.5 support.

Uproot technically works in Python 3.5, just as it works in Python 2.7 (and even Python 2.6 in some tests I performed personally), but we can't claim support if it's not regularly tested in CI, and it's hard to run CI on old platforms.

I have less of an attachment to Python 3.5 than 2.7 because use-cases are more likely to be stuck in Python 2.7 than 3.5—any system on a Python 3 branch can update rather easily, but Python 2 has locked in some code with large, messy codebases. So, having given up on trying to keep Python 2.7 in CI, I might as well give up on 3.5.

jpivarski:
Hi, @henryiii! Our messages crossed. I agree that it's hard to care about Python 3.5: if you're already on Python 3, you might as well upgrade already. (Also, its non-deterministic dict order has been a pain for me. I don't mind seeing it go.)

@ryuwd, if you don't have any other changes, give me a thumbs-up and I'll squash-and-merge. Thanks!

jpivarski:
@all-contributors please add @ryuwd for code

ryuwd:
Thanks @jpivarski !

chrisburr:
I've opened a PR to fix the issue you saw with `pip`: https://github.com/conda-forge/pip-feedstock/pull/68

jpivarski:
@chrisburr Thanks! I didn't think there would be this interest in Python 3.5. (Although now that I think about it, the failure should at least be prevented because it can be hard to understand the error message: "f-strings are not correct syntax.")

PR #256, opened 2021-01-29
Re-enable Python 3.5 CI

chrisburr:
I'm making this PR to check if conda-forge was fixed correctly I don't mind if it's merged or just closed.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -11,8 +11,10 @@ jobs:
     strategy:
       matrix:
         platform: ["windows-latest", "macos-latest", "ubuntu-latest"]
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.5", "3.6", "3.7", "3.8", "3.9"]
         exclude:
+          - platform: "macos-latest"
+            python-version: "3.5"
           - platform: "macos-latest"
             python-version: "3.6"
           - platform: "windows-latest"
@@ -52,7 +54,7 @@ jobs:
           conda list
 
       - name: "Install XRootD"
-        if: "runner.os != 'macOS'  &&  runner.os != 'Windows'"
+        if: "matrix.python-version != 3.5 && runner.os != 'macOS'  &&  runner.os != 'Windows'"
         run: |
           conda env list
           conda install xrootd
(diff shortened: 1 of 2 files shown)


jpivarski:
If it does work, I'll merge it!

henryiii:
If we say we support 3.5, I want to have at least one ci job running 3.5 (same goes for 2.7, actually). (Though Linux only would be just fine). If we claim to support 3.5 but actually don't, that's _worse_ than just changing our minimum requires to remove 3.5, causing pip to pull old versions instead. This is exactly the same issue that pip's conda feedstock had, they didn't change what they claimed to support, so it broke!

henryiii:
```
  OSError: XRootD error: [ERROR] Operation expired
E               in file root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root
```

chrisburr:
I'll mark this as ready for review as the issue it's claiming to fix is sorted.

I've left a script running to see if I can get enough of a reproducer to open a ticket against EOS...

chrisburr:
Ah I've finally figured it out, unauthenticated reads are _**often**_ failing but authenticated ones are _**always**_ fine.

I've opened a ticket: https://cern.service-now.com/service-portal?id=ticket&n=INC2684948

* With kerberos credentials:
```bash
$ ssh lxplus 'echo "Start" && time XRD_REQUESTTIMEOUT=20 xrdcp root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root - | head -c1000 > /dev/null'
Start
XRD_REQUESTTIMEOUT=20 xrdcp  -  0.01s user 0.03s system 39% cpu 0.106 total
head -c1000 > /dev/null  0.00s user 0.00s system 1% cpu 0.103 total
```

* Destroy kerberos credentials before reading:
```bash
$ ssh lxplus 'kdestroy && echo "Start" && time XRD_REQUESTTIMEOUT=20 xrdcp root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root - | head -c1000 > /dev/null'
Start
[0B/0B][100%][==================================================][0B/s]
Run: [ERROR] Operation expired:  (source)
XRD_REQUESTTIMEOUT=20 xrdcp  -  0.03s user 0.01s system 0% cpu 15.016 total
head -c1000 > /dev/null  0.00s user 0.00s system 0% cpu 15.016 total
```

chrisburr:
@jpivarski The EOS issue is resolved now so this is ready to merge.

If anyone is interested there is a nice explanation about the root cause on the support ticket.

PR #258, opened 2021-02-08
Cast start and length to int before passing to pyxrootd.client.File.read

nikoladze:
When requesting a single chunk with `uproot.source.xrootd.XRootDSource` the arguments have to be eventually casted to (python) int before giving them to `pyxrootd.client.File.read`, otherwise it will throw an exception. This happenend when requesting data via `uproot.TBranch.basket` because that passes the parameters through as `np.int64` (coming from `branch.member("fBasketSeek")`:

```pycon
>>> import uproot
>>> branch = uproot.open("root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root:Events/Muon_pt")
>>> branch.basket(0)
  File "/home/nikolai/python/uproot4/uproot/behaviors/TBranch.py", line 2484, in basket
    chunk, cursor = self.basket_chunk_cursor(basket_num)
  File "/home/nikolai/python/uproot4/uproot/behaviors/TBranch.py", line 2508, in basket_chunk_cursor
    chunk = self._file.source.chunk(start, stop)
  File "/home/nikolai/python/uproot4/uproot/source/xrootd.py", line 277, in chunk
    data = self._resource.get(start, stop)
  File "/home/nikolai/python/uproot4/uproot/source/xrootd.py", line 152, in get
    status, data = self._file.read(start, stop - start, timeout=self._xrd_timeout())
  File "/home/nikolai/.conda/envs/physlite-experiments/lib/python3.8/site-packages/XRootD/client/file.py", line 126, in read
    status, response = self.__file.read(offset, size, timeout)
TypeError: integer argument expected for offset
```

This should be fixed by this MR. I added a test that directly tries to pass `np.int64` to `uproot.source.xrootd.XRootDSource.chunk`.

### DIFF
--- a/tests/test_0001-source-class.py
+++ b/tests/test_0001-source-class.py
@@ -392,6 +392,20 @@ def test_xrootd_size():
     assert size1 == 3469136394
 
 
+@pytest.mark.network
+@pytest.mark.xrootd
+def test_xrootd_numpy_int():
+    pytest.importorskip("XRootD")
+    with uproot.source.xrootd.XRootDSource(
+        "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root",
+        timeout=10,
+        max_num_elements=None,
+        num_workers=1,
+    ) as source:
+        chunk = source.chunk(numpy.int64(0), numpy.int64(100))
+        assert len(chunk.raw_data) == 100
+
+
 def test_cursor_debug():
     data = numpy.concatenate(
         [
--- a/uproot/source/xrootd.py
+++ b/uproot/source/xrootd.py
@@ -149,7 +149,11 @@ in file {1}""".format(
 
         Returns a Python buffer of data between ``start`` and ``stop``.
         """
-        status, data = self._file.read(start, stop - start, timeout=self._xrd_timeout())
+        status, data = self._file.read(
+            int(start),
+            int(stop - start),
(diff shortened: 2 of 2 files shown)


jpivarski:
Great, thanks!

(Grumble, grumble... pyxrootd can't handle `np.int64`? C extensions should call `__int__` on any arguments intended to be integers—it's the sort of thing pybind11 does automatically.)

henryiii:
No, `__index__`, actually. `__int__` means "convert to integer even if you lose something", while `__index__` means "this is exactly this integer". But the argument still applies. ;)

henryiii:
(And it was slightly incorrectly used in the CPython API until 3.8+, a few things did call `__int__` instead of `__index__`)

henryiii:
This is based on the C-API; https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong. In CPython 3.8, `__index__` is called first if you ask for `PyLong_AsLong(obj)` and falling back on `__int__` is deprecated.

Anyway, `__float__` converts lossily to a float, `__str__` converts lossily to a string, and `__int__` converts lossily to an int. Generally, you would try `__index__`, and dispatch an int version, and then fall back on `__float__`. If you want (possibly) lossy conversant to ints, of course, then `__int__` is completely correct. In this case, it looks like this _is_ an index (into an array), so `__index__` is correct (and is exactly what pybind11 2.6.2 would call).

PR #260, opened 2021-02-09
Update basic.rst with conditional cut example

ChristopheRappold:
Update the basic.rst documentation, in the section "Computing expressions and cuts",
with an example of the conditional cut using the python syntax and explicitly with the
parentheses. Without the parentheses a TypeError will occur, and it is not necessarily
obvious when coming from ROOT and C++ syntax.

### DIFF
--- a/docs-sphinx/basic.rst
+++ b/docs-sphinx/basic.rst
@@ -480,6 +480,14 @@ The second argument is a ``cut``, or filter on entries. Whereas the uncut array 
     >>> events.arrays(["M"], "pt1 > 50", aliases={"pt1": "sqrt(px1**2 + py1**2)"})
     <Array [{M: 91.8}, {M: 91.9, ... {M: 96.1}] type='290 * {"M": float64}'>
 
+or with additional cut conditions expressed using parentheses, the cut array (below) has 269 entries.
+
+.. code-block:: python
+
+    >>> events.arrays(["M"], "(pt1 > 50) & ((E1>100) | (E1<90))", aliases={"pt1": "sqrt(px1**2 + py1**2)"})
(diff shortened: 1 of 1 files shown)


henryiii:
This is actually because this is a binary or being misused instead of a logical or. Due to short circuiting, logical or's cannot be overloaded, so numeric Python had to use the binary operators instead for these logical operations, but the precedence is wrong. There have been a couple of PEPs to fix this in Python, but they have been deferred and everyone seems too busy trying to add pattern matching.

I think it's a good idea, just thought I'd throw in some context.

ChristopheRappold:
Sure, some context is always interesting. Feel free to modify my PR to provide more explanation, I just wanted to make clear for anyone else that would face the same issue how to proceed. I tried to include && and || operators in the example so that people would have a better clue of the syntax.

jpivarski:
I really dislike the use of `&` and `|` instead of `and` and `or`, too, and it's good to show in the docs that it is necessary (as well as the parentheses).

The syntax of the strings is governed by a `language=uproot.language.python.PythonLanguage` that may someday be expanded to include TTreeFormula (through the `formulate` package or a new parser in Lark). Then the documentation examples might use `&&` and `||` without parentheses, but also include `language="root"` (simplifying that parameter setting in a similar way as `language="np"`).

Thanks for this PR!

jpivarski:
@all-contributors, please add @ChristopheRappold for documentation

jpivarski:
@all-contributors, please add @ChristopheRappold for doc

jpivarski:
@all-contributors please add @ChristopheRappold for doc

jpivarski:
(That's why I dislike tools configured by natural language...)

Thanks, @ChristopheRappold for your new contribution!

PR #269, opened 2021-02-15
Skip some tests because HEPData isn't producing the source file anymore.

jpivarski:
The TGraphAsymmErrors tests all assume that a ROOT file can be supplied by

https://www.hepdata.net/download/table/ins1755298/Expected%20limit%201lbb/3/root

However, this URL now returns a page saying, "Converter error encountered." Perhaps the file has been deleted or moved? Could we instead put it in [scikit-hep/scikit-hep-testdata](https://github.com/scikit-hep/scikit-hep-testdata) like the others, @kratsg?

At the end of scikit-hep/uproot4#240, we were talking about the possibility of setting up something that delivers test files. This would be an instance in which relying on a third-party server is a problem, though putting everything into a GitHub repo (scikit-hep-testdata) is not a great solution, either, because of the PyPI file limits and the fact that these data are not really versioned.

### DIFF
--- a/tests/test_0240-read_TGraphAsymmErrors.py
+++ b/tests/test_0240-read_TGraphAsymmErrors.py
@@ -31,6 +31,7 @@ def graph(datafile):
         yield f["Expected limit 1lbb/Graph1D_y1"]
 
 
+@pytest.mark.skip(reason="HEPData can't supply this file: https://www.hepdata.net/download/table/ins1755298/Expected%20limit%201lbb/3/root")
 def test_interpretation(graph):
     assert graph.classname == "TGraphAsymmErrors"
     assert graph.behaviors[0] == uproot.behaviors.TGraphAsymmErrors.TGraphAsymmErrors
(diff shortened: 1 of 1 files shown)


jpivarski:
I have to merge this now to be able to move forward on other issues, but we should definitely reinstate these tests when the files become accessible again.

kratsg:
Hrmm, we should probably ping the hepdata folx on this. I'll do it now and cc you. I'm a little surprised.

GraemeWatt:
Thanks for the alert.  It looks like a general failure of the HEPData converter (e.g. https://converter.hepdata.net/ping should return OK).  We share a Kubernetes cluster with https://inspirehep.net and the INSPIRE people made some changes to the Ingress today, which also affects https://hepdata.net.  I'll let you know when it's fixed (probably tomorrow).

GraemeWatt:
The issue with the HEPData converter certificate has now been fixed.  The bug (HEPData/hepdata#312) only affects table names ending in a `/` followed by an integer, so it's not relevant here.

PR #270, opened 2021-02-15
Probably fixes #265; please test TAxis.__getitem__ and TAxis.intervals, @nsmith-.

### LINKED ISSUE #265: Histogram protocol issue?
I am confused about `PlottableAxisContinuous.__getitem__`: for uproot4 it seems to give the one-indexed values where I was expecting zero-indexed, since `__len__` gives the number of bins without flow. For example:
```
>>> h = uproot.open('https://raw.githubusercontent.com/CoffeaTeam/coffea/master/tests/samples/testSF2d.histo.root:scalefactors_Tight_Electron')
>>> axis = h.axes[0]
>>> len(axis)
10
>>> axis.edges()
array([-2.5  , -2.   , -1.566, -1.444, -0.8  ,  0.   ,  0.8  ,  1.444,
        1.566,  2.   ,  2.5  ])
>>> axis.edges().shape
(11,)
>>> axis[0]
(-2.0, -1.566)
```

jpivarski:


nsmith-:
Hmm at least for my test example:
```python
h = uproot.open('https://raw.githubusercontent.com/CoffeaTeam/coffea/master/tests/samples/testSF2d.histo.root:scalefactors_Tight_Electron')
axis = h.axes[0]
len(axis)
axis.edges()
axis.edges().shape
axis[0]
```
still gives back `(-2.0, -1.566)`

jpivarski:
Okay, then this is probably correct and necessary. I've been staring at it long enough that I don't see how it could be anything but this. Your test confirms it.

By the way, you see the right thing for `intervals`, too, right? `intervals` gives all the `__getitem__`s in one batch, and therefore it's a different code path, but it needs to be consistent. The default `flow=False` corresponds to `__getitem__`, which doesn't take options like `flow`.

nsmith-:
Though this seems a bit weird:
```
In [21]: axis[-1]
Out[21]: (2.5, -2.5)
```

jpivarski:
> Though this seems a bit weird:
> 
> ```
> In [21]: axis[-1]
> Out[21]: (2.5, -2.5)
> ```
> 
> considering
> 
> ```
> In [22]: axis[len(axis)]
> ...
> IndexError: index 11 is out of bounds for axis 0 with size 11
> ```

This looks okay. That `axis[-1]` is normal Python indexing, which would be the same as `axis[len(axis) - 1]`.

PR #286, opened 2021-02-24
XRootD: raise an exception if read_vector failed

ryuwd:
In
https://github.com/scikit-hep/uproot4/blob/fc32791d3cb420ed95ded64ce544a6ca1484f481/uproot/source/xrootd.py#L221-L225

If the readv request fails, then a `NoneType` `response` can be returned to the `callback` function, which causes a cryptic traceback to be printed complaining that the `response` object has no `chunks` attribute.

This PR doesn't fix any underlying issue, but printing an error message from the xrootd server will show what is going wrong.

### DIFF
--- a/uproot/source/xrootd.py
+++ b/uproot/source/xrootd.py
@@ -211,8 +211,7 @@ in file {1}""".format(
 
         return uproot.source.futures.ResourceFuture(task)
 
-    @staticmethod
-    def callbacker(futures, results):
+    def callbacker(self, futures, results):
         """
         Returns an XRootD callback function to fill the ``futures`` and
         ``results``.
@@ -220,7 +219,7 @@ in file {1}""".format(
 
         def callback(status, response, hosts):
             if status.error:
-                raise OSError(status.message)
+                self._xrd_error(status)
 
             for chunk in response.chunks:
                 start, stop = chunk.offset, chunk.offset + chunk.length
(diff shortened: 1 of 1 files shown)


jpivarski:
Thanks for looking into this! I'm looking at the CI error, and it might be something that expects one error but is getting another.

Instead of raising `OSError`, maybe this should call `self._xrd_error(status)`?

https://github.com/scikit-hep/uproot4/blob/c719f858e775d8bd04b05a69552610e51cdb0047/uproot/source/xrootd.py#L96-L110

The trouble then is that the `callback` is defined in one of `XRootDResource`'s static methods, rather than a normal method. That could be fixed by changing `callbacker` from a static method into a normal method:

https://github.com/scikit-hep/uproot4/blob/c719f858e775d8bd04b05a69552610e51cdb0047/uproot/source/xrootd.py#L214-L215

(dropping `@staticmethod` and adding `self` as the first argument) and calling it as such:

https://github.com/scikit-hep/uproot4/blob/c719f858e775d8bd04b05a69552610e51cdb0047/uproot/source/xrootd.py#L337

(replacing `self.ResourceClass` with `self._resource`).

That ought to do it. I know that the callback/callbacker is not used outside of this file, and these are all the find-matches.

ryuwd:
Thanks @jpivarski! I made the change you suggested.

jpivarski:
And it seems that solved the error, great!

Thanks, again!

PR #289, opened 2021-03-03
feat: to boost histogram interface support

henryiii:
Adding interface from scikit-hep/boost-histogram#483.


jpivarski:
I'm okay with this. Merge when you're done!

henryiii:
As long as the new test line runs on at least one job, I'm happy with it

PR #293, opened 2021-03-04
Add timeout to vector_read

### LINKED ISSUE #292: Xrootd vector read has no timeout
In https://github.com/scikit-hep/uproot4/blob/7e3353cfdec91c340c7a0e26a31a695d2db3bcf1/uproot/source/xrootd.py#L341-L343 there is no timeout argument passed to `vector_read`. The signature supports it: https://xrootd.slac.stanford.edu/doc/python/xrootd-python/modules/client/file.html#XRootD.client.File.vector_read and the regular read does use it:
https://github.com/scikit-hep/uproot4/blob/7e3353cfdec91c340c7a0e26a31a695d2db3bcf1/uproot/source/xrootd.py#L152-L154

nsmith-:
Fixes #292

### DIFF
--- a/uproot/source/xrootd.py
+++ b/uproot/source/xrootd.py
@@ -339,7 +339,9 @@ class XRootDSource(uproot.source.chunk.Source):
             callback = self._resource.callbacker(futures, results)
 
             status = self._resource.file.vector_read(
-                chunks=request_ranges, callback=callback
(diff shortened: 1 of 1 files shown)


jpivarski:
Once these tests pass, I'll merge it. Thanks!

PR #295, opened 2021-03-04
Make MultithreadedXRootDSource the default for xrootd files

### LINKED ISSUE #294: XRootD vector_read leaks memory
XRootD vector_read leaks memory and should not be the default until the issue is fixed upstream.
Here is a reproducible example that demonstrates the problem:

```python
import uproot
import threading
from queue import Queue
import psutil
import time
import tqdm
import gc


notifications = Queue()

def worker():
    # just eat the chunks as they come in
    while True:
        notifications.get()
        notifications.task_done()


usage = []
def collect_usage():
    chunktype = uproot.source.chunk.Chunk
    usage.append({
        "dt": time.time() - tstart,
        "rss": process.memory_info().rss,
        "nchunks": sum(isinstance(o, chunktype) for o in gc.get_objects()),
    })

threading.Thread(target=worker, daemon=True).start()
process = psutil.Process()
step = 1024 * 32  # a typical basket size
step2 = 10  # like number of baskets per chunk

fn = "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root"
with uproot.source.xrootd.XRootDSource(fn, **uproot.open.defaults) as source:
    maxbytes = source.num_bytes  # how much memory would you like to use? :)
    ranges = [(start, start + step) for start in range(0, maxbytes, step)]

    usage = []
    tstart = time.time()
    for i in tqdm.trange(0, len(ranges), step2):
        subranges = ranges[i:i + step2]
        chunks = source.chunks(subranges, notifications)
        notifications.join()
        del chunks  # this *should* be the only reference to the data
        gc.collect()  # the chunks seem to have cycles
        collect_usage()

collect_usage()
import pandas
usage = pandas.DataFrame(usage).set_index("dt")
```

If I then plot the data:
```python
from matplotlib.ticker import EngFormatter

ax = usage.plot(secondary_y="nchunks")
ax.yaxis.set_major_formatter(EngFormatter(unit='b'))
ax.set_title(f"Reading {maxbytes/1e9:.1f}Gb from a file")
```
I get:
image

PS I reported this behavior a long time ago https://github.com/scikit-hep/uproot4/issues/2#issuecomment-629300538

nsmith-:
Fixes #294

### DIFF
--- a/uproot/reading.py
+++ b/uproot/reading.py
@@ -154,7 +154,7 @@ def open(
 
 open.defaults = {
     "file_handler": uproot.source.file.MemmapSource,
-    "xrootd_handler": uproot.source.xrootd.XRootDSource,
+    "xrootd_handler": uproot.source.xrootd.MultithreadedXRootDSource,
     "http_handler": uproot.source.http.HTTPSource,
     "object_handler": uproot.source.object.ObjectSource,
     "timeout": 30,


nsmith-:
Hmm should I have also adjusted `num_workers` ? Because `num_fallback_workers` is what was used for MultithreadedXRootDSource.

jpivarski:
When MultithreadedXRootDSource is the default and not the fallback, it uses `num_workers` to decide how many threads to run. That's `1` because multithreading should be opt-in.

Now I wonder why the same argument doesn't hold for `num_fallback_workers`. Maybe it should.

But anyway, I think `num_workers=1` by default is a good idea. It's easy to set a different number when passing arguments to the `uproot.open` function (and `uproot.iterate`, etc).

PR #296, opened 2021-03-04
Remove Future attributes once done with them

nsmith-:
This helps to break reference cycles in the future notifier, which improves the plot from https://github.com/scikit-hep/uproot4/issues/294#issuecomment-790984804 to:
image

### DIFF
--- a/uproot/source/futures.py
+++ b/uproot/source/futures.py
@@ -131,6 +131,7 @@ class Future(object):
         except Exception:
             self._excinfo = sys.exc_info()
         self._finished.set()
+        del self._task, self._args
 
 
 class Worker(threading.Thread):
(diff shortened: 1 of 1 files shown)


uproot/source/futures.py
```diff
@@ -288,6 +292,7 @@ def _run(self, resource):
         if self._notify is not None:
             self._notify()
             del self._notify
+            self._notify = None
```
nsmith-, line 295: Adding these should mean that `del blah` is redundant. I guess you want to guarantee the attribute exists even though it may be null?

jpivarski: The `del blah` is now redundant, but I just wanted to be 100% sure. I don't think `del` does any magic other than what reassignment does, but whatever.

Anyway, keeping the `del blah` in there expresses the intention more clearly.

nsmith-: Yeah as far as I know `del self.blah` is syntactic sugar for `self.__dict__.__delitem__("blah")`.

PR #297, opened 2021-03-05
Remove more cycles and release heavy stuff early

nsmith-:
Targeting specifically Chunk and TBasket objects, because they are some of the heaviest intermediate results.

Testing with the following:
```python
import uproot
import psutil
import time
import tqdm
import gc

def getmallinfo():
    import ctypes
    class MallInfo(ctypes.Structure):
        _fields_ = [(name, ctypes.c_int)
                    for name in ('arena', 'ordblks', 'smblks', 'hblks', 'hblkhd',
                                'usmblks', 'fsmblks', 'uordblks', 'fordblks',
                                'keepcost')]

    libc = ctypes.CDLL("libc.so.6")
    mallinfo = libc.mallinfo
    mallinfo.argtypes = []
    mallinfo.restype = MallInfo
    return mallinfo()


uproot.open.defaults["num_workers"] = 10

usage = []
def collect_usage(asize):
    otype = uproot.source.chunk.Chunk
    mallinfo = getmallinfo()
    usage.append({
        "dt": time.time() - tstart,
        "rss": process.memory_info().rss,
        "arraysize": asize,
        "ordblks": mallinfo.ordblks,
        "uordblks": mallinfo.uordblks,
        "fordblks": mallinfo.fordblks,
        "nobj": sum(isinstance(o, otype) for o in gc.get_objects()),
    })

process = psutil.Process()
tstart = time.time()
collect_usage(0)
fn = "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root:Events"
with uproot.open(fn, object_cache=None, array_cache=None) as tree:
    steps = tree.common_entry_offsets()
    steps = [s for s in zip(steps[:-1], steps[1:]) if s[1] < (tree.num_entries // 10)]
    for start, stop in tqdm.tqdm(steps):
        nbytes = sum(
            branch.array(entry_start=start, entry_stop=stop).nbytes
            for branch in tree
        )
        collect_usage(nbytes)

collect_usage(0)
import pandas
usage = pandas.DataFrame(usage).set_index("dt")
```

Initially the TBasket cycles take a while to be garbage collected, so they pile up and even if gc cleans up, glibc malloc might decide to keep the arena block around (fordblocks = free ordinary blocks, I think):
image

After refactoring `Model.concrete` we have overall lower and much less varied memory usage:
image

Lastly, try to make sure the TBaskets are released sooner (not a huge change)
arrays_fixbasket_oneatatime_dropearly

### DIFF
--- a/tests/test_0017-multi-basket-multi-branch-fetch.py
+++ b/tests/test_0017-multi-basket-multi-branch-fetch.py
@@ -133,6 +133,7 @@ def test_ranges_or_baskets_to_arrays():
             interpretation_executor,
             library,
             arrays,
+            False,
         )
         assert arrays[branch.cache_key].tolist() == [
             -15,
--- a/uproot/behaviors/TBranch.py
+++ b/uproot/behaviors/TBranch.py
@@ -3458,13 +3458,11 @@ def _ranges_or_baskets_to_arrays(
         obj = notifications.get()
 
         if isinstance(obj, uproot.source.chunk.Chunk):
-            chunk = obj
-            args = range_args[(chunk.start, chunk.stop)]
-            decompression_executor.submit(chunk_to_basket, chunk, *args)
+            args = range_args[(obj.start, obj.stop)]
+            decompression_executor.submit(chunk_to_basket, obj, *args)
 
         elif isinstance(obj, uproot.models.TBasket.Model_TBasket):
-            basket = obj
-            interpretation_executor.submit(basket_to_array, basket)
+            interpretation_executor.submit(basket_to_array, obj)
 
         elif obj is None:
             pass
@@ -3474,6 +3472,8 @@ def _ranges_or_baskets_to_arrays(
 
         else:
             raise AssertionError(obj)
+
+        obj = None  # release before blocking
 
 
 def _fix_asgrouped(arrays, expression_context, branchid_interpretation, library, how):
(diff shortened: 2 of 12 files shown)


nsmith-:
A nice way to watch object lifetime is to put this somewhere in the constructor:
```python
weakref.finalize(self, lambda s: print(f"Deleting {s} from thread {threading.get_ident()}"), repr(self))
```
The cycles were found with the help of https://code.activestate.com/recipes/523004-find-cyclical-references/

jpivarski:
Summary of what this PR does:

   * Adds a `update_ranges_or_baskets` (bool) parameter to `_ranges_or_baskets_to_arrays`, which is only True for `iterate`. The `ranges_or_baskets` are only are only replaced (see nested function `replace`) if True.
   * Objects like `basket` and `obj` (chunk or basket) that are no longer needed after a loop are decref'ed so that they don't hold memory while the `_ranges_or_baskets_to_arrays` blocks and waits for the next datum.
   * Instead of the concrete object holding a reference `_concrete = self`, it's `_concrete = None` and the `concrete` property returns `self` in case of `None`.
   * A million references to `_concrete` (attribute) have been changed to `concrete` (property) to get this benefit.

Everything here looks perfectly safe.

Issue #38, opened 2020-07-02
Columnar serialization within an entry: is it a ROOT option? What toggles it?


jpivarski:
Consider the weird serialization in scikit-hep/uproot#373, scikit-hep/uproot#374, scikit-hep/uproot#403, scikit-hep/uproot#475, and scikit-hep/uproot#495. It's field-at-a-time inside of each entry. I had thought it was Boost-inside-ROOT, but not for most of the above. It may be some ROOT serialization mode that I'm unaware of.

jpivarski:
Add to that one more: I think scikit-hep/uproot#510 is another example of that. Look at the second file, uproot-issue510b.root:

```python
>>> import uproot4
>>> t = uproot4.open("issue510.root")["EDepSimEvents"]
>>> b = t["Event"]["Trajectories.Points"]
>>> b.debug(0)
```
```
--+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+-
 64   0 101 102  64   9   0   1   0   0   0   2   0   1   0   0   0   0   2   0
  @ ---   e   f   @ --- --- --- --- --- --- --- --- --- --- --- --- --- --- ---
--+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+-
  0   0   0   1   0   0   0   0   2   0   0   0  64   0   0  60   0   4   0   1
--- --- --- --- --- --- --- --- --- --- --- ---   @ --- ---   < --- --- --- ---
--+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+-
  0   0   0   0   2   0   0   0  64   0   0  36   0   3   0   1   0   0   0   0
--- --- --- --- --- --- --- ---   @ --- ---   $ --- --- --- --- --- --- --- ---
--+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+-
  2   0   0   0  64 103 237  14  20  44 204 192  64  99 170 169 116  55  10  48
--- --- --- ---   @   g --- --- ---   , --- ---   @   c --- ---   t   7 ---   0
--+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+-
 64 192  63 130 249   6 230 103  63 240   0   0   0   0   0   0  64   0   0  60
  @ ---   ? --- --- --- ---   g   ? --- --- --- --- --- --- ---   @ --- ---   <
[...]

PR #298, opened 2021-03-05
feat: Memberwise for std::pair<>

kratsg:
Opening this to make it easier to compare but this is incredibly painful :D See #38 for lengthier details for now.

```
>>> [i.members for i in eff.member('fBeta_bin_params').tolist()]
[{'first': -1.0, 'second': -2.0}, {'first': 2.0, 'second': 4.0}, {'first': 4.0, 'second': 8.0}, {'first': 8.0, 'second': 16.0}, {'first': 16.0, 'second': 32.0}, {'first': 32.0, 'second': 64.0}, {'first': 64.0, 'second': 128.0}, {'first': 128.0, 'second': 256.0}, {'first': 256.0, 'second': 512.0}, {'first': 512.0, 'second': 1024.0}, {'first': 1024.0, 'second': 2048.0}, {'first': -1.0, 'second': -2.0}, {'first': 1.0, 'second': 1.0}]
```


kratsg:
make the ROOT file

```python
$ cat make_tefficiency.py 
import ROOT

fp = ROOT.TFile.Open("test-efficiency.root", "RECREATE")

nbins = 11

h_den = ROOT.TH1F('h_den', 'h_den', nbins, 0, 100)
h_num = ROOT.TH1F('h_num', 'h_num', nbins, 0, 100)

for i in range(1, nbins):
    h_num.SetBinContent(i, 2**i)
    h_den.SetBinContent(i, 2**(i+1))

eff = ROOT.TEfficiency(h_num, h_den)
eff.SetName('TEfficiencyName')
eff.SetTitle('TEfficiencyTitle')

eff.SetBetaBinParameters(0, -1.0, -2.0)
for i in range(1, nbins):
    eff.SetBetaBinParameters(i, 2**i, 2**(i+1))

eff.SetBetaBinParameters(nbins, -1.0, -2.0)

h_den.Write()
h_num.Write()
eff.Write()
fp.Close()
```

then run it

```python
$ cat debug.py 
import uproot

with uproot.open('test-efficiency.root') as fp:
    eff = fp['TEfficiencyName']
```

uproot/containers.py
```diff
@@ -792,19 +792,45 @@ def read(self, chunk, cursor, context, file, selffile, parent, header=True):
...
+    in file {2}""".format(
+                        type(self).__name__, _value_typename, selffile.file_path
+                    )
                 )
-            )
 
-        length = cursor.field(chunk, _stl_container_size, context)
+            # there's extra stuff, maybe?
+            _num_memberwise_bytes = cursor.field(chunk, _stl_container_size, context)
+            _something_else = cursor.field(chunk, struct.Struct(">H"), context)
```
kratsg, line 809: This portion that we're at right now looks like

```
--+---+---+---+---+---+---+---+---+---+
  0   0   0 215 190 210   0   0   0  13  
--- --- --- --- --- --- --- --- --- --- 
```

Unclear what the first 6 bytes are here, but the last 4 seem to be number of elements in the vector (`length`) in the line below. I assumed it might be 4 bytes / 2 bytes but still not sure what they're meant to match up with.

uproot/containers.py
```diff
@@ -792,19 +792,45 @@ def read(self, chunk, cursor, context, file, selffile, parent, header=True):
...
+
+            # length is number of elements in vector
+            length = cursor.field(chunk, _stl_container_size, context)
+
+            model = self._values.new_class(file, "max")
```
kratsg, line 814: can we always assume `max` version? I guess if we're inside `AsVector`, we must be looking at an STLType or a custom C++ type which generally means there wouldn't be a streamer defined before loading the streamer from the file.

uproot/containers.py
```diff
@@ -792,19 +792,45 @@ def read(self, chunk, cursor, context, file, selffile, parent, header=True):
...
+
+            # make a shell
+            values = numpy.empty(length, dtype=_stl_object_type)
+            for i in uproot._util.range(length):
+                values[i] = model.read(
+                    chunk, cursor, {**context, "reading": False}, file, selffile, parent
+                )
```
kratsg, line 821: here, we make a `shell`, which means we skip actually reading anything out (in this case, this means "moving the cursor"). We'd just want to inflate all of our classes/objects so we can ask them to each read the file one by one to fill member-wise (see next block).

PR #302, opened 2021-03-12
Allow all sources and caches to be pickled

nsmith-:
Cutting off at the resource acquisition.
The only negative I can think of is it might take people by surprise that the pickle does not completely contain the original data. But I also hope they do not expect it to behave like that.. probably worth a warning in the docs?

### DIFF
--- a/tests/test_0302-pickle.py
+++ b/tests/test_0302-pickle.py
@@ -1,0 +1,67 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot4/blob/main/LICENSE
+
+from __future__ import absolute_import
+
+import pickle
+import sys
+import os
+
+import pytest
+import skhep_testdata
+
+import uproot
+
+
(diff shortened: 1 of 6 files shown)


jpivarski (approved):
Some time ago (#58), I detached file objects from ROOT objects that didn't need them—anything that can't initiate reading (which is everything except for ReadOnlyDirectory and HasBranches)—so that they could be pickled.

This PR extends pickleability to those last few cases by reopening the file from a pickled state. It won't work if the file has been moved or the pickled object is sent to another computer or something, but that shouldn't be surprising. I think it's not bad to have both solutions, since detaching the file is a cleaner solution when it's possible. (Those unattached objects _can_ be fully used when sent somewhere remote.)

I don't think we have to document the fact that pickling empties caches and that a pickled TTree sent somewhere where the file can't be found would break file-reading. That's such an oddball case that bringing it up might cause more confusion than not mentioning it.

PR #305, opened 2021-03-12
Histograms should be part of the regular tests.

jpivarski:
I need to figure out what's going on with `uproot-issue33.root` conversion into boost-histogram. If I remember right, it's a histogram with labels.

Oh, and @henryiii, you might be interested in this.


henryiii:
We might need to update to the updated metadata system from (IIRC) boost-histogram 0.12+.

PR #314, opened 2021-03-28
Some more memberwise stuff

nsmith-:
I've been trying to read some CMS edm collections, and played around a bit, finding at least one memberwise interpretation successful, with a header like `>h: version` `>I: length` `>I: values_num_bytes`, e.g.
```
--+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+-
  0  10   0   0  16   1  64  22 127 198   0   9   0   0   0   0   0   0   0   0
--- --- --- --- --- ---   @ --- --- --- --- --- --- --- --- --- --- --- --- ---
```
for a class that indeed was version 10 and had the right amount of bytes. This class had a very simple set of members (just one):
```python
    _stl_container0 = uproot.containers.AsVector(False, numpy.dtype("u1"))
    base_names_versions = []
    member_names = ['data_']
    class_flags = {}
```
@kratsg might be interested.

### DIFF
--- a/uproot/containers.py
+++ b/uproot/containers.py
@@ -32,6 +32,7 @@ import uproot
 
 
 _stl_container_size = struct.Struct(">I")
+_stl_vector_header = struct.Struct(">hII")
 _stl_object_type = numpy.dtype(object)
 
 
@@ -796,15 +797,6 @@ class AsVector(AsContainer):
         # (e.g. type(self._values) == type)
         _value_typename = _content_typename(self._values)
         if is_memberwise:
-            # let's hard-code in logic for std::pair<T1,T2> for now
-            if not _value_typename.startswith("pair"):
-                raise NotImplementedError(
-                    """memberwise serialization of {0}({1})
-    in file {2}""".format(
-                        type(self).__name__, _value_typename, selffile.file_path
-                    )
-                )
-
             if not issubclass(self._values, uproot.model.DispatchByVersion):
                 raise NotImplementedError(
                     """streamerless memberwise serialization of class {0}({1})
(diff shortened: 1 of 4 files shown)


uproot/containers.py
```diff
@@ -796,31 +797,52 @@ def read(self, chunk, cursor, context, file, selffile, parent, header=True):
...
+                # FIXME no idea what the other 2 bytes is
+                assert 2 + values_num_bytes + _stl_vector_header.size == num_bytes
+                stuff = cursor.field(  # ignore: F841
+                    chunk, struct.Struct(">h"), context
+                )
+                # print("mystery bytes:", stuff)
+            elif values_version > 0 and values_num_bytes == 65536:
+                # seen in test_0087 files but they have inconsistent header presence in the members
+                length = maybe_length
+            elif values_version == 0:
```
nsmith-, line 839: @kratsg any idea if `pair<blah, blah>` always has this first field set to 0? It's true at least in the test example

uproot/model.py
```diff
@@ -1092,6 +1092,9 @@ def new_class(cls, file, version):
         returns a :doc:`uproot.model.UnknownClassVersion` (adding it to
         ``uproo4.unknown_classes`` if it's not already there).
         """
+        if version in cls.known_versions:
+            return cls.known_versions[version]
```
kratsg, line 1096:
```suggestion
        version = cls.known_versions.get(version)
```

uproot/containers.py
kratsg: Yeah, this part I've never been able to figure it. I thiiiink it is, for `pair` but it's not true for the HistFactory example I'm trying to read in (from RooStats) which seems to do something different... I haven't figured out a pattern yet.

nsmith-:
I think we can close this for now. Perhaps in the future we can revisit. At least some work has been documented.

kratsg:
Yeah, for now let's close this... Jim knows very well that I probably broke my brain trying to understand parts of this back then... and I'm happy to try to stab this again in the future when I have more spare cycles, or if we want to get a student working on this... I have a perfect project for them.

kratsg:
Yeah, that's where it would be nice to get a student/fellow to jump in. I have ROOT files they can use with RooFit/RooStats stuff which are always memberwise-serialized and it's a good test bench to try and get uproot to read those out correctly.

jpivarski:
I agree; we'd need some concerted effort from someone who can concentrate on just this for at least a month. Knowing that you have a lot of sample files is a good thing.

kratsg:
Even pedantically worse -- ROOT generates them in different formats pre 6.18 and post 6.20 -- where they're not backward-compatible readable even within ROOT... so it will be interesting to see if uproot can bridge this gap.

PR #316, opened 2021-03-31
Cleaning up Uproot

jpivarski:
@henryiii I'm starting my work on Uproot with a clean-up. One of the major items here will be to get Uproot flake8 complaint again and another will be to set up a pre-commit hook like the one in Vector. (I'm using Vector as a model of how to do these things because it's also a pure Python project, unlike Awkward.)

Things that I'm planning in this PR:

   - [x] Move the requirements*.txt files into setup.py because they had been based on a misunderstanding about what they were for (which I don't entirely remember now, only that I've been meaning to do this).
   - [ ] Make Uproot flake8 complaint again.
   - [ ] Start using pre-commit.
   - [ ] Start using isort. (I'm not planning on typing it with MyPy; that's too big of a task and remember that Uproot has to support Python 2.)
   - [ ] Start using the version number-generator.
   - [x] Move the `uproot` directory to `src/uproot`. (For my incremental development, I put a dummy version file in the directory and a symlink from the base repo into `src/uproot`. It's not a clean test but that's not what I need during incremental development. It was in the Vector work that I found this solution; typing `pip install .` slows down the debugging loop too much.)


henryiii:
> I found this solution; typing pip install . slows down the debugging loop 

For pure python packages, please use `pip install -e .`; this installs in "editable" mode, which basically means it makes a symlink to the directory, but _also_ supports all the niceties that require an installed package, like the version file from setuptools_scm, importlib.metadata, importlib.resouces, etc.

The version number won't update until you rerun this line, and if you have compiled components, they also don't update until you run this line, but otherwise, it's how one develops with packages, especially `/src` packages.

jpivarski:
I'll give `pip install -e .` a try. That would have to be after I move `uproot` to `src/uproot`, though.

henryiii:
This is also the main reason to keep `setup.py`, as pip still requires a setup.py to enable editable mode, since it calls `python setup.py develop` somewhere. Eventually editable mode will make it into PEP 517 hopefully. Poetry and flit each have a special method to activate editable mode, since they don't have a `setup.py` to call develop on.

You can use editable mode on a non-src package, though then it's a little arbitrary on whether you get the "package" version or the "local" version - in general, it _mostly_ doesn't matter. Except when it does.

jpivarski:
I bumped up the `uproot → src/uproot` task to try out `pip install -e .`. I was expecting that it would be an additional step before every debugging step, but one time sets up a link/not copy permanently (even after closing and opening a new terminal). So this is a good way to work and I'll use it from now on.

Since I moved all the files, I'll be closing this PR and doing the rest of the work in another PR. That way, the "move files" commit is kept separate from the "change file contents" commit and I can keep using the "Squash and merge" button.

setup.cfg
```diff
@@ -37,23 +38,30 @@ classifiers =
 
 
 [options]
-python_requires = >=2.6, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
 packages = find:
 install_requires =
     numpy
+python_requires = >=2.6, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
```
henryiii, line 44:
```suggestion
python_requires = >=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
```

This doesn't work as written, you picked it up from an incorrect example I wrote, I think.

setup.cfg
```diff
@@ -37,23 +38,30 @@ classifiers =
...
+    xrootd
 
+[tool:isort]
+profile = black
+multi_line_output = 3
 
 [flake8]
 ignore = E203, W503, E501, E266, N
 max-complexity = 100
 exclude = tests, setup.py, uproot/__init__.py, docs-sphinx
```
henryiii, line 67:
```suggestion
exclude = tests, setup.py, src/uproot/__init__.py, docs-sphinx
```

Perhaps?

PR #318, opened 2021-03-31
Draft: nested memberwise

kratsg:
Maybe could use some extra eyes from @nsmith- . I'm using a HistFactory workspace in a ROOT file as an example. I can pass along the ROOT file for sure so you can look into it.. still trying to work out how to handle the reading correctly for this sort of file where there's a more complicated memberwise structure buried inside.


kratsg:
See #314.

Issue #333, opened 2021-04-16
Network access sources leak threads (and maybe also memory)


nikoladze:
In different tests that run over multiple files, using both xrootd and http access sources i'm seeing an increase in the number of threads. Also it seems sometimes (in http and also in xrootd even after the fix from https://github.com/xrootd/xrootd/pull/1440) there might be a slight memory leak. On the one hand, vmem increases proportional to the increasing number of threads (don't know if that means anything). But on the other hand i also sometimes see an increase of RSS/PSS memory.

One reproducer i could come up with is the following script:

```python
#!/usr/bin/env python

import os
import uproot

base_url = "https://atlas-opendata.web.cern.ch/atlas-opendata/samples/2020/4lep/MC/"

filenames = [
    "mc_364100.Zmumu_PTV0_70_CVetoBVeto.4lep.root",
    "mc_364101.Zmumu_PTV0_70_CFilterBVeto.4lep.root",
    "mc_364102.Zmumu_PTV0_70_BFilter.4lep.root",
    "mc_364103.Zmumu_PTV70_140_CVetoBVeto.4lep.root",
    "mc_364104.Zmumu_PTV70_140_CFilterBVeto.4lep.root",
    "mc_364105.Zmumu_PTV70_140_BFilter.4lep.root",
    "mc_364106.Zmumu_PTV140_280_CVetoBVeto.4lep.root",
    "mc_364107.Zmumu_PTV140_280_CFilterBVeto.4lep.root",
    "mc_364108.Zmumu_PTV140_280_BFilter.4lep.root",
    "mc_364109.Zmumu_PTV280_500_CVetoBVeto.4lep.root",
    "mc_364110.Zmumu_PTV280_500_CFilterBVeto.4lep.root",
    "mc_364111.Zmumu_PTV280_500_BFilter.4lep.root",
    "mc_364112.Zmumu_PTV500_1000.4lep.root",
    "mc_364113.Zmumu_PTV1000_E_CMS.4lep.root",
    "mc_364114.Zee_PTV0_70_CVetoBVeto.4lep.root",
    "mc_364115.Zee_PTV0_70_CFilterBVeto.4lep.root",
    "mc_364116.Zee_PTV0_70_BFilter.4lep.root",
    "mc_364117.Zee_PTV70_140_CVetoBVeto.4lep.root",
    "mc_364118.Zee_PTV70_140_CFilterBVeto.4lep.root",
[...]

PR #334, opened 2021-04-16
Stop leaking threads in HTTP (through the fallback mechanism).

jpivarski:


jpivarski:
See #333: this fixes the thread leak and the memory leak.

It only applies to HTTP servers that don't support multi-part GET, since the threads and Python objects being leaked were the fallback MultithreadedHTTPSource.

PR #339, opened 2021-04-18
Add support for basic HTTP authentication

cozzyd:
`ROOT`'s `TFile::Open` supports transparent access not just to HTTP, but also HTTP with basic authentication, by embedding the username and password in the URL, e.g. `http(s)://username:password@example.org/path/to/file.root`. This provides a some level of security to your precious data which you may not want to make public (yet?). In practice, hosting files this way with some credentials widely known to a collaboration is used a fair amount to expose data in academic clusters. 

This pull request adds equivalent functionality to uproot. `urllib`'s parser already parses the username and password from an URL, so we can grab the credentials from the parsed_url when available and append them to the headers. Note that this is only secure at all when using HTTP, as otherwise any interloper can snoop the base64-encoded username and password. I didn't implement it this way, but it may be preferable to not allow basic auth on non-HTTPS connections. 

I have lightly tested this, and I assume I haven't hit all the possible corner cases (like, I assume either the normal or the fallback mode worked for me, but not both). 

Here is what a test might look like, although you probably shouldn't use my URL's because I can't guarantee that they'll be available for eternity:

```
import uproot

def test():
  # test file with no auth needed
  with uproot.open('https://users.rcc.uchicago.edu/~cozzyd/uproot-test/not-so-secret-sauce/mundane.root') as f:
    assert ( f['hrandom'].all_members['fTitle'] =='Random Numbers')
  # test file with auth needed
  with uproot.open('https://DarkHelmet:12345@users.rcc.uchicago.edu/~cozzyd/uproot-test/secret-sauce/top-secret.root') as f:
    assert (f['hmeaning'].all_members['fTitle']=='Meaning of Life') 
```

I have only tested this on Python3... it's possible it may not work on Python2 due to the differences in strings (I suspect the line that does the base64 encoding may fail on Python2, but I'm not sure how to make it compatible with both, or if what I do on that line is even sane).

### DIFF
--- a/src/uproot/reading.py
+++ b/src/uproot/reading.py
@@ -42,7 +42,7 @@ def open(
             rather than a file. Path objects are interpreted strictly as
             filesystem paths or URLs.
             Examples: ``"rel/file.root"``, ``"C:\\abs\\file.root"``,
-            ``"http://where/what.root"``, 
+            ``"http://where/what.root"``,
             ``"https://username:password@where/secure.root"``,
             ``"rel/file.root:tdirectory/ttree"``, ``Path("rel:/file.root")``,
             ``Path("/abs/path:stuff.root")``
@@ -117,10 +117,10 @@ def open(
       ``TTrees``.
 
     For remote ROOT files served over HTTP(S), basic authentication is supported.
-    In this case, the credentials may be provided part of the URL in, as in 
-    ``https://username:password@example.org/secure/protected.root.`` Note that 
+    In this case, the credentials may be provided part of the URL in, as in
+    ``https://username:password@example.org/secure/protected.root.`` Note that
     for security reasons, it is recommended basic authentication only be used
-    for HTTPS resources. 
+    for HTTPS resources.
     """
     if isinstance(path, dict) and len(path) == 1:
         ((file_path, object_path),) = path.items()
--- a/src/uproot/reading.py
+++ b/src/uproot/reading.py
@@ -42,8 +42,10 @@ def open(
             rather than a file. Path objects are interpreted strictly as
             filesystem paths or URLs.
             Examples: ``"rel/file.root"``, ``"C:\\abs\\file.root"``,
-            ``"http://where/what.root"``, ``"rel/file.root:tdirectory/ttree"``,
-            ``Path("rel:/file.root")``, ``Path("/abs/path:stuff.root")``
+            ``"http://where/what.root"``, 
+            ``"https://username:password@where/secure.root"``,
+            ``"rel/file.root:tdirectory/ttree"``, ``Path("rel:/file.root")``,
+            ``Path("/abs/path:stuff.root")``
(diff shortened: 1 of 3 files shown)


jpivarski:
Thank you for this! The test error is easy to fix:

```
src/uproot/source/http.py:90:28: E711 comparison to None should be 'if cond is None:'
```

I see that you're parsing the URL to get the username; isn't that something that urllib.parse would do for you? (That parsing has already been done earlier in uproot/rrading.py or uproot/_util.py, but it can be done again.)

I don't see how the password/secret is passed in. It's not part of the URL, naturally. Should it (and the username itself, optionally) be passed in as `uproot.open` options? That's an open-ended set of keyword options whose defaults are set in uproot/reading.py, but they're passed through into the Source constructors.

cozzyd:
Hi Jim,

Thanks for the quick response! 

Yes that error is easy to fix, but I'm more concerned about the build error on 2.7 (apparently the combined_dict = {**dict1,**dict2} trick was only introduced in python3.5 :( ). I'll try to fix these later today. 

I'm not adding any extra parsing step here, urllib is already parsing the username and password (see the table in https://docs.python.org/3/library/urllib.parse.html#url-parsing ) from the URL, since apparently the http(s)://user:password@website format is somewhat standard.  I'm just checking to see if the credentials are present in the parser result (by comparing username and password to None), and if so, creating the appropriate header and including it with every HTTP request. 

It might be a good idea to support passing it in outside of the URL as that might be more obvious than putting it in the URL, but putting it in the URL lets you treat local and HTTP data directories uniformly. Perhaps providing a helper method to craft the URL from a URL/username/password may be a better interface?

jpivarski:
I didn't realize that the password was coming in through the URL, too. At least everything is handled in a standard way (and I'm glad that it's the urllib parser that's doing it—note that the module name was different in Python 2.7, but you might have already noticed that).

As for the syntax error in Python 2.7, this would work:

```python
dict(first, **second)
```

The syntax for this has just been getting incrementally better.

It could also be useful to add a prominent note about this new feature in the `uproot.open` docstring, since it would likely go unnoticed otherwise. We Want people to know about this!

cozzyd:
Hi Jim,

It looks like my latest attempt has passed all checks. I'm also willing to write a test, although I would need to find a semi-permanent URL I can use that uses basic authentication. 

Thanks,
-Cosmin

jpivarski:
This looks great! I have read it over carefully and there's nothing I'd ask you to change. It would be better to have a test, but this is one of those circumstances in which we're constrained by what's available. I don't know how to set up an HTTPS service with password authentication and such a thing would have to be maintained as long as Uproot is maintained, which would be hard to do. It's already saying something that it _has_ been tested (on `users.rcc.uchicago.edu`), which is a huge step above not having tested it at all.

Thanks for contributing!

@all-contributors please add @cozzyd for code

PR #352, opened 2021-05-05
Make TH* and TGraph* models as a step toward making them writable.

jpivarski:


matthewfeickert:
As a heads up to myself (and to make this easier to track in the future when the next `uproot` release comes out) it looks from [`pyhf`'s nightly CI using the head of our dependencies](https://github.com/scikit-hep/pyhf/actions/workflows/dependencies-head.yml) that this breaks (the `uproot4` CI [broke on 2021-05-05](https://github.com/scikit-hep/pyhf/actions/runs/815279416))

- A breaking test: [all of `tests/test_mixins.py` due to `pyhf.readxml.parse` failing](https://github.com/scikit-hep/pyhf/blob/a38e7410b7dcf2d5b3c66b03beef85468dee04ff/tests/test_mixins.py#L6-L14) on [`hist.to_numpy()[0].tolist()`](https://github.com/scikit-hep/pyhf/blob/a38e7410b7dcf2d5b3c66b03beef85468dee04ff/src/pyhf/readxml.py#L75)

```pytb
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/pyhf/readxml.py:349: in parse
    channel, data, samples, channel_parameter_configs = process_channel(
src/pyhf/readxml.py:228: in process_channel
    parsed_data = process_data(data[0], rootdir, inputfile, histopath)
src/pyhf/readxml.py:212: in process_data
    data, _ = import_root_histogram(rootdir, inputfile, histopath, histoname)
src/pyhf/readxml.py:75: in import_root_histogram
    return hist.to_numpy()[0].tolist(), extract_error(hist)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <TH1F (version 2) at 0x7fd23819ecd0>, flow = False, dd = False

    def to_numpy(self, flow=False, dd=False):
        """
        Args:
            flow (bool): If True, include underflow and overflow bins; otherwise,
                only normal (finite-width) bins are included.
            dd (bool): If True, the return type follows
                `numpy.histogramdd <https://numpy.org/doc/stable/reference/generated/numpy.histogramdd.html>`__;
                otherwise, it follows `numpy.histogram <https://numpy.org/doc/stable/reference/generated/numpy.histogram.html>`__
                and `numpy.histogram2d <https://numpy.org/doc/stable/reference/generated/numpy.histogram2d.html>`__.
    
        Converts the histogram into a form like the ones produced by the NumPy
        histogram functions.
        """
        values = self.values(flow=flow)
>       xedges = self.axis(0).edges(flow=flow)
E       AttributeError: 'Model_TAxis_v10' object has no attribute 'edges'
```

jpivarski:
For the record, the above is fixed in #363.

PR #358, opened 2021-05-13
Conditionally apply XRootD workarounds based on it's version

chrisburr:
Now that there are versions assigned for the two XRootD workarounds I think it would be better to apply them conditionally. There are a few ways of doing this but I think using `pkg_resources` is the simplest while maintaining support for Python 2.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,8 +40,8 @@ download_url = https://github.com/scikit-hep/uproot4/releases
 [options]
 packages = find:
 install_requires =
-    setuptools
     numpy
+    setuptools
 python_requires = >=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
 package_dir =
     =src
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,6 +40,7 @@ download_url = https://github.com/scikit-hep/uproot4/releases
 [options]
 packages = find:
 install_requires =
+    setuptools
     numpy
 python_requires = >=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
 package_dir =
--- a/src/uproot/extras.py
+++ b/src/uproot/extras.py
@@ -10,6 +10,9 @@ error messages containing instructions on how to install the library.
 
 from __future__ import absolute_import
 
+import atexit
+import pkg_resources
+from distutils.version import LooseVersion
 import os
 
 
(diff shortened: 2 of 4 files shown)


jpivarski:
Does `pkg_resources.get_distribution("XRootD")` actually import XRootD? Because if it does, then this change means that importing Uproot will always import XRootD on startup, even if a user isn't using it.

Even if it doesn't, changing the default of an input parameter based on another package feels a bit too magical—it's the sort of surprise that could lead to another bug, especially of the "works on my computer, not on yours" variety. (Such a bug might not be XRootD-related, further confusing the issue.)

Instead of changing `uproot.open`'s default, could the default always be XRootDSource and when the XRootDSource is starting (at the end of its `__init__`), old versions of XRootD make it immediately turn on its fallback mechanism? The effect would be the same: old versions of XRootD don't attempt to use vector_read, new versions try it and fall back if necessary, but both of these would use the same mechanism and `uproot.open`'s defaults would be constants.

chrisburr:
> Does `pkg_resources.get_distribution("XRootD")` actually import XRootD?

It doesn't import XRootD.

> Even if it doesn't, changing the default of an input parameter based on another package feels a bit too magical

I see what you mean though I though the bug this is avoiding is a relatively small memory leak so it's not really possible to "fallback" to avoid the problem. I see `XRootDSource` being between 2-10 times faster than `MultithreadedXRootDSource` (on real workloads, often at the higher end of that range). The leak is small enough to ignore so I really wouldn't want the vector-reading based approach to be unavailable (plus XRootD 5.2.0 is still a release candidate so I can't upgrade yet).

I don't see the default changing as being that magical. It's no different to using more threads on machines with more cores or using runtime dispatch to use vector instructions on some CPUs.

This is a bit opinionated but personally I'd be tempted to always default to `XRootDSource` as the speed difference feels more dramatic than the memory leak and the bug isn't an uproot issue. If somebody is doing something that is long running they can change the default to be `MultithreadedXRootDSource` and take the performance hit or petition the XRootD devs to make a release so they can upgrade. I know you dislike them but this might be case where a warning might be useful after XRootD 5.2 is out as it's hard for a user to understand either a) why memory is leaking or b) why uproot is so slow compared to reading local files or using plain old ROOT (if they realise at all).

I'll let you decide how you want to handle it and update the PR accordingly.

jpivarski:
My mistake! The XRootDSource doesn't have a `_fallback()` method the way that HTTPSource does.

In that case, I'd say that XRootDSource should be the default and if the version is old (check in `extras.py` when we know sometime is actually using XRootD), we should raise a warning, saying that the bad version of XRootD is deprecated and will be excluded in Uproot 4.1.0. (I like temporary warnings for deprecations; it should be `FutureWarning`, since `DeprecationWarning` is hidden.)

Is the new version of XRootD accessible in pip as well as conda? Is like to give an estimate for when we can start excluding the old version. (Uproot does have minimum versions for third party packages, even runtime requirements.)

@nsmith- should be involved in this discussion, since he advocated the default change in the other direction. Is the problem you saw fixed?

chrisburr:
> Is the new version of XRootD accessible in pip as well as conda?

It's not available anywhere yet though I believe updating PyPI is part of there release process so it should be available promptly once the final 5.2.0 release is available. I would be tempted to leave it as a warning for the foreseeable future as I suspect people are slow to upgrade (the XRootD 4.x series is still widely used).

Maybe it is best to split this PR into two parts, one which conditionally applies the `atexit` handler and another which raises the `FutureWarning` and can wait until XRootD 5.2.0 is released on PyPI and conda-forge. I can ask on Monday to see if anyone knows how long this will take, I suspect it won't be that long looking at past release cycles.

jpivarski:
I've converted the version-dependent default into a warning, and that warning tells users to either upgrade or manually change the default. That warning would probably get annoying, but XRootD 5.2.0 is available on both [PyPI](https://pypi.org/project/xrootd/) and [conda-forge](https://anaconda.org/conda-forge/xrootd) now, so most users would be able to do the right thing and upgrade.

If you're okay with this, I can merge it and put it in Uproot [4.0.8](https://github.com/scikit-hep/uproot4/releases/tag/untagged-5ea6c0581ff469b1becf).

Let me know; thanks!

(I'll go fix whatever that error is.)

nsmith-:
>relatively small memory leak

I'm confused, with current version of xrootd, 100% of the bytes read leak. How is this small?

Regarding pickling, I though I didn't pickle the pyxrootd objects, only the URL with which they were instantiated?

jpivarski:
On pickling, it was attempting to pickle the XRootDResource, and that one hadn't been pickle-safe. I just made it so. Perhaps you intended the XRootDResources to not get pickled, but they are. Where we decide to draw the line between what is pickleable and what isn't is not too important.

On the "relatively small" memory leak, there's some miscommunication, but—the bug is fixed now, right? Small or large, is it true that there's no leak now?

nsmith-:
You're too fast! Maybe add a test for the alternative source alongside the default in https://github.com/scikit-hep/uproot4/blob/main/tests/test_0302-pickle.py#L60

jpivarski:
Well, I assumed that the test was everything you wanted to test and made it work again. (The test you're pointing to is the one that fails; the code path gets to the c272886 addition.) Besides, the change to XRootDResource exactly mirrors your `__getstate__`/`__setstate__` in XRootDSource. It's just extending that capability in a very straightforward way.

nsmith-:
Yes the test I put together in #294 now shows no large increase in RSS over the duration of reading:
Screen Shot 2021-05-20 at 3 34 10 PM

chrisburr:
@jpivarski Thanks for finishing this off, I think this is good to merge now that XRootD is updated (you must have checked a only few minutes after the conda-forge binary hit the CDN).

(Also thanks @nsmith- for verifying that the bug is actually fixed!)

PR #363, opened 2021-05-17
Explicit models need explicit behaviors.

jpivarski:
This addresses the issue raised in https://github.com/scikit-hep/uproot4/pull/352#issuecomment-841758979

I had forgotten that the mechanism that discovers behavioral mix-ins and applies them to classes only does so if the models are generated from streamers. If they're explicitly defined, then I can make the inheritance manually and statically. (It's more flexible and less magical, but I have to remember to do it!)

@matthewfeickert, can you let me know if this fixes your CI?

### DIFF
--- a/src/uproot/behaviors/TGraph.py
+++ b/src/uproot/behaviors/TGraph.py
@@ -1,0 +1,15 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot4/blob/main/LICENSE
+
+"""
+This module defines the behaviors of ``TGraph``.
+"""
+
+from __future__ import absolute_import
+
+
+class TGraph(object):
+    """
+    Behaviors for TGraph: get values as NumPy arrays.
+    """
+
+    # TO DO!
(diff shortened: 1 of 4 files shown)


matthewfeickert:
> can you let me know if this fixes your CI?

Yeah, let me setup a custom branch real quick that targets this branch instead of `master` for `uproot4` and then I'll trigger the CI.

matthewfeickert:
Okay this is running now on 

https://github.com/scikit-hep/pyhf/runs/2603396161?check_suite_focus=true

using

```yaml
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip setuptools wheel
        python -m pip --no-cache-dir --quiet install --upgrade --editable .[test]
        python -m pip uninstall --yes uproot
        python -m pip install --upgrade git+git://github.com/scikit-hep/uproot4.git@jpivarski/explicit-models-need-explicit-behaviors
```

matthewfeickert (approved):
@jpivarski Yup that passes for us now. Thanks!

PR #388, opened 2021-07-06
Avoid warning when XRootD is installed but not used.

jpivarski:


HDembinski:
I can confirm that this patch removes the warning for me, thanks for the super quick response!

jpivarski:
Okay, great! I'll merge it into main now.

PR #395, opened 2021-07-14
Treat self-built XRootD version strings

### LINKED ISSUE #394: Uproot fails with self-built XRootD
With Uproot 4.0.11 and a self-built XRootD with Python bindings, `import uproot` fails with `TypeError: '<' not supported between instances of 'str' and 'int'`. This is due to the following line: https://github.com/scikit-hep/uproot4/blob/d6f9bea0f1a9ca6806445b95da93efa37c5117ba/src/uproot/extras.py#L116

When one builds XRootD, the version number differs from the standard `x.y.z` - it is, e.g., `v20210712-58b374f12`, which causes `LooseVersion` to raise `TypeError`.

dcervenkov:
When one builds XRootD, the version number differs from the standard `x.y.z` - it is, e.g., `v20210712-58b374f12`, which causes `LooseVersion` to raise `TypeError`.

Fixes #394

### DIFF
--- a/src/uproot/extras.py
+++ b/src/uproot/extras.py
@@ -106,9 +106,10 @@ def XRootD_client():
 def older_xrootd(min_version):
     """
     Check if the installed XRootD bindings are newer than a given version
-    without importing. Defaults to False if XRootD is not installed. Self-built
-    XRootD has different version strings, starting with 'v'; False is returned
-    in this case.
+    without importing. Defaults to False if XRootD is not installed. Unrecognized
+    versions (i.e. self-built XRootD, whose version numbers are strings)
+    return False: that is, they're assumed to be new, so that no warnings
+    are raised.
     """
     try:
         dist = pkg_resources.get_distribution("XRootD")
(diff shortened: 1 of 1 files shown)


src/uproot/extras.py
```diff
@@ -106,14 +106,19 @@ def cleanup_open_files():
...
     try:
         dist = pkg_resources.get_distribution("XRootD")
     except pkg_resources.DistributionNotFound:
         return False
     else:
-        return LooseVersion(dist.version) < LooseVersion(min_version)
+        if dist.version.startswith("v"):
+            return False
+        else:
+            return LooseVersion(dist.version) < LooseVersion(min_version)
```
chrisburr, line 121: How about  doing this to make sure now other unusual version strings cause issues?
```suggestion
        try:
            return LooseVersion(dist.version) < LooseVersion(min_version)
        except Exception:
            return False
```

jpivarski:
I bet I can change it through the web interface...

jpivarski:
Thanks, @dcervenkov and @chrisburr! If there are no complaints with this version, I'll be merging it once it passes tests.

dcervenkov:
Definitely agree that the `TypeError` exception is a better approach. Thanks for the quick reactions.

PR #396, opened 2021-07-16
mark test 0220 with network tag

bnavigator:
This test is missing the pytest network mark despite trying to access a webserver at cern.ch


jpivarski:
Thanks!

PR #407, opened 2021-08-07
Add support for reading TDatime

veprbl:


src/uproot/models/TDatime.py
```diff
@@ -0,0 +1,78 @@
...
+            members.append(("@instance_version", numpy.dtype(">u2")))
+        members.append(("fDatime", numpy.dtype(">u4")))
+        return uproot.interpretation.objects.AsStridedObjects(
+            cls, members, original=original
+        )
+
+    @classmethod
+    def awkward_form(
+        cls, file, index_format="i64", header=False, tobject_header=True, breadcrumbs=()
+    ):
```
veprbl, line 45: I wasn't able to figure out how to provide test coverage for this method.

jpivarski: It's okay to not provide coverage for this. It's going to be replaced (or perhaps just modified) in the AwkwardForth project, which is coming up soon. We can address these methods at that time.

jpivarski:
Should this be linked in with

https://github.com/scikit-hep/uproot4/blob/c1c95c0ad6b6f1fe830ed7187a71420d5fff913b/src/uproot/_util.py#L814-L839

which convert ROOT's "datime" integers into Python `datetime` objects?

veprbl:
> Should this be linked in with

For sure, but in what way? Modify during read, or in `Model.postprocess` or add as a method?

jpivarski:
> For sure, but in what way? Modify during read, or in `Model.postprocess` or add as a method?

The canonical way would be to add a `src/behaviors/TDatime.py` containing a `TDatime` class and have this `Model_TDatime` inherit from that behavior. The behavior would then define a

```python
    def to_datetime(self):
        return uproot._util.code_to_datetime(self._members["fDatime"])
```

method for `Model_TDatime` to inherit.

That's how the methods are usually named: histograms have `to_numpy()`, `to_hist()` (methods, not properties, to allow for future parameterization). The goal is not to rewrite the functionality of ROOT in Uproot, but to export objects to their Python equivalents (which generally provide the functionality of ROOT by themselves; for instance, `datetime` has all the methods for getting year, day, hours, isoformat, etc.).

jpivarski:
[scikit-hep-testdata 0.4.6](https://pypi.org/project/scikit-hep-testdata/) is now available. Tests that are looking for `uproot-issue-407.root` should find it.

veprbl:
Could someone trigger the tests? I believe this is ready.

jpivarski:
Sorry, I didn't realize that it was being held up.

jpivarski (approved):
Looks good to me! I'll set it to merge when the tests pass.

PR #418, opened 2021-08-24
Add TTable support

veprbl:
TTable was a class in ROOT, until removed in 6.16. It is an "alternative" to TTree implementing a row-based C-structure data storage with schema migration support.


jpivarski:
This looks very complete, but was it necessary to override the models? Can they not be read using streamers? If you want to add methods and properties to classes that _can_ be read generically, that's what behaviors are for. That is, if it's possible to provide a high-level method by calling low-level `self.member("fWhatever")`, then defining behaviors means that the reading code can still be auto-generated from streamers, which allows them to get new Uproot features (e.g. the planned AwkwardForth update) and new class versions in ROOT. On the other hand, if TTable was removed from ROOT, then there won't be any new versions, and that's not an issue.

The answer might be "no, it's not possible." I'm looking at the column-appending, and if that's custom (not just loading a bunch of TArrays, which can be described with streamers), then this is the only way and we should go ahead with it. Keep in mind that if there are multiple versions of this class floating around, defining the model means that only this version can be read (whereas defining a behavior would mix those methods in to any auto-generated version).

jpivarski:
Looks like you're done. Thanks! I'll include this in Uproot 4.1.1.

PR #428, opened 2021-08-30
Final tweaks, bug fixes, and performance checks for this stage of Uproot-writing

jpivarski:
Follows up on #426 with similar work for TTrees.

### DIFF
--- a/src/uproot/writing/_cascadetree.py
+++ b/src/uproot/writing/_cascadetree.py
@@ -381,6 +381,8 @@ class Tree(object):
             "fBasketSeek": numpy.zeros(
                 self._basket_capacity, uproot.models.TBranch._tbranch13_dtype3
             ),
+            "arrays_write_start": 0,
+            "arrays_write_stop": 0,
             "metadata_start": None,
             "basket_metadata_start": None,
             "tleaf_reference_number": None,
@@ -674,6 +676,7 @@ class Tree(object):
 
                 content = layout.content
                 offsets = numpy.asarray(layout.offsets)
+
                 if offsets[0] != 0:
                     content = content[offsets[0] :]
                     offsets = offsets - offsets[0]
@@ -757,6 +760,8 @@ class Tree(object):
                 fBasketEntry[i + 1] = num_entries + fBasketEntry[i]
 
             datum["fBasketSeek"][self._num_baskets] = location
+
+            datum["arrays_write_stop"] = self._num_baskets + 1
 
         # update TTree metadata in file
         self._num_entries += num_entries
@@ -1145,7 +1150,6 @@ class Tree(object):
                 self._metadata["fEstimate"],
             ),
         )
-        sink.flush()
 
         for datum in self._branch_data:
             if datum["kind"] == "record":
@@ -1183,16 +1187,33 @@ class Tree(object):
                 ),
             )
 
-            fBasketBytes = uproot._util.tobytes(datum["fBasketBytes"])
-            fBasketEntry = uproot._util.tobytes(datum["fBasketEntry"])
-            fBasketSeek = uproot._util.tobytes(datum["fBasketSeek"])
+            start, stop = datum["arrays_write_start"], datum["arrays_write_stop"]
+
+            fBasketBytes_part = uproot._util.tobytes(datum["fBasketBytes"][start:stop])
+            fBasketEntry_part = uproot._util.tobytes(
+                datum["fBasketEntry"][start : stop + 1]
+            )
+            fBasketSeek_part = uproot._util.tobytes(datum["fBasketSeek"][start:stop])
 
             position = base + datum["basket_metadata_start"] + 1
-            sink.write(position, fBasketBytes)
-            position += len(fBasketBytes) + 1
-            sink.write(position, fBasketEntry)
-            position += len(fBasketEntry) + 1
-            sink.write(position, fBasketSeek)
+            position += datum["fBasketBytes"][:start].nbytes
+            sink.write(position, fBasketBytes_part)
+            position += len(fBasketBytes_part)
+            position += datum["fBasketBytes"][stop:].nbytes
+
+            position += 1
+            position += datum["fBasketEntry"][:start].nbytes
+            sink.write(position, fBasketEntry_part)
+            position += len(fBasketEntry_part)
+            position += datum["fBasketEntry"][stop + 1 :].nbytes
+
+            position += 1
+            position += datum["fBasketSeek"][:start].nbytes
+            sink.write(position, fBasketSeek_part)
+            position += len(fBasketSeek_part)
+            position += datum["fBasketSeek"][stop:].nbytes
+
+            datum["arrays_write_start"] = datum["arrays_write_stop"]
 
             if datum["kind"] == "counter":
                 position = (
@@ -1208,6 +1229,8 @@ class Tree(object):
                         datum["tleaf_maximum_value"],
                     ),
                 )
+
+        sink.flush()
 
     def write_np_basket(self, sink, branch_name, compression, array):
         fClassName = uproot.serialization.string("TBasket")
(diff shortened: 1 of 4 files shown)


jpivarski:
Performance testing, just Uproot and C++ ROOT this time. Uses the same (singly) jagged dataset as https://arxiv.org/abs/2102.13516 .

**scaling-trees-uproot.py**

```python
import time

import numpy as np
import awkward as ak
import uproot

content = np.fromfile("/home/jpivarski/storage/data/chep-2021-jagged-jagged-jagged/sample-content.float32", dtype=np.float32)
offsets = np.fromfile("/home/jpivarski/storage/data/chep-2021-jagged-jagged-jagged/sample-offsets1.int64", dtype=np.int64)

array = ak.Array(ak.layout.ListOffsetArray64(ak.layout.Index64(offsets), ak.layout.NumpyArray(content)))

print(repr(array))
# <Array [[-0.423, 2.34, -0.757, ... 1.41, 0.72]] type='134217728 * var * float32'>

ENTRIES_PER_BASKET = 50000

starttime = time.time()

with uproot.recreate("/tmp/output.root", compression=None) as fout:
    fout.mktree("tree", {"branch": array.type})
    for start in range(0, len(array), ENTRIES_PER_BASKET):
        fout["tree"].extend({"branch": array[start : start + ENTRIES_PER_BASKET]})

print(f"entries per basket {ENTRIES_PER_BASKET}, time {time.time() - starttime}")
```

```
entries per basket 1000, time 61.869351863861084
entries per basket 1000000, time 15.633187770843506
entries per basket 2000, time 34.52807021141052
entries per basket 500000, time 14.124958038330078
entries per basket 5000, time 16.41567373275757
entries per basket 200000, time 13.47899842262268
entries per basket 10000, time 15.130658864974976
entries per basket 100000, time 13.349303483963013
entries per basket 20000, time 13.790864944458008
entries per basket 50000, time 13.850778579711914
```

**scaling-trees-ROOT.cpp**

```c++
#include <iostream>
#include <chrono>

#include "TFile.h"
#include "TTree.h"
#include "TBranch.h"

const int64_t ENTRIES_PER_BASKET = 50000;

int main(int argc, char** argv) {
  FILE* content_file = fopen("/home/jpivarski/storage/data/chep-2021-jagged-jagged-jagged/sample-content.float32", "r");
  FILE* offsets_file = fopen("/home/jpivarski/storage/data/chep-2021-jagged-jagged-jagged/sample-offsets1.int64", "r");

  float* content = new float[1073741824];
  int64_t* offsets = new int64_t[134217729];

  for (int64_t step = 0;  step < 1073741824;  step += 1024*1024) {
    fread(&content[step], sizeof(float), 1024*1024, content_file);
  }
  fread(offsets, sizeof(int64_t), 134217729, offsets_file);

  auto starttime = std::chrono::high_resolution_clock::now();

  auto f = new TFile("/tmp/output.root", "RECREATE");
  f->SetCompressionLevel(0);
  auto t = new TTree("tree", "");
  t->SetAutoFlush(0);
  t->SetAutoSave(0);

  uint32_t counter;
  float array[16];  // this is big enough

  t->Branch("nbranch", &counter, "nbranch/I", 100*1024*1024);      // temp TBasket must be bigger
  t->Branch("branch", array, "branch[nbranch]/F", 100*1024*1024);  // than written TBasket size

  for (int64_t i = 0;  i < 134217729 - 1;  i++) {
    int64_t start = offsets[i];
    int64_t stop = offsets[i + 1];
    counter = stop - start;

    for (int64_t j = start;  j < stop;  j++) {
      array[j - start] = content[j];
    }

    t->Fill();
    if (i % ENTRIES_PER_BASKET == ENTRIES_PER_BASKET - 1) {
      t->FlushBaskets();
    }
  }

  t->Write();
  f->Close();

  auto stoptime = std::chrono::high_resolution_clock::now();

  std::cout << "entries per basket " << ENTRIES_PER_BASKET << ", "
            << "time " << std::chrono::duration_cast<std::chrono::microseconds>(
                 stoptime - starttime
               ).count() / 1e6
            << std::endl;

  return 0;
}
```

```
entries per basket 1000, time 26.1168
entries per basket 1000000, time 15.4534
entries per basket 2000, time 16.5195
entries per basket 500000, time 15.6402
entries per basket 5000, time 14.8063
entries per basket 200000, time 15.2132
entries per basket 10000, time 15.4887
entries per basket 100000, time 15.0006
entries per basket 20000, time 14.5771
entries per basket 50000, time 14.309
```

Now the most significant effect is the number of entries per TBasket/`extend`/flush/cluster. For these data, 1 entry is very nearly 40 bytes, so 1000 entries is 40 kB and 1000000 entries is 40 MB.



So for _writing_ performance on par with C++ ROOT, you need to be writing at least 100 kB per branch in each call to `extend`. (The lines converge at 5000 entries per flush, which is 200 kB, but that's for 2 branches, and the data vs overhead scales with the number of branches, so that's why the rule is "100 kB per branch in each call to `extend`.)

src/uproot/writing/_cascadetree.py
```diff
@@ -668,9 +676,10 @@ def extend(self, file, sink, data):
 
                 content = layout.content
                 offsets = numpy.asarray(layout.offsets)
+
                 if offsets[0] != 0:
                     content = content[offsets[0] :]
-                    offsets -= offsets[0]
+                    offsets = offsets - offsets[0]
```
jpivarski, line 682: This is an important bug-fix. Don't let it get lost among the tweaks & tunes!

allenxiangxin: Basically, it improves speed by not writing small baskets. So is there a way to set the min size per tree, such as 
```
uproot.newtree(..., flushsize=100000)
```
And then do:
```
file.extend(..., flush=False)
```
And this way, keep stuff in memory and only write to file when the basket is full. So is this (https://indico.cern.ch/event/840667/contributions/3527109/attachments/1908764/3153297/uproot-irisfellow-final.pdf) out of date for uproot4 now?

allenxiangxin:
Basically, it improves speed by not writing small baskets. So is there a way to set the min size per tree, such as 
```
uproot.newtree(..., flushsize=100000)
```
And then do:
```
file.extend(..., flush=False)
```
And this way, keep stuff in memory and only write to file when the basket is full. So is this (https://indico.cern.ch/event/840667/contributions/3527109/attachments/1908764/3153297/uproot-irisfellow-final.pdf) out of date for uproot4 now?

jpivarski:
Uproot 3 is out of date now. Reik's IRIS-HEP presentation on Uproot TTree-writing was the Uproot 3 implementation. The Uproot 4 implementation (#321) now completely replaces it,

  * with a different public API (not much different, mostly [allowing assignment](https://uproot.readthedocs.io/en/latest/basic.html#writing-ttrees-to-a-file) and [uproot.WritableDirectory.mktree](https://uproot.readthedocs.io/en/latest/uproot.writing.writable.WritableDirectory.html#mktree) instead of introducing an `uproot.newtree` function),
  * with considerably reorganized internal mechanics for maintainability,
  * without flushing logic, which is what you're referring to above. All that Uproot 3's `flushsize` did (or attempted to do; it [sometimes didn't work](https://github.com/scikit-hep/uproot3/issues/522)) was gather array data until the limit is reached, then concatenate and write. That's easy enough to do in a script, but mixing that logic with the file-writing logic made quite a mess. So we also dropped `append`.

The downside of dropping the `flushsize` logic is that each [uproot.WritableTree.extend](https://uproot.readthedocs.io/en/latest/uproot.writing.writable.WritableTree.html#extend) is inexorably bound to one TBasket—no exceptions. Writing files with lots of little TBaskets is problematic. Hence, the new interface, if used unwisely, would lead users to create files with lots of little TBaskets. I just hope I've posted enough warnings to guard against that.

This PR doesn't seem to introduce any performance improvements, not like #426 for histograms. It's just where I tested it at scale and fixed a few bugs that were revealed by doing so.

PR #431, opened 2021-09-01
Recreate truncates if file exists

jrueb:
(As discussed before on Gitter)
Truncate the file to 0 bytes opened with uproot.recreate before writing contents. This behavior agrees with ROOT Recreate.
The previous code lines seemed to intend to do this, however assumingly due to an erroneous 'not', this was not done.
If the file does not exist, an empty file is created. This is the same behavior as before.


jpivarski:
I thought I had reversed the `not`, but it looks like I didn't. Anyway, this version would always do the right thing: whether the file had previously existed or not, you end up with an empty, existing file before any writing starts.

PR #432, opened 2021-09-01
Fix #430, global_index for tuples of DataFrames.

jpivarski:


jpivarski:
Windows Python 2.7 is the _only_ test that does not install Awkward Array, and it demonstrates that Uproot really can be used with nothin' but NumPy.

Now all the tests run in ~10 minutes, much better!

PR #441, opened 2021-09-16
Recursively fix forms for objects that require ak.from_iter

nsmith-:
If a sub-branch cannot be optimized by awkward, the parent cannot either and must have its form fixed for use with lazy arrays.


nsmith-:
Also by mistake I added the fix to `TRefArray` member `refs`, which should be a ListArray.

jpivarski:
Merging because all looks good and this is a blocker.

(Note for posterity: @nsmith- and I talked about these issues on Slack while he was discovering them.)

Issue #456, opened 2021-10-01
iteration blocks on exception from callback in XRootDResource


btovar:
When the callback() in XRootDResource.callbacker has an exception, then an iteration never returns.  We found out this by getting socket timeout errors from xrootd.

The simplest way I found to reproduce this is to change the callback [here](https://github.com/scikit-hep/uproot4/blob/03cb08ccb1f47ce0b082117f41a112af1ab403ee/src/uproot/source/xrootd.py#L232) to immediately "raise" an exception, as:

```python
        def callback(status, response, hosts):
            self._xrd_error(status)
            ...rest of code
```
and use this example:

```python
import uproot

source='root://eospublic.cern.ch//eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetDown.root'
for a in uproot.behaviors.TBranch.iterate(source, 'H3_isMuon'):
    print(a)

print("not printed if callback fails")
```

When there is an exception, it is printed to the console, but the script never finishes. 

This was tested with uproot 4.1.3 from a pip install.

jpivarski:
I was about to say that we'd need a reproducible test, but I guess we could add an intentional Exception to the callback, at least for a manual test. I don't know how we could do that in CI, but at least for coming up with a solution.

The HTTP Source passes exceptions on to the main thread. The same kind of thing should be possible here.

ioanaif:
From https://github.com/scikit-hep/uproot5/pull/480 it looks like this has been fixed and tested locally as it would be hard to catch this in the CI.

PR #480, opened 2021-10-20
Pass exceptions to main thread in xrootd

btovar:
Sketch of a solution for issue #456. Simply catch the exception on the
callback and update the futures accordingly (as HTTP source does).

### DIFF
--- a/src/uproot/source/xrootd.py
+++ b/src/uproot/source/xrootd.py
@@ -235,7 +235,7 @@ in file {1}""".format(
             if status.error:
                 try:
                     self._xrd_error(status)
-                except:
+                except Exception:
                     excinfo = sys.exc_info()
                     for future in futures.values():
                         future._set_excinfo(excinfo)


btovar:
Sounds good @jpivarski, it looks good to me.

I agree with the difficulty of the testing.  I tried it here setting the timeouts to low values to trigger Operation Expired errors. The current pr does recover from those errors, but it was hard to reliably test it.

jpivarski:
Okay, then I'll take it once pre-commit passes (https://results.pre-commit.ci/run/github/262422450/1634748035.QZ3W9c9lRrKb9ZLne5bcAQ). It's failing because you declared `e` (the exception object) and then didn't use it.

I'll turn on auto-merge, so that it merges once that's fixed.

jpivarski:
@all-contributors please add @btovar for code

btovar:
Great, thanks Jim.

PR #501, opened 2021-11-05
Detect XRootD versions before 4.11.1

### LINKED ISSUE #486: XRootD version detection doesn't work for versions before 4.11.1
There are two independent places where the XRootD version is detected:

https://github.com/scikit-hep/uproot4/blob/9607057b07b5936d6f2925e33ad6386a7dfc7705/src/uproot/extras.py#L121 and https://github.com/scikit-hep/uproot4/blob/9607057b07b5936d6f2925e33ad6386a7dfc7705/src/uproot/reading.py#L170

These both rely on `pkg_resources.get_distribution("XRootD")`, but the name of the Python module [changed from `pyxrootd` to `xrootd`](https://github.com/xrootd/xrootd/commit/dfa0806f02775dc04e7835e8fc871b81a2bd786f) in [v4.11.1](https://github.com/xrootd/xrootd/blob/v4.11.1/docs/ReleaseNotes.txt#L26). Critically, this means that with XRootD v4.11.0 and earlier, the workaround here (for the deadlock condition described in https://github.com/scikit-hep/uproot3/issues/504):

https://github.com/scikit-hep/uproot4/blob/9607057b07b5936d6f2925e33ad6386a7dfc7705/src/uproot/extras.py#L87-L107

doesn't get applied, and you don't get the warning in:

https://github.com/scikit-hep/uproot4/blob/9607057b07b5936d6f2925e33ad6386a7dfc7705/src/uproot/reading.py#L169-L178

masonproffitt:
Resolves #486. Checks for the package name `pyxrootd` if searching for `XRootD` fails.

### DIFF
--- a/src/uproot/extras.py
+++ b/src/uproot/extras.py
@@ -117,15 +117,29 @@ def older_xrootd(min_version):
     return False: that is, they're assumed to be new, so that no warnings
     are raised.
     """
-    try:
-        dist = pkg_resources.get_distribution("XRootD")
-    except pkg_resources.DistributionNotFound:
+    version = xrootd_version()
+    if version is None:
         return False
     else:
         try:
-            return LooseVersion(dist.version) < LooseVersion(min_version)
+            return LooseVersion(version) < LooseVersion(min_version)
         except TypeError:
             return False
+
+
+def xrootd_version():
+    """
+    Gets the XRootD version if installed, otherwise returns None.
+    """
+    try:
+        version = pkg_resources.get_distribution("XRootD").version
+    except pkg_resources.DistributionNotFound:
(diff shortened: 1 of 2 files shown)


jpivarski:
Let me know when you're done with this by requesting a review. Thanks!

masonproffitt:
Okay, I can't directly request a review, but it is done

jpivarski (approved):
I didn't realize it wouldn't let you request a review. But here's one: it looks great! Thanks for doing this and for consolidating the two places where the version is needed with a function call.

jpivarski:
@all-contributors please add @masonproffitt  for code

PR #506, opened 2021-11-12
Remove underscore prefixes from data members uproot.model.TTable

veprbl:
My original plan to "publish" them was to do so by adding `@property` accessors without the underscore, but to simply rename might be even better.

### DIFF
--- a/src/uproot/models/TTable.py
+++ b/src/uproot/models/TTable.py
@@ -246,6 +246,12 @@ in file {1}""".format(
         )
         self._data = numpy.frombuffer(buf, dtype=dtype)
 
+    @property
+    def data(self):
+        view = self._data.view()
+        view.flags.writeable = False
+        return view
+
     base_names_versions = []
     member_names = ["fN", "fMaxIndex", "fSize"]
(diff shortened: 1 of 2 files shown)


jpivarski:
The style everywhere else in the codebase is to use properties, so that they can be read-only or types can be checked upon assignment. By itself, what you have here looks fine, but it goes against the grain of everything else.

I only came to this style of "underscore + property for nearly everything" gradually, but Uproot (and Awkward Array) were written entirely in this style. It puts a lot of ceremony around class attributes, which I think is good because they should be kept at a minimum and mutability should be discouraged.

As for whether this particular class should have attributes, rather than properties, I'll let you decide, but I just want to let you know what the rest of the codebase is doing. (And anyway, Python lets you postpone that decision, since the public semantics are identical.)

jpivarski:
@veprbl Is this ready to merge? It looks like it is.

veprbl:
Yes, it is.

jpivarski:
Okay, then I'll merge it to get it in the next release (maybe tomorrow). Thanks!

PR #519, opened 2021-12-02
[MemmapSource] Remove unnecessary(?) copies

pfackeldey:
Dear `uproot4` developers,

this PR removes unnecessary(?) copies of the `numpy.memmap`. This should be safe, as `self._file` is only opened in `r`-mode, see: https://github.com/scikit-hep/uproot4/blob/8ae497a89ec92f06724980fccafa1c4286ce6040/src/uproot/source/file.py#L113 `uproot3` also did not do copies on the `numpy.memmaps`: https://github.com/scikit-hep/uproot3/blob/master/uproot3/source/memmap.py#L54-L57

Since this is a very fundamental piece of code I would really appreciate your feedback here!
(I hope I did not overlook a trivial reason why now a copy is needed.)

Thank you very much for reviewing!

Best, Peter

### DIFF
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -2540,6 +2540,12 @@ in file {3}""".format(
         Returns a :doc:`uproot.source.chunk.Chunk` and
         :doc:`uproot.source.cursor.Cursor` as a 2-tuple for a given
         ``basket_num``.
+
+        If the file source is :doc:`uproot.source.file.MemmapSource`
+        and the file gets closed, accessing the Chunk would cause
+        a segfault. If that's a possibility, be sure to call
+        :doc:`uproot.source.chunk.Chunk.detach_memmap` to ensure
+        that any memmap-derived data gets copied for safety.
         """
         if 0 <= basket_num < self._num_normal_baskets:
             start = self.member("fBasketSeek")[basket_num]
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -2616,16 +2616,21 @@ in file {3}""".format(
             start = self.member("fBasketSeek")[basket_num]
             stop = start + uproot.reading._key_format_big.size
             cursor = uproot.source.cursor.Cursor(start)
+
+            # Chunk will not be retained; we don't have to detach_memmap()
             chunk = self._file.source.chunk(start, stop)
+
             return uproot.reading.ReadOnlyKey(
                 chunk, cursor, {}, self._file, self, read_strings=False
             )
+
         elif 0 <= basket_num < self.num_baskets:
             raise ValueError(
                 "branch {0} basket {1} is an embedded basket, which has no TKey".format(
                     repr(self.name), basket_num
                 )
             )
+
         else:
             raise IndexError(
                 """branch {0} has {1} baskets; cannot get basket chunk {2}
--- a/src/uproot/models/TBasket.py
+++ b/src/uproot/models/TBasket.py
@@ -252,6 +252,9 @@ class Model_TBasket(uproot.model.Model):
             num_entries = self._members["fNevBuf"]
             key_length = self._members["fKeylen"]
 
+            # Embedded TBaskets are always uncompressed; be sure to copy any memmap arrays
+            chunk = chunk.detach_memmap()
+
             if maybe_entry_size * num_entries + key_length != self._members["fLast"]:
                 raw_byte_offsets = cursor.bytes(
                     chunk, 8 + self.num_entries * 4, context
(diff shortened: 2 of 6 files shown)


jpivarski:
Originally, MemmapSource didn't copy, but there was a problem: if uncopied data gets included in an object that continues to live after the file is closed, later accessing that object raises a segfault. Let me see if I can demo that...

(The test failure might be an example of that already.)

In Uproot's main branch:

```python
>>> file = uproot.open(skhep_testdata.data_path("uproot-Zmumu-uncompressed.root"))
>>> basket = file["events/px1"].basket(0)
>>> file.close()
>>> basket.data
array([192,  68, 152, ..., 241,  47, 230], dtype=uint8)
```

In this PR branch:

```python
>>> file = uproot.open(skhep_testdata.data_path("uproot-Zmumu-uncompressed.root"))
>>> basket = file["events/px1"].basket(0)
>>> file.close()
>>> basket.data
Segmentation fault (core dumped)
```

Note that it only segfaults if the data are not ever copied. So, for instance, if you ask for an array, there's a copy because the baskets get concatenated (still in the PR branch):

```python
>>> file = uproot.open(skhep_testdata.data_path("uproot-Zmumu-uncompressed.root"))
>>> array = file["events/px1"].array(library="np")  # NumPy to keep things simple
>>> file.close()
>>> array
array([-41.19528764,  35.11804977,  35.11804977, ...,  32.37749196,
        32.37749196,  32.48539387])
```

Also, there's a copy if you have to decompress something:

```python
>>> file = uproot.open(skhep_testdata.data_path("uproot-Zmumu-zlib.root"))
>>> basket = file["events/px1"].basket(0)
>>> file.close()
>>> basket.data
array([192,  68, 152, ..., 241,  47, 230], dtype=uint8)
```

Whether something is a copy or not is haphazard. It might be unusual for a regular user to ask for a basket, rather than an array, but you could also get the segfault if you ask for an uncompressed histogram—I think the bin contents are viewed from the original array, rather than copied. Randomly getting segfaults is no good.

Actually, I think I have an idea of how to make this better. I'm going to try committing to the PR branch.

pfackeldey:
Thank you very much for your detailed reply @jpivarski !
I did not think about accessing data from a file after closing it, I personally would then make a copy of that data before closing by myself. (This would feel the most natural way for me.)
Also this is how it also would be in plain `numpy`:
```python
In [1]: import numpy as np

In [2]: np.save("foo.npy", np.array([1,2,3,4]))

In [3]: data = np.memmap("foo.npy", dtype=np.uint8, mode="r")

In [4]: data
Out[4]:
memmap([147,  78,  85,  77,  80,  89,   1,   0, 118,   0, 123,  39, 100,
        101, 115,  99, 114,  39,  58,  32,  39,  60, 105,  56,  39,  44,
         32,  39, 102, 111, 114, 116, 114,  97, 110,  95, 111, 114, 100,
        101, 114,  39,  58,  32,  70,  97, 108, 115, 101,  44,  32,  39,
        115, 104,  97, 112, 101,  39,  58,  32,  40,  52,  44,  41,  44,
         32, 125,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,
         32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,
         32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,
         32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  32,
         32,  32,  32,  32,  32,  32,  32,  32,  32,  32,  10,   1,   0,
          0,   0,   0,   0,   0,   0,   2,   0,   0,   0,   0,   0,   0,
          0,   3,   0,   0,   0,   0,   0,   0,   0,   4,   0,   0,   0,
          0,   0,   0,   0], dtype=uint8)

In [5]: data._mmap.close()

In [6]: data
Out[6]: 'ipython' terminated by signal SIGSEGV (Address boundary error)
```
However a segementation fault is obviously the worst a user can get back. So I am very happy that you have already an idea! I'll keep an eye on your update :) 

Best, Peter

jpivarski:
If we let memmap arrays go uncopied, we now have to ensure that `detach_memmap()` is called on any Chunk that could be passed to the user. Fortunately, there aren't too many places where this could happen: if it's an object from a TDirectory (such as a histogram), we know whether it's compressed or not and call `detach_memmap()` if it's uncompressed. If it's a TBasket of a TTree, we find out whether it's compressed or not when constructing the TBasket object (`read_members`). The only other place where we might be asking for new Chunks is in RNTuple handling.

If a Chunk ever slips by and gets out of the scope in which it was created, even as an array sliced out of it by `cursor.bytes(chunk, ...)`, then it will be a subtle segfault bug that happens sometime after the file gets closed. We'd really like to have something like Rust's borrow checker to validate it! But, reading the code carefully and rereading it, I'm pretty sure this is right.

jpivarski:
@all-contributors please add @pfackeldey for code

PR #526, opened 2021-12-10
chore: drop Python 2 syntax

henryiii:
First commit is mostly automated, second commend is hand cleanup for things the first one didn't catch. Many `.format`'s are left, this is just the easy ones that clearly are simpler as f-strings. Manual comparisons with sys.version_info are always preferred, and can be automatically cleaned up, so I had to look around for the try/catch ones. I'll note some cleanups inline.

### DIFF
--- a/docs-sphinx/prepare_docstrings.py
+++ b/docs-sphinx/prepare_docstrings.py
@@ -76,8 +76,8 @@ main.write(
     :caption: Main Interface
     :hidden:
 
-{0}""".format(
-        "".join("    {0}\n".format(x) for x in common)
+{}""".format(
+        "".join(f"    {x}\n" for x in common)
     )
 )
 toctree = open("uproot.toctree", "w")
@@ -94,7 +94,7 @@ toctree2 = None
 def ensure(filename, content):
     overwrite = not os.path.exists(filename)
     if not overwrite:
-        overwrite = open(filename, "r").read() != content
+        overwrite = open(filename).read() != content
     if overwrite:
         open(filename, "w").write(content)
         sys.stderr.write(filename + " (OVERWRITTEN)\n")
@@ -190,7 +190,7 @@ def handle_class(classname, cls):
                     fill.append("")
                     if basecls is not cls:
                         fill.append(
-                            "Inherited from :doc:`{0}`.".format(
+                            "Inherited from :doc:`{}`.".format(
                                 basecls.__module__ + "." + basecls.__name__
                             )
                         )
@@ -213,7 +213,7 @@ def handle_class(classname, cls):
 
     fullfilename = importlib.import_module(cls.__module__).__file__
     shortfilename = fullfilename[fullfilename.rindex("uproot/") :]
-    link = "`{0} <https://github.com/scikit-hep/uproot4/blob/{1}/src/{2}>`__".format(
+    link = "`{} <https://github.com/scikit-hep/uproot4/blob/{}/src/{}>`__".format(
         cls.__module__, latest_commit, shortfilename
     )
     try:
@@ -232,30 +232,30 @@ def handle_class(classname, cls):
         inheritance_header = """.. table::
     :class: note
 
-    +-{0}-+
-    | **Inheritance order:** {1}|
-    +={2}=+
+    +-{}-+
+    | **Inheritance order:** {}|
+    +={}=+
     | """.format(
             "-" * longest_cell, " " * (longest_cell - 22), "=" * longest_cell
         )
         inheritance_footer = """ |
-    +-{0}-+""".format(
+    +-{}-+""".format(
             "-" * longest_cell
         )
         inheritance = [x + " " * (longest_cell - len(x)) for x in inheritance]
         inheritance_sep = """ |
     | """
 
-    content = """{0}
-{1}
+    content = """{}
+{}
 
-Defined in {2} on {3}.
+Defined in {} on {}.
 
-{4}
+{}
 
-.. autoclass:: {5}
+.. autoclass:: {}
 
-{6}
+{}
 """.format(
         title,
         "=" * len(title),
@@ -285,19 +285,19 @@ def handle_function(functionname, cls):
 
     fullfilename = importlib.import_module(cls.__module__).__file__
     shortfilename = fullfilename[fullfilename.rindex("uproot/") :]
-    link = "`{0} <https://github.com/scikit-hep/uproot4/blob/{1}/src/{2}>`__".format(
+    link = "`{} <https://github.com/scikit-hep/uproot4/blob/{}/src/{}>`__".format(
         cls.__module__, latest_commit, shortfilename
     )
     linelink = "`line {0} <https://github.com/scikit-hep/uproot4/blob/{1}/src/{2}#L{0}>`__".format(
         inspect.getsourcelines(cls)[1], latest_commit, shortfilename
     )
 
-    content = """{0}
-{1}
+    content = """{}
+{}
 
-Defined in {2} on {3}.
+Defined in {} on {}.
 
-.. autofunction:: {4}
+.. autofunction:: {}
 """.format(
         title, "=" * len(title), link, linelink, functionname
     )
@@ -317,7 +317,7 @@ for modulename in order:
         toctree2 = open(modulename + ".toctree", "w")
         toctree2.write(
             """.. toctree::
-    :caption: {0}
+    :caption: {}
     :hidden:
 
 """.format(
@@ -328,10 +328,8 @@ for modulename in order:
     handle_module(modulename, module)
     if module.__file__.endswith("__init__.py") and modulename != "uproot":
         for submodulename in sorted(
-            [
-                modulename + "." + name
-                for loader, name, is_pkg in pkgutil.walk_packages(module.__path__)
-            ]
+            modulename + "." + name
+            for loader, name, is_pkg in pkgutil.walk_packages(module.__path__)
         ):
             submodule = importlib.import_module(submodulename)
             handle_module(submodulename, submodule)
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -5,7 +5,6 @@ This module defines utilities for internal use. This is not a public interface
 and may be changed without notice.
 """
 
-from __future__ import absolute_import
 
 import datetime
 import glob
@@ -71,7 +70,7 @@ def ensure_str(x):
     elif isinstance(x, str):
         return x
     else:
-        raise TypeError("expected a string, not {0}".format(type(x)))
+        raise TypeError(f"expected a string, not {type(x)}")
 
 
 def ensure_numpy(array, types=(numpy.bool_, numpy.integer, numpy.floating)):
@@ -86,9 +85,7 @@ def ensure_numpy(array, types=(numpy.bool_, numpy.integer, numpy.floating)):
         except (ValueError, numpy.VisibleDeprecationWarning):
             raise TypeError("cannot be converted to a NumPy array")
         if not issubclass(out.dtype.type, types):
-            raise TypeError(
-                "cannot be converted to a NumPy array of type {0}".format(types)
-            )
+            raise TypeError(f"cannot be converted to a NumPy array of type {types}")
         return out
 
 
@@ -145,7 +142,7 @@ def regularize_filter(filter):
     else:
         raise TypeError(
             "filter must be None, callable, a regex string between slashes, or a "
-            "glob pattern, not {0}".format(repr(filter))
+            "glob pattern, not {}".format(repr(filter))
         )
 
 
@@ -214,7 +211,7 @@ def regularize_rename(rename):
 
     raise TypeError(
         "rename must be None, callable, a '/from/to/' regex, an iterable of "
-        "regex rules, or a dict, not {0}".format(repr(rename))
+        "regex rules, or a dict, not {}".format(repr(rename))
     )
 
 
@@ -355,7 +352,7 @@ def file_path_to_source_class(file_path, options):
         return out, file_path
 
     else:
-        raise ValueError("URI scheme not recognized: {0}".format(file_path))
+        raise ValueError(f"URI scheme not recognized: {file_path}")
 
 
 if isinstance(__builtins__, dict):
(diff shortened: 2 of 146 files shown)


src/uproot/_util.py
```diff
@@ -457,7 +452,7 @@ def memory_size(data, error_message=None):
     if error_message is None:
         raise TypeError(
             "number of bytes or memory size string with units "
-            "(such as '100 MB') required, not {0}".format(repr(data))
+            "(such as '100 MB') required, not {}".format(repr(data))
```
henryiii, line 455: PyUpgrade doesn't simplify things like this, but it should, IMO:
```suggestion
            f"(such as '100 MB') required, not {data!r}"
```

src/uproot/_util.py
```diff
@@ -520,7 +515,7 @@ def awkward_form(
                     '"float64"'
                 )
             else:
-                raise AssertionError("{0}: {1}".format(repr(model), type(model)))
+                raise AssertionError(f"{repr(model)}: {type(model)}")
```
henryiii, line 518:
```suggestion
                raise AssertionError(f"{model!r}: {type(model)}")
```

src/uproot/behaviors/TAxis.py
```diff
@@ -84,8 +83,7 @@ def __iter__(self):
             for x in fLabels:
                 yield str(x)
         else:
-            for low, high in self.intervals():
-                yield low, high
+            yield from self.intervals()
```
henryiii, line 86: :heart:

src/uproot/exceptions.py
```diff
@@ -63,25 +62,25 @@ def __str__(self):
...
+            in_object = f"\nin object {self.object_path}"
 
         if self.cycle == "any":
-            return """not found: {0} (with any cycle number){1}{2}{3}{4}""".format(
+            return """not found: {} (with any cycle number){}{}{}{}""".format(
                 repr(self.key), because, with_keys, in_file, in_object
             )
         elif self.cycle is None:
-            return """not found: {0}{1}{2}{3}{4}""".format(
+            return """not found: {}{}{}{}{}""".format(
```
henryiii, line 80: Something like this would be nicer nicer as an f-string too, but pyupgrade is too frugal.

src/uproot/behaviors/TAxis.py
jpivarski: That's clever! (Avoid unpacking and repacking that tuple, and the meaning is clearer, too. I haven't been using `yield from` for some reason that I don't remember, but it's certainly old enough to use it—it's not even a Python 3 thing.)

src/uproot/exceptions.py
jpivarski: If the expressions on the next line are long, it may be best as it is. As a `.format`, they don't have to maintain zero whitespace between expressions.

src/uproot/extras.py
```diff
@@ -23,21 +22,21 @@ def awkward():
...
+        raise ModuleNotFoundError(
             """install the 'awkward' package with:
 
     pip install awkward
 
 Alternatively, you can use ``library="np"`` or globally set ``uproot.default_library``
 to output as NumPy arrays, rather than Awkward arrays.
 """
-        )
+        ) from None
```
henryiii, line 34: This is still an open question, actually, didn't realize I put one in. Should we hide the original exception (like Python 2), or should it be chained on? I'm thinking hiding it _might_ be okay for these? It does reduce the burden on the user, but it can hide, for example, if a failure is due to some problem other than a simple installation. But in that rare case, then a user could just try importing it.

src/uproot/_util.py
jpivarski: These `!r` formatters were available in Python 2, also, but I hadn't been using them. I guess I should.

src/uproot/exceptions.py
henryiii: Yes, but if it fits in one line, then it's best in the original line (especially when they are all together like this). If there are more than a line, or if they have expressions other than just simple values, yes. (I haven't tried putting these in to see if it would overflow a line)

henryiii:
Okay, no problem. Will stick with this. :) I'm going to do a couple more changes (I should be able to detect and replace `{repr(x)}` with `{x!r}`). 

I think the main question: do you want to chain or hide the ModuleNotFoundErrors in extras?

henryiii:
I've also got a few more changes I want to annotate above, so I put it into draft

src/uproot/extras.py
jpivarski: I haven't been hiding any exceptions. The chained exception output is verbose, but the only important bit for a user (non-debugger) is at the end. I'd think Python users are used to only looking at the end of an exception stack trace.

jpivarski: Let's keep chaining exceptions. Changing that (if we do it at all) doesn't have to be done at the same time as this.

src/uproot/writing/_cascade.py
```diff
@@ -2003,7 +1996,7 @@ def __init__(
         self._begin = 100
 
     def __repr__(self):
-        return "{0}({1}, {2}, {3}, {4}, {5}, {6}, {7}, {8}, {9})".format(
+        return "{}({}, {}, {}, {}, {}, {}, {}, {}, {})".format(
```
henryiii, line 1999: This one is easier to read in this form vs. an f-string.

src/uproot/writing/_cascadetree.py
```diff
@@ -129,22 +128,18 @@ def __init__(
...
                         awkward = uproot.extras.awkward()
-                    except ImportError as err:
+                    except ModuleNotFoundError as err:
                         raise TypeError(
-                            "not a NumPy dtype and 'awkward' cannot be imported: {0}\n\n{1}".format(
-                                repr(branch_type), str(err)
-                            )
-                        )
+                            f"not a NumPy dtype and 'awkward' cannot be imported: {branch_type!r}"
+                        ) from err
```
henryiii, line 134: Here's the biggest manual change. Since Python 3 allows exception chaining (in fact, it's the default, though I made it explicit here), the original exception is already present, so I removed appending it manually. Also this has always been the wrong exception, you should not manually throw a ImportError, but it's subclass, ModuleNotFoundError, in Python 3.

src/uproot/extras.py
henryiii: Python 2 does not chain exceptions. For Python 2, you would never see the exception that caused the one you throw. Python 3 does, and does it by default. So the question would be would this be cleaner restoring the Python 2 output, or should we embrace the Python 3 output? :)

Yes, I can remove it, didn't realize I changed one. I started to change it, then decided to ask, and seems like I left this one in.

```suggestion
        )
```

src/uproot/behaviors/TAxis.py
henryiii: It is very much a Python 3.3 thing. :) 3.3 happens to be over 9 years old, but it't newer than 2.7!

src/uproot/_util.py
henryiii: They are more important for readability when you have to nest the `repr()` inside the brackets, but yes, they are likely still slightly nicer for `.format` - you can use them to keep the formatting part (showing a repr) separate from the data part.

src/uproot/behaviors/TAxis.py
jpivarski: Oh, I thought it was in Python 2. Well, if it started in Python 3.3 (9 years ago), then that's why I hadn't been using it, but can now.

src/uproot/extras.py
jpivarski: I know: only Python 3 shows the chained stack trace, so there had been a difference between how it looked in Python 2 and 3. As a user of Python 3, I preferred the chains, and I'm glad they're the default.

By "not changing this," I mean keep the chained exceptions: i.e. not change the Python 3 behavior.

src/uproot/writing/_cascadetree.py
jpivarski: Okay. And it can be ModuleNotFoundError now. (I'm pretty sure that word didn't exist in Python 2, which is why we were catching the superclass.)

src/uproot/writing/_cascade.py
jpivarski: I agree.

src/uproot/_util.py
henryiii: I used `git ls-files | sed -i 's/{repr(\([^)]*\))}/{\1!r}/g' to clean all of these up.

docs-sphinx/prepare_docstrings.py
```diff
@@ -328,10 +328,8 @@ def handle_function(functionname, cls):
     handle_module(modulename, module)
     if module.__file__.endswith("__init__.py") and modulename != "uproot":
         for submodulename in sorted(
-            [
-                modulename + "." + name
-                for loader, name, is_pkg in pkgutil.walk_packages(module.__path__)
-            ]
+            modulename + "." + name
+            for loader, name, is_pkg in pkgutil.walk_packages(module.__path__)
```
henryiii, line 332: These sorts of fixes are not really always python upgrade syntax related, but welcome. It no longer has to construct an extra temporary list before running the sort, and removes brackets too, a win-win. ;)

henryiii: I'd actually prefer `f"{modulename}.{name}"`, faster and I think it reads better than using a +, but okay.

src/uproot/interpretation/library.py
```diff
@@ -611,18 +608,13 @@ def group(self, arrays, expression_context, how):
         elif how is list:
             return [arrays[name] for name, _ in expression_context]
         elif how is dict:
-            return dict(
-                (_rename(name, c), arrays[name]) for name, c in expression_context
-            )
+            return {_rename(name, c): arrays[name] for name, c in expression_context}
```
henryiii, line 611: This was available in 2.7+, IIRC.

jpivarski: But not in 2.6. That's also why the format strings didn't have any `{}` in them (also a 2.7+ feature). None of that matters now, though.

tests/test_0520-dynamic-classes-cant-be-abc-subclasses.py
```diff
@@ -1,16 +1,15 @@
...
 import numpy as np
 import pytest
 import skhep_testdata
 
-import uproot
-
 
 @pytest.mark.skipif(
-    uproot._util.py36,
+    sys.version_info < (3, 7),
```
henryiii, line 12: I dropped this one helper; the sys.version_info is much easier to handle statically (pyupgrade can remove branches when it uses this syntax, for example), reads better (you don't have to look up the meaning of py36), and was the only place this helper was used.

src/uproot/extras.py
```diff
@@ -147,27 +146,9 @@ def lzma():
...
-    conda install backports.lzma
+    import lzma
 
-or use Python >= 3.3."""
-            )
-        else:
-            return lzma
-    else:
-        return lzma
+    return lzma
```
henryiii, line 151: This function is kind of pointless now, don't know if you want to deprecate it.

docs-sphinx/prepare_docstrings.py
jpivarski: I only use `+` for short things. There's a threshold for "short" that gets reevaluated now that the f-string syntax is available. `f"{modulename}.{name}"` > `modulename + "." + name` > `"{0}.{1}".format(modulename, name)`.

henryiii: Yes, not making statements on the old versions of things (I'd use + over format for something like this), but pointing out exactly this change (plus f-strings are wickedly fast for things like this, no extra function call(s)).

src/uproot/writing/_cascadetree.py
henryiii: Yes, again, all my statements are based on Python 3, not saying this should have been different in Python 2. Though this was always wrong in Python 3, you should be able to catch a ModuleNotFound, ImportError can also mean other import-related failures when a module is present but broken.

henryiii: But it would have been irritating to try to toggle based on Python version.

tests/test_0520-dynamic-classes-cant-be-abc-subclasses.py
jpivarski: There was a pack of `py26`, `py27`, `py35`, `py36` booleans, as well as an `iswin`. The `py36` was only needed in one case, which is why I still consider Python 3.6 acceptable—it needs very few work-arounds.

Meanwhile, there are a lot of different ways of saying `sys.version_info[0] == 3 and sys.version_info[1] < 7`, etc. (which mean subtly different things). This tuple-inequality idiom is succinct, and if it's recognized by static checkers, then it's the one to use. But I didn't know that there was a "one to use."

Similarly for the Windows-check: `iswin` has been redefined a few times as some ways of checking platform haven't been as robust as the one we have there now. By putting it in a boolean, we could fix it centrally.

Another one is `isstr`, which was a lot more important when we had to deal with different meanings of `str`, `bytes`, and `unicode`.

The `isint` is one that I think should stay. There isn't a good idiom for "integer but not boolean." There was some version of NumPy where NumPy scalars weren't being recognized as integers (can't remember which, exactly) and anyway Python and NumPy differ on whether booleans should be considered integers. Having a central facility for that makes it possible to control from one location.

src/uproot/extras.py
jpivarski: I had that in the back of my mind: we don't have to get `lzma` from extras.py anymore; it can be a top-of-the-module import. I think it only appears in compression.py.

henryiii:
Okay, I think I'm done with this. Some things (like + -> strings) could be worked on in the future, but don't need to be here.

jpivarski:
I scanned through all the changes and it looks good! With so many changes, the real proof that it's safe is in the tests. After it's merged, the Coffea tests should pick it up (is that right, @nsmith-?) and as a squashed commit, we have the option to revert it if there's a problem.

Thank you very much! (Also for that note about when `tobytes` was added to NumPy. We're still testing from 1.13.1 onward.)

nsmith-:
Yes coffea requires `awkward>=1.5.1,<2` and the daily tests will end up picking the latest version once its on pypi

PR #528, opened 2021-12-10
docs: correct version support badge

henryiii:
You could switch to the standard badge, it's shorter by a bit with 2.7 gone. ( will look more like ) when the next release happens. Then you don't have to maintain this.


jpivarski:
Sorry that I missed this earlier!

I noticed that the Awkward badge was out of date and wanted to make both of these be "3.6‒3.10" (3.6 is included and I prefer a range over enumeration, especially now that it's contiguous). That badge update is in #534, along with actually testing 3.10. Awkward Array with Python 3.10 might have to compile from source; that might fail, but Awkward 1.8.0 [is due](https://github.com/scikit-hep/awkward-1.0#roadmap), so I should be putting that in PyPI pretty soon. #534 can wait for that if necessary.

PR #534, opened 2022-01-04
Add test with Python 3.10.

jpivarski:


jpivarski:
@henryiii Everything went without troubles (though Awkward Array for Python 3.10 takes a little longer to compile, at least until Awkward 1.8.0 is out). I requested you as a reviewer just to make sure you're aware of this change and don't foresee problems downstream.

henryiii:
Okay, sounds good. Only thing missing for 3.10 in general is Numba, and that should be out any day. So far the addition of 3.10 has been very smooth - I do foresee a few more bumps for the 3.11 transition - it will be removing a lot of old Python 2.7 compatibility shims that packages may still be using. The performance changes are also occasionally affecting the C API (but pybind11 insulates downstream packages from that - but as a pybind11 developer I have to deal with some of them :) )

PR #538, opened 2022-01-17
Allow reading and writing from/to already opened files

jrueb:
`FileSink` seems to be designed such that it can also handle file objects. The only thing that is blocking this functionality is that `FileSink.from_object` does not return anything, which I assume is a mistake.
This change adds the missing return statement and a bit of documentation to open/recreate/update.

### DIFF
--- a/src/uproot/sink/file.py
+++ b/src/uproot/sink/file.py
@@ -57,6 +57,7 @@ class FileSink:
     * that has 'read', 'write', 'seek', and 'tell' methods
     * is 'readable() and writable() and seekable()'"""
             )
+        return self
 
     def __init__(self, file_path):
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
That's right: this was most likely a feature partially implemented and then forgotten about before it could be taken all the way. The fact that there's no test is why the mistake of not returning self got through. (It's a method that _looks like_ an `__init__`, but it's a non-default constructor as a `classmethod`, so `return self` has to be explicit.) It's not in the documentation because I forgot about it.

Did you test it? Do you want to include a test? I approve of this PR and would merge it, but if you want to include a test first, I'll wait.

Otherwise, just write a message or thumbs up and I'll merge it.

Issue #438, opened 2021-09-10
bugs with uproot and Delphes interactions


kratsg:
```python
>>> import uproot
>>> uproot.__version__
'4.1.1'
```

A simple script like the following will confirm bugs on both of these files:

```python
import uproot
with uproot.open('test.root:Delphes') as tree:
  tree.arrays(entry_stop=1)
```

which either results in one of two different errors (depends on Delphes versions... and what process is generated...)

<details>
  <summary>bug in <code>AsGroup</code> interpretation not handling <code>TCloneArrays</code></summary>
  <pre>
Traceback (most recent call last):
  File "inout.py", line 5, in <module>
    events = tree.arrays(entry_stop=1)
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/uproot/behaviors/TBranch.py", line 1130, in arrays
    False,
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/uproot/behaviors/TBranch.py", line 3494, in _ranges_or_baskets_to_arrays
    uproot.source.futures.delayed_raise(*obj)
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/uproot/source/futures.py", line 46, in delayed_raise
    raise exception_value.with_traceback(traceback)
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/uproot/behaviors/TBranch.py", line 3470, in basket_to_array
    branch,
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/uproot/interpretation/objects.py", line 228, in final_array
    output = library.finalize(output, branch, self, entry_start, entry_stop)
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/uproot/interpretation/library.py", line 570, in finalize
    (_object_to_awkward_json(form, x) for x in array), highlevel=False
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/awkward/operations/convert.py", line 885, in from_iter
    for x in iterable:
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/uproot/interpretation/library.py", line 570, in <genexpr>
    (_object_to_awkward_json(form, x) for x in array), highlevel=False
  File "/tank/scratch/kratsg/py3/lib64/python3.6/site-packages/uproot/interpretation/library.py", line 315, in _object_to_awkward_json
    return [_object_to_awkward_json(subform, x) for x in obj]
[...]

PR #570, opened 2022-03-04
fBits is 4 bytes

jpivarski:
This PR _reverses_ the second "fix" in #454, which was motivated by issue #438, raised by @kratsg.

Reading the text of PR #454, it's clear that I didn't believe `fBits` is actually 1 byte, I did it as an expedient because the file from Delphes had sometimes 4 bytes, sometimes 6 bytes, and at least assuming it's 1 byte would prevent it from raising error messages in `AsGrouped`. Reverting that may break @kratsg's workflow, but it didn't actually need the `fBits`—they were just coming along for the ride in `AsGrouped`. Would it be possible to instead avoid `AsGrouped`?

The file that this PR is based on, from @tfruth, demonstrates that the `fBits` really is 4 bytes. Maybe the Delphes ROOT writer is just wrong here. Anyway, the code can only be one way (PR #454 shows that I tried and failed to find any indicator for "4 bytes" vs "6 bytes") and this appears to be the correct one.

Both of you are affected by this. If it matters, please test against this PR!

### DIFF
--- a/src/uproot/interpretation/identify.py
+++ b/src/uproot/interpretation/identify.py
@@ -378,13 +378,7 @@ def interpretation_of(branch, context, simplify=True):
                 out = _float16_or_double32(branch, context, leaf, is_float16, dims)
 
             else:
-                if (
-                    branch.member("fClassName", none_if_missing=True) == "TObject"
-                    and branch.name.split(".")[-1] == "fBits"
-                ):
-                    from_dtype = numpy.dtype(">u1")
-                else:
-                    from_dtype = _leaf_to_dtype(leaf, getdims=False).newbyteorder(">")
+                from_dtype = _leaf_to_dtype(leaf, getdims=False).newbyteorder(">")
 
                 if context.get("swap_bytes", True):
                     to_dtype = from_dtype.newbyteorder("=")
--- a/tests/test_0569-fBits-is-4-bytes.py
+++ b/tests/test_0569-fBits-is-4-bytes.py
@@ -1,0 +1,14 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot4/blob/main/LICENSE
+
+import numpy as np
+import pytest
+import skhep_testdata
+
+import uproot
+
+
+def test():
+    with uproot.open(skhep_testdata.data_path("uproot-issue-569.root")) as f:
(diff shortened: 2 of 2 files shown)


kratsg:
The Delphes file i uploaded in the linked issue doesn't seem to work on this PR so not sure what the plan is here. Certainly Delphes is a common enough output format that uproot should consider supporting on some level, but not really sure.

jpivarski:
I don't see how we can make both work. The Delphes file has some 4-byte `fBits` and some 6-byte `fBits`, but `fBits` is supposed to be 4 bytes. You can read it without `AsGrouped`, right? If you avoid reading the `fBits`, that works, doesn't it?

kratsg:
Yeah but I don't know how Delphes is doing it. It doesn't touch fBits directly.. that's all on ROOT.

jpivarski:
I thought that Delphes had an independent ROOT-writer—writing ROOT files without using ROOT, the way that Uproot does. (Or maybe I'm thinking of GEANT?)

kratsg:
> I thought that Delphes had an independent ROOT-writer—writing ROOT files without using ROOT, the way that Uproot does. (Or maybe I'm thinking of GEANT?)

I think you're thinking of GEANT. Delphes requires ROOT.

tfruth:
Thank you @jpivarski. This would solve our issue. (We're also just using ROOT to write out the file.)

kratsg:
Yeah, I'll drop `fBits` -- but we need to probably loop up with coffea to update the code there to make sure coffea doesn't break.. @nsmith-

tfruth:
Thanks @jpivarski - will there be a new tag with this fix?

jpivarski:
It's [4.2.1](https://github.com/scikit-hep/uproot4/releases/tag/4.2.1).

nsmith-:
> Yeah, I'll drop `fBits` -- but we need to probably loop up with coffea to update the code there to make sure coffea doesn't break.. @nsmith-

I'll need some help here. I search for `fBits` in coffea and get nothing https://github.com/CoffeaTeam/coffea/search?q=fBits

kratsg:
Oh, it looks like it was already removed back then when I was sturggling with this.. and I'm on an old version of my old branch 😅 my bad....

PR #573, opened 2022-03-14
Restore performance hack for AsArray(True, False, AsVector(False, dtype)).

jpivarski:
Please test, @tamasgal.

### DIFF
--- a/src/uproot/containers.py
+++ b/src/uproot/containers.py
@@ -652,7 +652,6 @@ class AsArray(AsContainer):
                     "as": "array",
                     "header": self._header,
                     "speedbump": self._speedbump,
-                    "inner_shape": list(self._inner_shape),
                 }
             },
         )


jpivarski:
Oh, and also @nsmith-: you didn't actually need the `inner_shape` to be in the parameters, did you? (That information is in the NumpyForm itself. Adding it to the parameters made the back-door for this data type in Awkward to not recognize it.)

The AsArray has `inner_shape`, too. The information is not lost—I just don't know if you're relying on it in Coffea. (If so, I'll tell you where to get the information from instead of the parameter.)

nsmith-:
It looks like in coffea we don't care about any of the uproot fields in the parameters https://github.com/CoffeaTeam/coffea/blob/ff7878ce20db5d425134808ecc0c1a55fddaa5fa/coffea/nanoevents/mapping/uproot.py#L59 so I think its probably safe to delete this parameter. I cannot be sure why I added it there.

tamasgal:
Thanks @jpivarski I confirm that it works now just like before 4.1.3! `:)`

jpivarski:
> I cannot be sure why I added it there.

Probably because you were adding it to the constructor, and there are other constructor arguments in the parameters, so I think you were going for symmetry.

Okay, I'll merge this and make a release!

PR #598, opened 2022-05-17
Add `to_dataframe` to `TTree`

tzuhanchang:
Add ```to_dataframe``` to ```TTree``` for converting TTree to a ```pandas.DataFrame```.

### DIFF
--- a/src/uproot/behaviors/TTree.py
+++ b/src/uproot/behaviors/TTree.py
@@ -163,7 +163,7 @@ class TTree(uproot.behaviors.TBranch.HasBranches):
     def to_dataframe(self):
         pandas = uproot.extras.pandas()
         dict = {}
-        for branch in (self.keys()):
+        for branch in self.keys():
             dict[branch] = self[branch].array(library="np")
         dataFrame = pandas.DataFrame(dict)
         return dataFrame
--- a/src/uproot/behaviors/TTree.py
+++ b/src/uproot/behaviors/TTree.py
@@ -159,3 +159,11 @@ class TTree(uproot.behaviors.TBranch.HasBranches):
         self._chunk = chunk
         self._lookup = {}
         return self
+
+    def to_dataframe(self):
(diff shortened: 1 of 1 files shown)


jpivarski:
Hi! Thanks for your first pull request, but did you know that `library="pd"` can be passed to any array-fetching function (e.g. [uproot.TTree.arrays](https://uproot.readthedocs.io/en/latest/uproot.behaviors.TTree.TTree.html#arrays))? That does what this `to_dataframe` would do. (It's [here in the documentation](https://uproot.readthedocs.io/en/latest/basic.html#reading-a-tbranch-as-an-array).)

tzuhanchang:
> Hi! Thanks for your first pull request, but did you know that `library="pd"` can be passed to any array-fetching function (e.g. [uproot.TTree.arrays](https://uproot.readthedocs.io/en/latest/uproot.behaviors.TTree.TTree.html#arrays))? That does what this `to_dataframe` would do. (It's [here in the documentation](https://uproot.readthedocs.io/en/latest/basic.html#reading-a-tbranch-as-an-array).)

Thank you very much. There are few challenges for using `library="pd"`. It does not play nice with "vector" leaf type, it creates subentry for an events, instead we want 1 event per row in the df. Also `tree.arrays(library="pd")` takes way too long to run when "vector" leaf type variables are present in the tree.

jpivarski:
By "vector" leaf type, do you mean jagged arrays? A function that puts a list of numbers into a Pandas cell (as opposed to expanding that list into one number per cell, tracked by a MultiIndex) would have to have a name that clearly specifies that that's what it does.

I don't know of any Pandas functions that can deal adequately with non-numeric values in cells. It represents them as Python lists or NumPy arrays, and then you're forced to manipulate them with Python iteration, using one of the various `map` functions in Pandas. That doesn't scale to large datasets.

tzuhanchang:
Thank you so much for your help!

PR #609, opened 2022-06-14
Add the uproot4.num_entries function

kkothari2001:
Usage commands:

```python
import uproot
import skhep_testdata
from pprint import pprint as pp

file1 = skhep_testdata.data_path("uproot-Zmumu.root") + ":events"
file2 = skhep_testdata.data_path("uproot-HZZ.root") + ":events"
```

Outputs:
```python
>>> pp(uproot.num_entries(file1))
[(2304,)]
>>> pp(uproot.num_entries([file1,file2]))
[(2304,), (2421,)]
```

Tests, docstring and cleanup: WIP

### DIFF
--- a/src/uproot/models/TTree.py
+++ b/src/uproot/models/TTree.py
@@ -964,10 +964,17 @@ def num_entries(paths):
     Returns an iterator over the number of entries over each TTree in the input.
     This is a shortcut method and reads lesser data than normal file opening.
     """
-    paths = _regularize_files(paths)
+    paths2 = _regularize_files(paths)
 
-    for file_path, object_path in paths.items():
-        yield uproot.open(
+    if isinstance(paths, dict):
+        paths = list(paths.items())
+    elif not uproot._util.isstr(paths):
+        paths = [(uproot._util.file_object_path_split(path)) for path in paths]
+    else:
+        paths = [uproot._util.file_object_path_split(paths)]
+
+    for i, (file_path, object_path) in enumerate(paths2.items()):
+        yield paths[i][0], paths[i][1], uproot.open(
             {file_path: object_path}, custom_classes={"TTree": Model_TTree_NumEntries}
         ).all_members["fEntries"][0]
 
--- a/src/uproot/models/TTree.py
+++ b/src/uproot/models/TTree.py
@@ -8,13 +8,7 @@ functions.
 """
 
 
-import glob
-import itertools
-import os
-import re
 import struct
-from collections.abc import Iterable
-from urllib.parse import urlparse
 
 import numpy
 
@@ -964,7 +958,7 @@ def num_entries(paths):
     Returns an iterator over the number of entries over each TTree in the input.
     This is a shortcut method and reads lesser data than normal file opening.
     """
-    paths2 = _regularize_files(paths)
+    paths2 = uproot._util.regularize_files(paths)
 
     if isinstance(paths, dict):
         paths = list(paths.items())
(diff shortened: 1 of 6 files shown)


kkothari2001:
Final usage:
```python
>>> from pprint import pprint as pp
>>> file1 = skhep_testdata.data_path("uproot-Zmumu.root") + ":events"
>>> file2 = skhep_testdata.data_path("uproot-HZZ.root") + ":events"
>>>
>>> for val in uproot.num_entries([file1,file2]):
...     print(val)
...
2304
2421
```
This PR is now ready for review.

jpivarski:
From your example, it looks like the output is an iterable of numbers. The input paths can be changed by the `_regularize_files` function—it might be expanding wildcards on local file systems, maybe also expanding home directory tildes and turning relative paths into absolute paths. It would be much safer to use this function if the output had the actual paths used associated with the numbers of entries. I think we discussed `yield file_path, object_path, num_entries`. That would provide the most information (and there are idioms for wrapping that up as a dict, etc.).

kkothari2001:
Thanks for the review! Sorry I forgot we discussed about yielding all 3 vars.

New usage:
```python
>>> file0 = '~/.local/skhepdata/uproot-Zmumu.root:events' # including the tilde
>>> file2 = skhep_testdata.data_path("uproot-HZZ.root") + ":events"
>>>
>>> for val in uproot.num_entries([file0,file2]):
...     print(val)
...
('~/.local/skhepdata/uproot-Zmumu.root', 'events', 2304)
('/home/kmk/.local/skhepdata/uproot-HZZ.root', 'events', 2421)
>>> # Internally, the tilde is expanded for our use, but then only the representation provided by the user is yielded
```
I just realised the problem with the tests having specific to my computer. Do I change them to `/home/runner` or just test the `num_entries` obtained?

kkothari2001:
I dont think this is failing due to something I changed.

jpivarski:
> I dont think this is failing due to something I changed. Because other builds seem to have succeded.

It looked like ROOT's behavior changed (#612).

src/uproot/models/TTree.py
```diff
@@ -905,3 +911,161 @@ def read_members(self, chunk, cursor, context, file):
...
+            if key not in seen:
+                out[file_path] = object_path
+                seen.add(key)
+        else:
+            out[file_path] = object_path
+
+    if len(out) == 0:
+        raise uproot._util._file_not_found(files)
+
+    return out
```
jpivarski, line 1071: It looks like `_regularize_files` and its helper function are (still) [also defined in TBranch.py](https://github.com/scikit-hep/uproot4/blob/7fc2b39bc208fed8fda12233279f2b166e9b9977/src/uproot/behaviors/TBranch.py#L2987-L3119).

There should only be one definition of this function, and since both behaviors/TBranch.py and models/TTree.py want access to it, it should be in a neutral location: _util.py. I also selected `_regularize_object_path` because that should go along wherever `_regularize_files` goes.

Actually, a lot of the bulk of TBranch.py is helper functions, and many of them ought to be in _util.py. But let's just focus on these three (`_regularize_files`, its helper function, and `_regularize_object_path`).

jpivarski (changes requested):
Great! It looks like this does it. The Uproot 3 version also had to prevent the ReadOnlyFile from reading TStreamerInfo in this `numentries` fast-path, but Uproot 4 avoids reading TStreamerInfo by default, so there's nothing extra to do there.

I just have one request about the location (and duplication) of `_regularize_files`.

jpivarski:
Oh, and if you merge with `main`, the failing test should be fixed. (I meant to do that before re-running those tests...)

kkothari2001:
I have moved the functions to `_util.py` but this was causing a circular import error as `HasBranches` and `_NoClose` classes could not be moved from `behaviors/TBranch.py`. I have kind of avoided that by:
- Importing the required classes only inside the `regularize_files` and `regularize_object_path` functions.
- Passing the `HasBranches` class into `_regularize_files_inner` as a new argument, as it didn't make sense to import it at every call of `_regularize_files_inner`

I could possibly define `_regularize_files_inner` inside `regularize_files` in order to avoid the unnecessary argument (already tested this and it works). Would you suggest that? Please have a look and suggest further changes if necessary.

jpivarski:
Importing `_util` late is one solution, but Python also makes a distinction between

```python
import uproot._util
```

and

```python
from uproot._util import _regularize_files
```

If it knows that it only has to import one symbol, it somehow manages to avoid the circularity. We use that in a few places.

In general, I don't like to make functions nested unless the inner one is enclosing some variables of the outer one's scope, which can simplify it by reducing the number of arguments. It's like avoiding too-deep class hierarchies in object-oriented programming: we _can_ nest then, but we shouldn't do so unless there's a reason. Another potential reason is that it's the only way to make the helper function absolutely private in Python. Nested functions, unlike underscored functions, absolutely can't be accessed. That could be a good reason for this pair, because there's no reason the helper function could ever be used by anything else. However, that logic wasn't applied to the rest of the Uproot codebase.

I should look at what you've done, because late import is also a fine solution to the problem. (I can't check today, though.) Just as long as it minimizes complexity, I'm happy.

src/uproot/models/TTree.py
```diff
@@ -905,3 +905,69 @@ def read_members(self, chunk, cursor, context, file):
...
+            raise NotImplementedError(
+                """memberwise serialization of {}
+in file {}""".format(
+                    type(self).__name__, self.file.file_path
+                )
+            )
+        cursor.skip_over(chunk, context)
+        cursor.skip_over(chunk, context)
+        cursor.skip_over(chunk, context)
+        cursor.skip_over(chunk, context)
```
jpivarski, line 929: I'm just giving this some documentation, to explain why we skip exactly 4 times.

```suggestion
        cursor.skip_over(chunk, context)  # TNamed
        cursor.skip_over(chunk, context)  # TAttLine
        cursor.skip_over(chunk, context)  # TAttFill
        cursor.skip_over(chunk, context)  # TAttMarker
```

jpivarski (approved):
This looks perfect!

I don't know how you got around the circular import issue because this is doing a full import of `uproot._util` and yet... oh, I get it: there's no explicit

```python
import uproot._util
```

at the top of the file, but inside of functions, you freely `uproot._util._regularize_object_path` because by that point, it's imported. I think that's fine. It loses the self-documentation of having the import at the top of the file declare that the module is being used, but you can also determine that the module is being used by a pattern-match of all "`uproot._util.`" (the value of fully qualified function names).

So yes, I'll merge this once the tests go through again: I'm just adding comments to the four `skip_over` lines, a note to future maintainers what those four lines are for.

kkothari2001:
Thanks!

PR #610, opened 2022-06-14
Adding a AwkwardForth based ROOT reader to uproot4

aryan26roy:


src/uproot/_awkward_forth.py
```diff
@@ -0,0 +1,80 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot4/blob/main/LICENSE
+
+"""
+This module defines utilities for adding components to the forth reader.
+"""
+
+
+class forth_obj:
```
jpivarski, line 8: PEP8 class names: `forth_obj` → `ForthObj`.

src/uproot/_awkward_forth.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    Util class for holding information related to Forth code generation.
```
jpivarski, line 10:
```suggestion
    This class is passed through the Forth code generation, collecting Forth snippets and concatenating them at the end.
```

src/uproot/containers.py
```diff
@@ -11,10 +11,12 @@
 import struct
 import types
 from collections.abc import KeysView, Mapping, Sequence, Set, ValuesView
+from re import L  # noqa : F401
```
jpivarski, line 14:
```suggestion
```

Not used.

src/uproot/containers.py
```diff
@@ -11,10 +11,12 @@
...
 
 import numpy
 
 import uproot
+from uproot.deserialization import DeserializationError  # noqa : F401
```
jpivarski, line 19:
```suggestion
```

src/uproot/containers.py
```diff
@@ -406,35 +408,106 @@ def typename(self):
         else:
             return self._typename
 
+    # def descent(self, context):
+
```
jpivarski, line 412:
```suggestion
```

Don't leave in commented-out code.

src/uproot/containers.py
```diff
@@ -406,35 +408,106 @@ def typename(self):
...
+        fheader = []
+        finit = []
+        if "forth" in context.keys():
+            awkward = uproot.extras.awkward()
+            forth = True
+            forth_obj = context["forth"]
+            isinstance(
+                forth_obj.aform,
+                (awkward.layout.NumpyArray,),
+            )
```
jpivarski, line 470: This code doesn't do anything.

Also, if `aform` is really a Form, it can never have type `NumpyArray`.

jpivarski (commented):
I'm not going to merge this now; it involves a lot of changes to the `read` methods. There are a lot of `read` methods in the Uproot codebase—adding all of this to each one of them would add a lot of complexity.

There should be a way to generalize what you've done with these two `read` methods (on `AsVector` and `AsString`) so that it's less invasive. Somehow, the code that performs this logic should be mostly outside of `read`, maybe something that's just called.

jpivarski:
@all-contributors please add @aryan26roy for code

jpivarski:
Is this superseded by #644? If so, let's close the old ones so that I don't get confused about which one to look at.

PR #611, opened 2022-06-17
awkward_form arguments are now (self, file, context).

jpivarski:
Instead of `index_format, header, tobject_header, breadcrumbs`.

This change makes it possible to add as many arguments to `awkward_form` as we need.

### DIFF
--- a/src/uproot/interpretation/objects.py
+++ b/src/uproot/interpretation/objects.py
@@ -104,22 +104,11 @@ class AsObjects(uproot.interpretation.Interpretation):
         else:
             return uproot.model.classname_decode(self._model.__name__)[0]
 
-    def awkward_form(
-        self,
-        file,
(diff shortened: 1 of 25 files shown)


src/uproot/interpretation/objects.py
```diff
@@ -137,7 +126,15 @@ def basket_array(
...
-            form = self.awkward_form(branch.file, index_format="i64")
+            form = self.awkward_form(
+                branch.file,
+                {
+                    "index_format": "i64",
+                    "header": False,
+                    "tobject_header": True,
+                    "breadcrumbs": (),
+                },
+            )
```
jpivarski, line 137: This is the one and only place where the `context` for `awkward_form` gets set (that won't be going away, as `uproot.lazy` will).

PR #615, opened 2022-06-21
Updating docs and test in response to the removal of uproot.lazy in Uproot5

kkothari2001:


src/uproot/behaviors/TBranch.py
```diff
@@ -395,6 +395,8 @@ def lazy(
     **options,  # NOTE: a comma after **options breaks Python 2
 ):
     """
+    .. warning::
+        uproot.lazy has been was removed from Uproot5. Users of Uproot4 can continue using it as is until EOL.
```
jpivarski, line 399:
```suggestion
        Deprecated. This function will be removed in Uproot 5 (target date: Dec 2022). Please use :doc:`uproot.behaviors.TBranch.dask` instead.
```

src/uproot/behaviors/TBranch.py
```diff
@@ -501,196 +503,9 @@ def lazy(
     * :doc:`uproot.behaviors.TBranch.lazy` (this function): returns a lazily
       read array from ``TTrees``.
     """
-    awkward = uproot.extras.awkward()
-    files = uproot._util.regularize_files(files)
-    decompression_executor, interpretation_executor = _regularize_executors(
-        decompression_executor, interpretation_executor, None
+    raise NotImplementedError(
+        "uproot.lazy was deprecated in Uproot5 in favour of uproot.dask. If you still wish to use uproot.lazy please install Uproot4"
```
jpivarski, line 507:
```suggestion
        "uproot.lazy is removed in Uproot5 (only here for the deprecation message in the online docs)"
```

jpivarski (approved):
Just changed the messages based on what we talked about in Slack.

src/uproot/behaviors/TBranch.py
jpivarski: I suppose this message assumes that the reader (the user) knows what version of Uproot they're using; that's not always the case.

PR #616, opened 2022-06-21
Completed the Forth based AsStrings reader.

aryan26roy:


src/uproot/interpretation/strings.py
```diff
@@ -152,74 +157,150 @@ def basket_array(
             cursor_offset=cursor_offset,
             library=library,
         )
+        if (
+            isinstance(library, uproot.interpretation.library.Awkward)
+            and byte_offsets is None
+        ):
+            import awkward.forth
```
jpivarski, line 164: Before this line, call `uproot._extras.import_awkward()` to make sure users without Awkward get a good error message.

src/uproot/interpretation/strings.py
```diff
@@ -152,74 +157,150 @@ def basket_array(
...
+                        output out-main uint8
+                        output out-offsets int64
+
+                        0 out-offsets <- stack
+
+                        begin
+                            stream !B-> stack dup 255 = if drop stream !I-> stack then dup out-offsets +<- stack stream #!B-> out-main
+                        again"""
+
+                    self._code_complete = True
```
jpivarski, line 181: You don't have to be careful about not regenerating the "1-5" Forth code string and the "4" Forth code string, since they're constants (`.pyc`-compile-time constants for Python: they're not made, they're just looked up).

src/uproot/interpretation/strings.py
```diff
@@ -152,74 +157,150 @@ def basket_array(
...
+                if not self._vm_made and self._code_complete:
+                    self._forth_vm.vm = awkward.forth.ForthMachine64(self._forth_code)
+                    self._vm_made = True
```
jpivarski, line 184: But this `self._vm_made` guard _is_ a good idea (along with making it thread-local).

src/uproot/interpretation/strings.py
```diff
@@ -246,6 +327,32 @@ def final_array(
             library=library,
             branch=branch,
         )
+        if (
+            isinstance(library, uproot.interpretation.library.Awkward)
+            and self._special_case
+        ):
```
jpivarski, line 333:
```python
if any(not isinstance(x, StringArray) for x in basket_arrays.values()):
```

or the opposite of that for the StringArray case. Avoid extra attributes like `self._special_case`, especially mutable ones, especially if they don't have descriptive names. `:)`

src/uproot/interpretation/strings.py
```diff
@@ -246,6 +327,32 @@ def final_array(
...
+                entry_start=entry_start,
+                entry_stop=entry_stop,
+                entry_offsets=entry_offsets,
+                library=library,
+                branch=branch,
+                output=output,
+            )
+            output = library.finalize(output, branch, self, entry_start, entry_stop)
+
+            return output
```
jpivarski, line 355: A short-circuit `return` at the beginning of a function can be good if the if-block it's in is short and represents an exceptional case. It lets readers ignore that exception and focus on the "normal" cases below.

Here, the if-statement is a decision between two normal cases; bracketing them with `if`-`else` (and maybe `elif`) clarifies this. This one needs an `else`.

jpivarski (changes requested):
Just a few things.

jpivarski (approved):

PR #617, opened 2022-06-21
changed arguments for awkward_form

aryan26roy:


src/uproot/interpretation/grouped.py
```diff
@@ -84,7 +84,18 @@ def typename(self):
...
+    def awkward_form(
+        self,
+        file,
+        context=None,
+        index_format="i64",
+        header=False,
+        tobject_header=True,
+        breadcrumbs=(),
+    ):
+        context = super()._make_context(
```
jpivarski, line 96: No `super()`; just `self`.

src/uproot/interpretation/__init__.py
```diff
@@ -56,6 +56,17 @@ def numpy_dtype(self):
         """
         raise AssertionError
 
+    @classmethod
+    def _make_context(self, context, index_format, header, tobject_header, breadcrumbs):
```
jpivarski, line 60: First argument of a `@classmethod` must be `cls`, not `self`.

But it also doesn't need to be a `@classmethod`; this could be a `@staticmethod` with no special first argument.

jpivarski (changes requested):
Just the two things below.

jpivarski (approved):
This is done. I'll enable auto-(squash-and)-merge!

PR #618, opened 2022-06-21
changed arguments for awkward_form (→ main-v4)

jpivarski:
#617 went into `main`; this PR goes into `main-v4`.

### DIFF
--- a/src/uproot/interpretation/__init__.py
+++ b/src/uproot/interpretation/__init__.py
@@ -56,6 +56,17 @@ class Interpretation:
         """
         raise AssertionError
 
+    @classmethod
+    def _make_context(self, context, index_format, header, tobject_header, breadcrumbs):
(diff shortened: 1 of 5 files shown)


src/uproot/interpretation/__init__.py
```diff
@@ -56,6 +56,17 @@ def numpy_dtype(self):
         """
         raise AssertionError
 
+    @staticmethod
+    def _make_context(context, index_format, header, tobject_header, breadcrumbs):
+        if context is None:
+            context = {}
+        context["index_format"] = "i64"
```
btovar, line 63: I wonder if here you meant:

```
         context["index_format"] = index_format
etc...
```

jpivarski: Yeah, you're right. One moment...

PR #625, opened 2022-06-22
Fix annoying gaps in test files.

jpivarski:
This went into `main-v4`; it should also go into `main`.


jpivarski:
This is also a test of the CI machinery. The commits to `main-v4` have to run tests that trigger on a merge to `main-v4` and the commits to `main` have to run tests that trigger on a merge to `main`.

That seems to be working.

PR #629, opened 2022-06-23
Cleaning up string generation in streamers.py

aryan26roy:


src/uproot/streamers.py
```diff
@@ -815,42 +777,34 @@ def class_code(
 
         if streamerinfo.name == "TBranch" and self.name == "fBasketSeek":
             read_members.append(
-                "        if context.get('speedbump', True):\n"
-                "            if cursor.bytes(chunk, 1, context)[0] == 2:\n"
-                "                tmp = numpy.dtype('>i8')"
+                """        if context.get('speedbump', True):
+                            if cursor.bytes(chunk, 1, context)[0] == 2:
+                                tmp = numpy.dtype('>i8')"""
```
jpivarski, line 782: Be careful with this and Python's indentation rules. The string that this generates is actually

```
        if context.get('speedbump', True):
                            if cursor.bytes(chunk, 1, context)[0] == 2:
                                tmp = numpy.dtype('>i8')
```

which is technically okay for indented Python but the `if cursor.bytes...` is indented 20 spaces more than the line above it, rather than 4 spaces more. For this code, that works because the indentation rule is a stack: a nested block only needs to be indented _more_ than its surroundings; it doesn't need to be only one level more.

However, if this is to be concatenated with some more code that assumes the indented block is only 4 spaces more, or if the second line wasn't an indentation of the first, then there would be Python syntax errors or wrong behavior. (This is a reason why Python isn't a good language for metaprogramming! Forth's minimal syntax allows the code that generates it to take priority in aesthetics over the code that is generated.)

How about this? It adds a carriage return at the beginning and the end so that we can write it inline, like ordinary Python with the right indentation, and the `.strip("\n")` removes the outermost `"\n"` (but none of the internal ones, and not other whitespace like `" "`).

```suggestion
                """
        if context.get('speedbump', True):
            if cursor.bytes(chunk, 1, context)[0] == 2:
                tmp = numpy.dtype('>i8')
""".strip("\n")
```

src/uproot/streamers.py
```diff
@@ -815,42 +777,34 @@ def class_code(
...
             )
             read_member_n.append("    " + read_members[-1].replace("\n", "\n    "))
 
         else:
             read_members.append(
-                "        if context.get('speedbump', True):\n"
-                "            cursor.skip(1)"
+                """        if context.get('speedbump', True):
+                               cursor.skip(1)"""
```
jpivarski, line 789: Same thing:

```suggestion
                """
        if context.get('speedbump', True):
            cursor.skip(1)
""".split("\n")
```

src/uproot/streamers.py
```diff
@@ -815,42 +777,34 @@ def class_code(
...
-                ),
-                "            parameters={'uproot': {'as': 'TStreamerBasicPointer', "
-                "'count_name': " + repr(self.count_name) + "}}",
-                "        )",
+                f"        contents[{repr(self.name)}] = ListOffsetForm(context['index_format'], uproot._util.awkward_form(cls._dtype{len(dtypes)}, file, context),",
+                "            parameters={'uproot': {'as': 'TStreamerBasicPointer', 'count_name': "
+                + repr(self.count_name)
+                + "}},)",
             ]
         )
```
jpivarski, line 809: This one is a bit convoluted: the `repr(self.count_name)` is still concatenated with `+`, though it's an f-string. The number of curly brackets gets exciting because you need to fill some code inside JSON. And then the whole things is an `awkward_form.extend` with a list of two strings: the `extend` + comma is doing the same work that concatenation would do. Maybe it was so that the JSON wouldn't be part of the f-string, so that the curly brackets wouldn't have to be doubled?

src/uproot/streamers.py
```diff
@@ -1155,21 +1090,17 @@ def class_code(
...
+                    repr(self.typename.rstrip("*"))
+                ),
                 "        contents["
                 + repr(self.name)
-                + "] = ListOffsetForm(context['index_format'], "
-                "tmp, parameters={'uproot': {'as': TStreamerLoop, 'count_name': "
+                + "] = ListOffsetForm(context['index_format'], tmp, parameters={'uproot': {'as': TStreamerLoop, 'count_name': "
                 + repr(self.count_name)
                 + "}})",
             ]
```
jpivarski, line 1091: Same comment as above.

jpivarski (approved):
This is looking much better! I have a few suggestions, and then another part that I just commented on because I don't know what is the best course of action. (Maybe two `append`s instead of that big `extend`? That could help.)

Any time you're happy with it and all the tests pass, you can squash-and-merge.

Actually, one more thing: some of these might not be active in any tests. You can find this out by putting "`raise Exception`" before every string, run the tests, see the failures, and remove each "`raise Exception`" that you see in the test failures until all the tests run fine but there are still some "`raise Exception`" lines left. Be extra-careful about those lines: those are the ones that could introduce an error that we don't see.

(This technique is a poor-man's coverage tool, but it's also a directed coverage tool: at the moment, you're only interested in the coverage of these string-generations.)

PR #631, opened 2022-06-29
chore: add dependabot for actions

henryiii:
Will help keep actions up to date.


jpivarski:
Sounds good to me. I'll be merging this into `main` (Uproot v5 development). Should it also be in `main-v4` (Uproot v4 maintenance)? We're maintaining two central branches until Awkward v1 is dropped in December. If this is important for any Uproot v4 updates, I'll cherry-pick this PR commit into the `main-v4` branch.

PR #636, opened 2022-06-30
Forth based ROOT reader (revised)

aryan26roy:


src/uproot/_awkward_forth.py
```diff
@@ -0,0 +1,159 @@
...
+                return temp_node
+            else:
+                temp_node = self.awkward_model
+                self.awkward_model = self.awkward_model["content"]
+                return temp_node
+        if isinstance(self.awkward_model, list):
+            for elem in self.awkward_model:
+                if name in elem.values():
+                    return elem
+            self.awkward_model.append(
```
jpivarski, line 87: Here, too.

```suggestion

            self.awkward_model.append(
```

src/uproot/containers.py
```diff
@@ -876,20 +964,34 @@ def awkward_form(self, file, context):
...
+                    fcode_pre.append(f"{temp_jump} stream skip\n")
         else:
             is_memberwise = False
 
         # Note: self._values can also be a NumPy dtype, and not necessarily a class
         # (e.g. type(self._values) == type)
         _value_typename = _content_typename(self._values)
         if is_memberwise:
+            if forth:
+                raise NotImplementedError
```
jpivarski, line 994: These should eventually give a reason: `NotImplementedError` because of memberwise serialization.

src/uproot/interpretation/objects.py
```diff
@@ -135,11 +142,82 @@ def basket_array(
...
+
         output = None
-        if isinstance(library, uproot.interpretation.library.Awkward):
-            form = self.awkward_form(branch.file)
+        if "forth" not in context.keys():
+            forth_obj = uproot._awkward_forth.ForthObj()
 
-            if awkward_can_optimize(self, form):
+            context["forth"] = forth_obj
+            # self._model.descent(context)
```
jpivarski, line 209: Commented-out: is it going away?

src/uproot/models/TRef.py
```diff
@@ -115,13 +116,56 @@ def name(self):
...
                     type(self).__name__, self.file.file_path
                 )
             )
+        if forth:
+            key1 = forth_obj.get_key()
+            key2 = forth_obj.get_key()
+            key3 = forth_obj.get_key()
+            key4 = forth_obj.get_key()
+            key5 = forth_obj.get_key()
+            key6 = forth_obj.get_key()
```
jpivarski, line 139: Maybe you want a `forth_obj` member function that can be used as

```python
keys = forth_obj.get_keys(6)
```

src/uproot/streamers.py
```diff
@@ -672,13 +672,24 @@ def class_code(
...
+        forth = False
+        forth_obj = None
+        fcode_pre = []
+        fcode_post = []
+        temp = None
+        if "forth" in context.keys():
+            forth = True
+            forth_obj = context["forth"]
+            key = forth_obj.get_key()\n"""
+        )
```
jpivarski, line 685: I think you don't want to put code into the `read_member_n`, which is for the memberwise format. (Unless a particular type has memberwise implemented in Python as well.)

tests/test_610-awkward-form.py
```diff
@@ -0,0 +1,590 @@
+import json
```
jpivarski, line 1: First line should be point to the LICENSE.

These tests are probably the same as the last 10 or so in the new test file I made, so those will probably automatically work. (No reason to avoid the duplication, either. You can test the same thing here and there.)

tests/test_610-awkward-form.py
```diff
@@ -0,0 +1,590 @@
...
+
+import numpy
+import pytest
+import skhep_testdata
+
+import uproot
+
+
+def test_awkward_strings():
+    awkward = pytest.importorskip("awkward")
```
jpivarski, line 12: These can go at the top of the module file; then you'd only have to write it once.

src/uproot/_awkward_forth.py
```diff
@@ -0,0 +1,159 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot4/blob/main/LICENSE
+
+"""
+This module defines utilities for adding components to the forth reader.
+"""
+
+
+class ForthObj:
```
jpivarski, line 8: A more descriptive name may be helpful: `ForthGenerator`?

jpivarski (approved):
I'm in favor of merging this! I made a lot of comments above, but they can be dealt with in subsequent PRs.

jpivarski (approved):
Approved! (In person.)

PR #637, opened 2022-07-01
Set up tests for AsObjects, for the AwkwardForth reader

jpivarski:


src/uproot/interpretation/objects.py
```diff
@@ -205,7 +205,11 @@ def final_array(
 
             start = stop
 
-        if all(uproot._util.from_module(x, "awkward") for x in basket_arrays.values()):
+        if len(basket_arrays) == 0:
+            output = numpy.array([], dtype=self.numpy_dtype)
+        elif all(
+            uproot._util.from_module(x, "awkward") for x in basket_arrays.values()
+        ):
```
jpivarski, line 212: This was an actual bug-fix, discovered while arranging these tests but otherwise unrelated to them.

jpivarski:
@aryan26roy, this is a suite of test-stubs that you can use to set up test-driven development for adding Forth code. Only that which is tested by these tests needs to be developed as part of your project. The comments in the code with your name are cross-listed with the test numbers, so that you can plan which ones to solve when. By the end of your project, those comment lines should be removed (even the ones you can't do anything with, the "untested" ones).

Let me know if it would be easier for you to merge your branch first or for me to merge this one first. The target is `main`, not `main-v4`.

tests/test_0031-test-stl-containers.py
```diff
@@ -14,6 +14,31 @@
 from uproot.interpretation.objects import AsObjects
 
 
+@pytest.mark.skip(
+    reason="Implement non-memberwise std::map; we have a sample (map<string,double>)"
+)
+def test_nonmemberwise_asmap():
```
jpivarski, line 20: This is for me to implement non-memberwise `std::map` in Python, so that you can develop Forth code for both the memberwise and non-memberwise cases. (Just for `std::map`.) These four tests will be added to test_0637.

PR #652, opened 2022-07-18
Dask awkward support for uproot.dask

kkothari2001:
@jpivarski @douglasdavis  please have a look

Usage:
```python
>>> 
>>> filename1 = skhep_testdata.data_path("uproot-Zmumu.root") + ":events"
>>> filename2 = skhep_testdata.data_path("uproot-Zmumu-uncompressed.root") + ":events"
>>> 
>>> uproot.dask(filename1, library='ak')
dask.awkward<from-delayed, npartitions=1>
>>> 
>>> uproot.dask(filename1, library='ak').compute()
<Array [{Type: 'GT', Run: 148031, ...}, ...] type='2304 * {Type: string, Ru...'>
>>> 
>>> uproot.dask(filename1, step_size='100 B',library='ak')
dask.awkward<from-delayed, npartitions=2304>
>>> uproot.dask(filename1, step_size='100 B',library='ak').compute()
<Array [{Type: 'GT', Run: 148031, ...}, ...] type='2304 * {Type: string, Ru...'>
>>> 
>>> uproot.dask(filename1, step_size=76,library='ak')
dask.awkward<from-delayed, npartitions=31>
>>> uproot.dask(filename1, step_size=76,library='ak').compute()
<Array [{Type: 'GT', Run: 148031, ...}, ...] type='2304 * {Type: string, Ru...'>
>>> 
>>> 
>>> uproot.dask([filename1,filename2], step_size=76,library='ak')
dask.awkward<from-delayed, npartitions=62>
>>> uproot.dask([filename1,filename2], step_size=76,library='ak').compute()
<Array [{Type: 'GT', Run: 148031, ...}, ...] type='4608 * {Type: string, Ru...'>
>>> 
>>> 
>>> uproot.dask([filename1,filename2],library='ak',open_files=True)
dask.awkward<from-delayed, npartitions=2>
>>> uproot.dask([filename1,filename2],library='ak', open_files=False)
dask.awkward<from-delayed, npartitions=2>
>>> uproot.dask([filename1,filename2],library='ak', open_files=False).compute()
<Array [{Type: 'GT', Run: 148031, ...}, ...] type='4608 * {Type: string, Ru...'>
```

WIP:
Tests remain
A few things to cleanup (move dask_awkward to extras, make awkward the default library to use)


src/uproot/behaviors/TBranch.py
```diff
@@ -3548,6 +3575,151 @@ def delayed_get_array(ttree, key):
...
+            stop = min(start + entry_step, entry_stop)
+            length = stop - start
+
+            delayed_array = del_fn(ttree, start, stop)
+            dask_arrays.append(delayed_array)
+
+        for start in range(entry_start, entry_stop, entry_step):
+            foreach(start)
+
+    return dask_awkward.from_delayed(dask_arrays)
```
kkothari2001, line 3694: Relevant dask section

src/uproot/behaviors/TBranch.py
```diff
@@ -3548,6 +3575,151 @@ def delayed_get_array(ttree, key):
...
+    for file_path, object_path in files:
+        delayed_tree = delayed_open_fn(
+            file_path, object_path, custom_classes, allow_missing, real_options
+        )
+        delayed_array = delayed_get_array(delayed_tree)
+
+        dask_arrays.append(
+            delayed_array
+        )
+    return dask_awkward.from_delayed(dask_arrays)
```
kkothari2001, line 3726: Relevant dask section

douglasdavis:
Hey @kkothari2001 thanks for the ping! I'll take a look at the PR in detail when I get a chance. 

But in the mean time I wanted to point you to a proof-of-concept for reading ROOT into dask-awkward here: https://github.com/ContinuumIO/dask-awkward/pull/44 - the method I used in that PoC does not use delayed, directly constructing so-called "high level" task graphs (dask-awkward has a generic API for creating dask-awkward arrays from any mapping with `dask_awkward.from_map`). The main benefit of the `from_map` API over using delayed is that the task graph created by `from_map` can be optimized in ways that task graphs of delayed objects cannot.

src/uproot/behaviors/TBranch.py
```diff
@@ -3549,6 +3578,187 @@ def delayed_get_array(ttree, key):
...
+
+    first5 = hasbranches[0].arrays(common_keys, entry_start=0, entry_stop=5)
+    meta = dask_awkward.core.typetracer_array(first5)
+
+    return dask_awkward.from_map(
+        _UprootRead(hasbranches, common_keys),
+        partition_args,
+        label="from-uproot",
+        meta=meta,
+    )
```
kkothari2001, line 3703: Relevant dask part cc @douglasdavis

src/uproot/behaviors/TBranch.py
```diff
@@ -3549,6 +3578,187 @@ def delayed_get_array(ttree, key):
...
+
+    first5 = obj.arrays(common_keys, entry_start=0, entry_stop=5)
+    meta = dask_awkward.core.typetracer_array(first5)
+
+    return dask_awkward.from_map(
+        _UprootOpenAndRead(custom_classes, allow_missing, real_options, common_keys),
+        files,
+        label="from-uproot",
+        meta=meta,
+    )
```
kkothari2001, line 3759: Relevant dask part cc @douglasdavis

src/uproot/behaviors/TBranch.py
kkothari2001: Notes: First I tried using the same code to create the typetracer array with `ak.Array(first5.layout.typetracer.forget_length())` but it was throwing an error that made me realise it probably wasn't the correct way to get the tytracer array.

Going through the code for dask-awkward I found this function `dask_awkward.core.typetracer_array`, it works, but please confirm if I'm using it right.

src/uproot/behaviors/TBranch.py
```diff
@@ -598,8 +598,8 @@ def dask(
     """
     files = uproot._util.regularize_files(files)
     library = uproot.interpretation.library._regularize_library(library)
-    if library.name != "np":
-        raise NotImplementedError()
+    # if library.name != "np":
+    #     raise NotImplementedError()
```
jpivarski, line 602: In a committed PR, old code should be removed, not commented out.

But I think you still want to complain about Pandas.

```suggestion
    if library.name == "pd":
        raise NotImplementedError()
```

src/uproot/behaviors/TBranch.py
```diff
@@ -609,31 +609,60 @@ def dask(
...
-            filter_typename,
-            filter_branch,
-            recursive,
-            full_paths,
-            step_size,
-            custom_classes,
-            allow_missing,
-            real_options,
-        )
+    if library.name == "np":
```
jpivarski, line 612: We want the code in TBranch.py to avoid special cases that depend on the value of `library`; that's what `library.*` methods are for. Can this be refactored to call `library.something` here, and then have the different `library` implementations do the different things?

src/uproot/behaviors/TBranch.py
```diff
@@ -3549,6 +3578,187 @@ def delayed_get_array(ttree, key):
...
+    class _UprootRead:
+        def __init__(self, hasbranches, branches) -> None:
+            self.hasbranches = hasbranches
+            self.branches = branches
+
+        def __call__(self, i_start_stop):
+            i, start, stop = i_start_stop
+            return self.hasbranches[i].arrays(
+                self.branches, entry_start=start, entry_stop=stop
+            )
```
jpivarski, line 3675: I think this class can be defined at top-level, outside of this function. That might be necessary to serialize it and send it to a remote machine. (In general, nested classes and functions can depend on local variables, which can prevent them from being pickleable, which Dask needs to send data to remote processors.)

Even if that's not necessary for this helper function, having a common practice of making functions and classes top-level by default would make it so that we don't have to ask ourselves this questions whenever we see a nested function/class.

If you want to introduce a new src/uproot/_dask.py to have a place to put all of these helpers, that would be good. TBranch.py is getting very full (and it's always easier to merge files than split them).

src/uproot/behaviors/TBranch.py
jpivarski: What motivated the number 5? Is it possible for these 5 entries to not have something that you're looking for, and in that case, does it fail?

Since reading happens at a per-TBasket level, getting the whole first TBasket might make more sense. [TBranch.basket_entry_start_stop](https://uproot.readthedocs.io/en/latest/uproot.behaviors.TBranch.TBranch.html#basket-entry-start-stop) would give you the `entry_start` and `entry_stop` numbers for a given TBasket (such as number `0`).

src/uproot/behaviors/TBranch.py
```diff
@@ -3549,6 +3578,187 @@ def delayed_get_array(ttree, key):
...
+    )
+    common_keys = obj.keys(
+        recursive=recursive,
+        filter_name=filter_name,
+        filter_typename=filter_typename,
+        filter_branch=filter_branch,
+        full_paths=full_paths,
+    )
+
+    class _UprootOpenAndRead:
```
jpivarski, line 3731: Another nested class that might not need to be nested.

jpivarski (commented):
This is looking good; sorry that I was slow to respond. Most of what I recommend below is refactoring; you can introduce src/uproot/_dask.py if you need a place to put all the new helper functions (and that would help keep most of this separate from the rest of TBranch.py, which is mostly not involved in anything Dask-related).

The tests look good—I don't think we can test dask-distributed in our continuous integration, but can you test it manually? Like, if you use these tests and switch to a multiprocessing Dask scheduler, that would verify that everything that needs to be serializable is serializable.

src/uproot/behaviors/TBranch.py
douglasdavis: > I think this class can be defined at top-level, outside of this function.

Yes! it's definitely best to have dask callables as standalone classes or functions.

tests/test_0652_dask-for-awkward.py
```diff
@@ -0,0 +1,96 @@
...
+def test_single_dask_awkward_array():
+    test_path = skhep_testdata.data_path("uproot-Zmumu.root") + ":events"
+    ttree = uproot.open(test_path)
+
+    ak_array = ttree.arrays()
+    dak_array = uproot.dask(test_path, library="ak")
+
+    assert len(ak_array) == len(dak_array)
+
+    assert dak_array.compute().to_list() == ak_array.to_list()
```
douglasdavis, line 22: When comparing a dask-awkward array to a concrete awkward array try to use [dask-awkward's `assert_eq`](https://github.com/ContinuumIO/dask-awkward/blob/main/src/dask_awkward/testutils.py#L18); it does more checks than just for equality

jpivarski (approved):
We discussed some final refactorings on Zoom, but when you're done with them, feel free to squash-and-merge!

src/uproot/extras.py
```diff
@@ -270,3 +270,20 @@ def dask():
...
+    Imports and returns ``dask_awkward``.
+    """
+    try:
+        import dask_awkward
+    except ModuleNotFoundError:
+        raise ModuleNotFoundError(
+            """for uproot.dask, install the (complete) 'dask' package with:
+    pip install "dask[complete]"
+or
+    conda install dask"""
```
masonproffitt, line 286: This error text doesn't make sense here. Installing `dask[complete]` won't install `dask_awkward`.

PR #660, opened 2022-08-03
Set ReadOnlyDirectory attributes when fSeekKeys == 0

### LINKED ISSUE #658: It's possible for a ReadOnlyDirectory to be created without all fields.
ReadOnlyDirectory has a few attributes that usually get filled with meaningful values when `fSeekKeys != 0`. But in that special case:

https://github.com/scikit-hep/uproot5/blob/914dd25ec03225c5a37da3c30d8d83d68d34fea4/src/uproot/reading.py#L1405-L1407

the `self._keys_lookup` and `self._len` attributes don't get set (any others? scroll down to check). They should be set in any case, possibly before that `if`-branch so that conditional initialization doesn't bite us again.

kakwok:
As titled. Fixes #658.

### DIFF
--- a/src/uproot/reading.py
+++ b/src/uproot/reading.py
@@ -1406,7 +1406,7 @@ class ReadOnlyDirectory(Mapping):
             self._header_key = None
             self._keys = []
             self._keys_lookup = {}
-            self._len = None 
+            self._len = None
         else:
             keys_start = self._fSeekKeys
             keys_stop = min(keys_start + self._fNbytesKeys + 8, file.fEND)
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
#660 and #661 fix the problem, and I'll merge them as soon as #661 passes tests (some technicality; I'm taking care of it as well as I can through the web interface). As soon as they go through, I'll start a new Uproot v4 release.

jpivarski:
@all-contributors please add @kakwok for code

PR #661, opened 2022-08-03
[v4 backport] Set ReadOnlyDirectory attributes when fSeekKeys == 0

### LINKED ISSUE #658: It's possible for a ReadOnlyDirectory to be created without all fields.

kakwok:
Backport of #660 to `main-v4`, fixes #658

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -234,7 +234,7 @@ def regularize_rename(rename):
             def applyrules(x):
                 for matcher, trans in rules:
                     if matcher.search(x) is not None:
-                        return matcher.sub(trans, x)   # noqa: B023
+                        return matcher.sub(trans, x)  # noqa: B023
                 else:
                     return x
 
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
#660 and #661 fix the problem, and I'll merge them as soon as #661 passes tests (some technicality; I'm taking care of it as well as I can through the web interface). As soon as they go through, I'll start a new Uproot v4 release.

jpivarski:
I don't know why the v4 branch is complaining, whereas the v5 branch is not, and these should (ideally) be handled more centrally than tagging each use—that was probably done on the v5 branch. Nevertheless, we don't plan to develop this branch much more, just bug-fixes, so obfuscating it with "`# noqa: B023`" is not a problem here—just don't do it in the `main` (v5) branch!

PR #662, opened 2022-08-03
Implement stl containers for RNTuple

Moelf:
See:
- https://github.com/root-project/root/blob/master/tree/ntuple/v7/doc/specifications.md#stl-types-and-collections

And similar existing tests for TTree:
- https://github.com/scikit-hep/uproot5/blob/main/tests/test_0031-test-stl-containers.py
- https://github.com/scikit-hep/uproot5/blob/main/tests/test_0033-more-interpretations-2.py


Moelf:
## Variant (Union) logic

We make a field with type `std::variant<std::int32_t,double>`,and actual content: `[1.0, 4, 3, 2, 1]`

the field and column records look like the following:
```python
# fields
([MetaData('Field, parent_field_id=0, struct_role=3, field_name='variant_int32_double', type_name='std::variant<std::int32_t,double>', type_alias='', field_desc=''),
  MetaData('Field, parent_field_id=0, struct_role=0, field_name='_0', type_name='std::int32_t', type_alias='', field_desc=''),
  MetaData('Field, parent_field_id=0, struct_role=0, field_name='_1', type_name='double', type_alias='', field_desc='')],
# columns
 [MetaData('ColumnRecordFrame', type=3, nbits=64, field_id=0, flags=0),
  MetaData('ColumnRecordFrame', type=11, nbits=32, field_id=1, flags=0),
  MetaData('ColumnRecordFrame', type=7, nbits=64, field_id=2, flags=0)])
```

Let's focus on the `Switch` column, which has column id 0, treating Switch column's content as `uint64`, we see `[8589934592 4294967296 4294967297 4294967298 4294967299]`, since each number is split into:
>Lower 44 bits like kIndex64, higher 20 bits are a dispatch tag to a column ID

 we look at the raw bits of these five numbers `numpy.unpackbits`:
```
0000000000000000000000000000000000000010000000000000000000000000
0000000000000000000000000000000000000001000000000000000000000000
0000000100000000000000000000000000000001000000000000000000000000
0000001000000000000000000000000000000001000000000000000000000000
0000001100000000000000000000000000000001000000000000000000000000
```

since the last four elements in actual content are all `int32`, I would expect the "higher 20 bits" to be the same for the last four numbers, but that doesn't make sense, 20 bits from right-hand-side don't make to where "1" is.

ideas? @jpivarski

Moelf:
hmm

https://github.com/root-project/root/blob/cc28da57f0b8e2878a2c047504a49762d7d8dfb5/tree/ntuple/v7/src/RColumnElement.cxx#L124-L126

Moelf:
```python
> ipython -c 'import numpy as np; from pprint import pprint; import skhep_testda
ta; import uproot as up; import awkward as ak; filename = skhep_testdata.data_path("test_ntuple_stl_containers.root
"); r = up.open(filename)["ntuple"]; pprint(r.arrays().to_list());'
[{'string': 'one',
  'tuple_int32_string': {'_0': 1, '_1': 'one'},
  'variant_int32_float': 1.0,
  'vector_int32': [1],
  'vector_string': ['one'],
  'vector_tuple_int32_string': [{'_0': 1, '_1': 'one'}],
  'vector_variant_int32_float': [1],
  'vector_vector_int32': [[1]],
  'vector_vector_string': [['one']]},
 {'string': 'two',
  'tuple_int32_string': {'_0': 2, '_1': 'two'},
  'variant_int32_float': 2.0,
  'vector_int32': [1, 2],
  'vector_string': ['one', 'two'],
  'vector_tuple_int32_string': [{'_0': 1, '_1': 'one'}, {'_0': 2, '_1': 'two'}],
  'vector_variant_int32_float': [1, 2.0],
  'vector_vector_int32': [[1], [2]],
  'vector_vector_string': [['one'], ['two']]}]
```

src/uproot/models/RNTuple.py
```diff
@@ -52,6 +51,9 @@ def _renamemeA(chunk, cursor, context):
 
 
 class Model_ROOT_3a3a_Experimental_3a3a_RNTuple(uproot.model.Model):
+    def debug(self):
+        return self.header.field_records, self.header.column_records
```
jpivarski, line 55: We can leave in `debug` methods if they provide enough value to future maintainers (ourselves). This one, though, doesn't provide much benefit above fetching these two properties manually, so it should be dropped.

src/uproot/models/RNTuple.py
```diff
@@ -218,26 +220,62 @@ def page_list_envelopes(self):
 
         return self.column_innerlist_dict
 
-    def col_form(self, field_id):
-        for cr in self.header.column_records:
+    def base_col_form(self, col_id):
+        cr = self.header.column_records[col_id]
+        form_key = f"column-{col_id}"
+        if cr.type == 3:  # switch (UnionForm)
```
jpivarski, line 226: This "`3`" and all others like it (e.g. the "`2`" a couple of lines down) should all be constants defined in `uproot.const`. They should have a common prefix to indicate that they're for RNTuple.

src/uproot/models/RNTuple.py
```diff
@@ -218,26 +220,62 @@ def page_list_envelopes(self):
...
+                return form_key
+        elif type_name == "std::string":  # string field splits->2 in col records
+            assert len(rel_crs_idxs) == 2
+            cr_char = rel_crs[-1]
+            assert cr_char.type == 5  # char
+            inner = self.base_col_form(rel_crs_idxs[-1])
+            form_key = f"column-{rel_crs_idxs[0]}"
+            return ak._v2.forms.ListOffsetForm(
+                "u32", inner, form_key=form_key, parameters={"__array__": "string"}
+            )
```
jpivarski, line 264: Perhaps you didn't get an error, but the ListOffsetForm has to have `parameters={"__array__": "string"}` _and also_ the NumpyForm has to have `parameters={"__array__": "char"}`. It's a non-local requirement (an invariant that applies to sets of nodes, rather than a single node), so it's not checked in the node constructor, but [ak.is_valid](https://awkward-array.readthedocs.io/en/latest/_auto/ak.is_valid.html) checks for it.

src/uproot/models/RNTuple.py
```diff
@@ -323,22 +378,48 @@ def arrays(
...
+                # split Switch column into index and tag
+                # TODO what's this -1 hack
+                tags = (content >> 44).astype("int8") - 1
+                kindex = numpy.bitwise_and(content, numpy.int64(0x00000000000FFFFF))
+                D[f"{key}-index"] = kindex
+                D[f"{key}-tags"] = tags
+            else:
+                # don't distinguish data and offsets
+                D[f"{key}-data"] = content
+                D[f"{key}-offsets"] = content
```
jpivarski, line 401: With a small set of exceptions, I'm generally against using `break` and `continue` to express logic (as well as early `return` from functions, unless it's handling a special case). These keywords are the last vestiges of `goto`.

(Note: [I have used goto](https://github.com/scikit-hep/awkward/blob/af0f4569ed90eab64690098e5a64ae610282c4bf/src/libawkward/forth/ForthMachine.cpp#L3680-L3685); there are exceptions for most of these rules. Generally, these Bad Things should just be used sparingly.)

```suggestion
            if key in target_cols:
                content = self.read_col_page(i, L)
                if cr.type == 3:
                    # split Switch column into index and tag
                    # TODO what's this -1 hack
                    tags = (content >> 44).astype("int8") - 1
                    kindex = numpy.bitwise_and(content, numpy.int64(0x00000000000FFFFF))
                    D[f"{key}-index"] = kindex
                    D[f"{key}-tags"] = tags
                else:
                    # don't distinguish data and offsets
                    D[f"{key}-data"] = content
                    D[f"{key}-offsets"] = content
```

Also, that `44` and `0x00000000000FFFFF` either need to be named constants _or_ the two lines that split tags and index should be a short, named function. It's all about isolating the "`44`" and "`0x00000000000FFFFF`" and explaining what they do, so they can still be magic numbers if they're inside a well-named function that does only this one thing.

tests/test_0662-rntuple-stl-containers.py
```diff
@@ -0,0 +1,59 @@
...
+        assert r["vector_string"][0] == ["one"]
+        assert ak.all(r["vector_string"][1] == ["one", "two"])
+
+        assert ak.all(r["vector_vector_string"] == [["one"], [["one"], ["two"]]])
+        assert type(r["variant_int32_float"][0]) == numpy.float32
+        assert type(r["variant_int32_float"][1]) == numpy.float32
+
+        assert type(r["vector_variant_int32_float"][0][0]) == numpy.int32
+        assert type(r["vector_variant_int32_float"][1][0]) == numpy.int32
+        assert type(r["vector_variant_int32_float"][1][1]) == numpy.float32
```
jpivarski, line 48: Integers and floating point numbers are close enough that this test (checking `type`) could legitimately break in the future, if Awkward takes a new policy on return types. (This, for instance, would fail in Awkward v1. It's recognized as a backward-incompatible change, of course.) Awkward's `simplify_uniontype` would turn a union of integers and floating point numbers into an array of floating point numbers, following NumPy's type promotion rules, and maybe that would get used on this someday.

A better test of variant would be integers and strings: those are very different from one another and would never be merged upon `simplify_uniontype`.

jpivarski (changes requested):
This is looking nice! I have some small requests below.

src/uproot/models/RNTuple.py
Moelf: In this case I did the original way to avoid indenting many lines without an else branch

jpivarski: Right, that's understandable. The indentation as syntax thing fails for long blocks, and these are a bit long. (There are longer ones elsewhere in Uproot...) An editor that displays vertical lines at each indentation depth can help, though admittedly I don't use one.

tests/test_0662-rntuple-stl-containers.py
Moelf: let me make a new stl container test file and make a PR to the test file repo first...

Moelf:
@jpivarski eh, pre-commit CI is doing some stripping and then flake8 errors on the stripped stuff, what's going on?

jpivarski:
Today's meeting will be a good chance to ask @henryiii about that. I've had to hold back pre-commit in #657 because the update is broken; I wonder if this is a partial update? It's definitely an incompatibility between the part that removes unnecessary "`# noqa`" and the part that needs those noqas.

Moelf:
I wonder if changing the order helps? idk how it determines "unnecessary" without flake8 running through it

tests/test_0662-rntuple-stl-containers.py
Moelf: addressed in https://github.com/scikit-hep/scikit-hep-testdata/pull/82

src/uproot/const.py
```diff
@@ -114,3 +114,9 @@
 ############# other
 
 kStreamedMemberWise = numpy.uint16(1 << 14)
+
+############ RNTuple https://github.com/root-project/root/blob/master/tree/ntuple/v7/doc/specifications.md
+rt_role_leaf = 0
+rt_role_vec = 1
+rt_role_struct = 2
+rt_role_union = 3
```
jpivarski, line 122: No reason to be so brief:

```suggestion
rntuple_role_leaf = 0
rntuple_role_vec = 1
rntuple_role_struct = 2
rntuple_role_union = 3
```

But then you have to change their usages below. I found a few more examples of literal values that should be replaced with constants.

src/uproot/models/RNTuple.py
```diff
@@ -218,26 +217,64 @@ def page_list_envelopes(self):
 
         return self.column_innerlist_dict
 
-    def col_form(self, field_id):
-        for cr in self.header.column_records:
+    def base_col_form(self, col_id, parameters=None):
+        cr = self.header.column_records[col_id]
+        form_key = f"column-{col_id}"
+        if cr.type == uproot.const.rt_role_union:  # switch (UnionForm)
```
jpivarski, line 222: With named constants, comments like this one become unnecessary.

src/uproot/models/RNTuple.py
```diff
@@ -218,26 +217,56 @@ def page_list_envelopes(self):
...
+            if cr.field_id > field_id:
+                break
+        if len(rel_crs) == 1:  # base case
+            return self.base_col_form(
+                rel_crs[0], rel_crs_idxs[0]
+            )
+        elif type_name == "std::string":  # string field splits->2 in col records
+            assert len(rel_crs_idxs) == 2
+            cr_char = rel_crs[-1]
+            assert cr_char.type == 5  # char
```
jpivarski, line 246: Named constant.

src/uproot/models/RNTuple.py
```diff
@@ -218,26 +217,56 @@ def page_list_envelopes(self):
...
+        sr = fr.struct_role
+        if sr == 0:
             # base case of recursive
-            # these two roles have exactly one column belong to them
-            return self.col_form(this_id)
-        elif fr.struct_role == 1:
+            # n.b. the split may be in column
+            return self.col_form(this_id, fr.type_name)
+        elif sr == 1:
+            # mother of a collection, i.e. vector (meaning field is jagged)
```
jpivarski, line 267: More instances of named constants ("`0`" and "`1`").

------------------

But also, why abbreviate names like `self.header.field_records` → `frs`? It makes me have to go look them up. From the perspective of writing code, using long, fully qualified names feels redundant and you want to adopt a shorthand. But from the perspective of reading code, especially when hunting a bug or something where I'm only interested in a small local part of the codebase, answering questions like "what does "`frs`" mean?" is an extra burden.

I prefer fully qualified names, even if they're pretty long, to prioritize the needs of code-readers (over code-writers). I'm particularly aggravated when reading Python code that has a lot of `from some.library import SomeClass` at the top of the file, so when I encounter `SomeClass`, I have to go up to the top of the file to find out what it is and where it's defined. Maybe code analysis tools help with that, but those tools will get confused if there's too much/too complex indirection. I had to be won over to the idea of `import numpy as np`, but became okay with that when I saw how standard that particular substitution was. It's legal to `import numpy as qz`, but that would be terrible.

The downside of using fully qualified names almost everywhere is that code formatters limit the line length. I haven't entirely figured out how to reconcile the value of fully qualified names with how much black obfuscates a line when it's really long. So you don't have to use

```python
self.header.field_records[this_id]
```

everywhere you currently use

```python
fr
```

and get those deeply nested black-formattings. But maybe the substitutions don't have to be 2 and 3 letters long:

```python
field_records = self.header.field_records
field_record = field_records[this_id]
```

(and use them consistently in any function that uses it, kind of like how "`np`" always means "`numpy`")?

src/uproot/models/RNTuple.py
```diff
@@ -247,8 +274,8 @@ def field_form(self, this_id, seen):
             )
             inner = self.field_form(child_id, seen)
             return ak._v2.forms.ListOffsetForm("u32", inner, form_key=keyname)
-        elif fr.struct_role == 2:
-            # struct field
+        elif sr == 2:
```
jpivarski, line 277: Named constant.

src/uproot/models/RNTuple.py
```diff
@@ -257,6 +284,15 @@ def field_form(self, this_id, seen):
             recordlist = [self.field_form(i, seen) for i in newids]
             namelist = [frs[i].field_name for i in newids]
             return ak._v2.forms.RecordForm(recordlist, namelist, form_key="whatever")
+        elif sr == 3:
```
jpivarski, line 287: Named constant.

src/uproot/models/RNTuple.py
```diff
@@ -257,6 +284,15 @@ def field_form(self, this_id, seen):
...
+            keyname = self.col_form(this_id)
+            newids = []
+            for i, fr in enumerate(frs):
+                if i not in seen and fr.parent_field_id == this_id:
+                    newids.append(i)
+            recordlist = [self.field_form(i, seen) for i in newids]
+            return ak._v2.forms.UnionForm("i8", "i64", recordlist, form_key=keyname)
         else:
             # everything should recursive above this branch
             raise (AssertionError("this should be unreachable"))
```
jpivarski, line 298: No biggie, but Python doesn't need all of these parentheses.

```suggestion
            raise AssertionError("this should be unreachable")
```

To minimize noise, the fewer the better. I don't know why black didn't do this automatically. (It's supposed to keep the AST unchanged, and removing parentheses would keep the AST unchanged; it's an AST, not a parsing tree!)

src/uproot/models/RNTuple.py
```diff
@@ -266,9 +302,7 @@ def to_akform(self):
         recordlist = []
         topnames = self.keys()
         seen = []
-        for (i, fr) in enumerate(frs):
-            if fr.parent_field_id == i:
-                topnames.append(fr.field_name)
+        for (i, _) in enumerate(frs):
```
jpivarski, line 305: Also unnecessary parentheses. Better style to let them hang free.

```suggestion
        for i, _ in enumerate(frs):
```

tests/test_0662-rntuple-stl-containers.py
```diff
@@ -0,0 +1,60 @@
...
+        assert ak.all(r["vector_string"][1] == ["one", "two"])
+
+        assert ak.all(r["vector_vector_string"] == [["one"], [["one"], ["two"]]])
+        assert type(r["variant_int32_string"][0]) == numpy.int32
+        assert type(r["variant_int32_string"][1]) == str
+
+        assert r["vector_variant_int64_string"][0][0] == "one"
+        assert r["vector_variant_int64_string"][1][0] == "one"
+        assert r["vector_variant_int64_string"][1][1] == 2
+        assert type(r["vector_variant_int64_string"][1][1]) == numpy.int64
```
jpivarski, line 49: For the variant tests, could you test the values, rather than the types? There's a 1% chance we'll change the type of

```python
r["variant_int32_string"][0]
```

from `np.int32` to `int` or `ak.Scalar` or something. (In Awkward v1, it was `int`, but I think we're doing the right thing now by making it `np.int32`. The `ak.Scalar` thing was talked about.)

It would be a more "what we care about" test to say

```python
        assert r["variant_int32_string"][0] == 1
        assert r["variant_int32_string"][1] == "two"
```

(though I don't know the actual values).

jpivarski (changes requested):
The only actual "request to change" are the magic numbers → named constants and expanding the names of those constants. Comments on style are just generally to be taken into consideration, not a must-have for this PR.

src/uproot/models/RNTuple.py
Moelf: in this case the `5` is in `_rntuple_col_types`, do we want to have a reverse look up table?

jpivarski: Reverse lookup tables can be useful and they can be built at startup, like this `_primitive_to_dtype_dict` and `_dtype_to_primitive_dict` pair:

https://github.com/scikit-hep/awkward/blob/af0f4569ed90eab64690098e5a64ae610282c4bf/src/awkward/_v2/types/numpytype.py#L66-L93

If that's what's needed here, go ahead and make the reverse lookup.

jpivarski: (In my example, the use of the word "dict" in the variable name is because there are also functions:

https://github.com/scikit-hep/awkward/blob/af0f4569ed90eab64690098e5a64ae610282c4bf/src/awkward/_v2/types/numpytype.py#L23-L56

That part didn't need to be copied into `rntuple_col_num_to_dtype_dict` and `rntuple_col_type_to_num_dict`, but it's fine either way.)

jpivarski (approved):
This is looking great! I like the idea of moving the whole `rntuple_col_num_to_dtype_dict` to src/uproot/const.py; I agree that it belongs there. It's great to be able to cross-link it to the specification.

You can merge this PR whenever you're ready.

PR #679, opened 2022-08-12
feat: `from_map` like optimization for dask arrays

kkothari2001:


jpivarski:
@kkothari2001, you'll want to merge `main` (get everything) or `pre-commit-ci-update-config` (just get the flake8-fix) to fix the tests.

jpivarski:
@douglasdavis, you said that you want to subscribe to this PR and check @kkothari2001's usage of `from_map` and high-level graphs.

src/uproot/_dask.py
```diff
@@ -168,6 +170,162 @@ def dask(
         raise NotImplementedError()
 
 
+class _PackedArgCallable:
```
douglasdavis, line 173: I'm wondering if uproot should use dask-awkward's implementation of `_PackedArgCallable` rather than rewriting it. What's the dependency situation right now? If a user wants to use `uproot.dask` do we already require them to have dask-awkward installed?

douglasdavis:
I think this looks good as a replacement for the arrays backed by low level graphs. I think there is potential to squash some code duplication, but that would require the answer to my inline question above to be "yes" and I just realized one can use `uproot.dask` without having dask-awkward installed, so I think the answer is "no" :) So I think this is good, and perhaps there will be a future where we can remove some code duplication if some of the dask-awkward from_map logic can eventually make it into upstream dask.

src/uproot/_dask.py
jpivarski: > If a user wants to use `uproot.dask` do we already require them to have dask-awkward installed?

Currently no: if they call `uproot.dask` with `library="np"`, they don't need dask-awkward, but if `library="ak"`, then they do. I wasn't 100% sure of that decision.

You know, dask-awkward doesn't seem to be on conda-forge yet. That might be a blocker for some people. (Personally, I pip-install a minimal number of packages into a conda environment, if there's no conda equivalent.)

> I'm wondering if uproot should use dask-awkward's implementation of `_PackedArgCallable` rather than rewriting it.

I don't know; this is pretty small and straightforward:

```python
class _PackedArgCallable:
    """Wrap a callable such that packed arguments can be unrolled.
    Inspired by dask.dataframe.io.io._PackedArgCallable.
    """

    def __init__(self, func, args=None, kwargs=None, packed=False):
        self.func, self.args, self.kwargs, self.packed = func, args, kwargs, packed

    def __call__(self, packed_arg):
        if not self.packed:
            packed_arg = (packed_arg,)
        return self.func(*packed_arg, *(self.args or []), **(self.kwargs or {}))
```

But if we do want to pick it up from dask-awkward, it will have to have a non-underscored name.

Going in the other direction, if this came from Dask DataFrame and now both Dask-Awkward and Uproot need it, maybe it should be a public API function in Dask? (I know that that's not a quick change to make. But in principle?)

src/uproot/_dask.py
```diff
@@ -168,6 +170,162 @@ def dask(
...
+    return dask.array.core.Array(hlg, name, chunks, dtype=dtype)
+
+
+class _UprootReadNumpy:
+    def __init__(self, hasbranches, key) -> None:
+        self.hasbranches = hasbranches
+        self.key = key
+
+    def __call__(self, i_start_stop):
+        i, start, stop = i_start_stop
```
jpivarski, line 323: This makes me pine for Python 2's [tuple parameter unpacking](https://peps.python.org/pep-3113/)!

jpivarski (approved):
This looks good to me and if @douglasdavis is happy with the Dask usage, then we're ready to merge.

Doug, your point about code duplication is understood, though fixing it would require either a new version of Dask-Awkward or a new version of Dask, so that we can get the function without an underscore. That's beyond this PR, though we can and should come back to it later.

@kkothari2001, do a `git merge main` (with manual resolution) to fix the tests, and then this should be ready to go.

src/uproot/_dask.py
douglasdavis: The version in dask is inside of dask.dataframe (which if used will require `import pandas`); but now I'm thinking it should be moved out of `dask.dataframe` and into `dask`!

Issue #668, opened 2022-08-10
The library="pd" option for ragged arrays should return ragged arrays wrapped by awkward-pandas


jpivarski:
Currently, `library="pd"` explodes ragged arrays and sets of ragged arrays to one or more DataFrames using MultiIndex, in a _reimplementation_ of [ak.to_pandas](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_pandas.html) (the reimplementation does not rely on `awkward` being installed). This has been a cause of confusion on multiple occasions, especially when the ragged arrays would have incompatible MultiIndexes; here's one occasion: #583.

With https://github.com/intake/awkward-pandas, it will be possible to simply wrap each ragged array as a `pd.Series`, which can always be a grouped into a single `pd.DataFrame`, regardless of the ragged structure. We should make this the new behavior in v5. It's backward-incompatible, but still possible to reproduce the old behavior via `library="ak"` and `ak.to_dataframe` (see https://github.com/scikit-hep/awkward/issues/1546), and it will remove a lot of complex code.

Also, @kkothari2001 won't have to re-re-implement the explosion logic for `uproot.dask` with `library="pd"`.

PR #680, opened 2022-08-18
Installation requirements and error text for dask/dask-awkward in extras.py

jpivarski:
This is addressing https://github.com/scikit-hep/uproot5/pull/652#discussion_r948495191 (@masonproffitt).

@kkothari2001, please make sure that `uproot.extras.dask_awkward` is only called if `library != "np"` (i.e. `uproot.dask` with `library="np"` requires Dask, but `uproot.dask` with `library="ak"` requires Dask-Awkward). If Dask-Awkward strictly depends on `dask[complete]` (I haven't checked: please check it!), then the imports in the `dask_awkward()` function can be reduced to just `import dask_awkward`. We want to minimize the number of error messages that say, "Stop what you're doing and go install some other library," but not require people to install things that they don't need. (Consider the example of `lz4` and `xxhash`, which are always needed together, so the error message for both says to install both.)

I'm not sure what will be required for the `library="pd"` case; quite possibly Awkward-Pandas. (If we no longer explode ragged data into a MultiIndex but wrap them using Awkward-Pandas, as described in #668, then every function with `library="pd"` may need Awkward-Pandas.)


jpivarski:
Actually, `dask-awkward` is not on `conda-forge` yet, so it needs to be pip-installed even if one's main installed is conda (and if Dask is installed via conda).

jpivarski:
Replaced by #690.

PR #683, opened 2022-08-24
ci: use mamba

henryiii:
This should be faster.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -35,14 +35,15 @@ jobs:
         with:
           auto-update-conda: true
           python-version: ${{ matrix.python-version }}
-          channels: "conda-forge"
+          miniforge-variant: Mambaforge
+          use-mamba: true
 
       - name: "Install most dependencies"
         run: |
           conda env list
           conda info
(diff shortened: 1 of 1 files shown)


henryiii:
`tests/test_0637-setup-tests-for-AwkwardForth.py::test_03` timed out (after 6 hours) on Python 3.8 & 3.10, but not 3.9, on Windows. Is this a known flake? Shouldn't this be time capped via pytest-timeout?

henryiii:
Every check is under 10 minutes now. It used to take more than that just to install ROOT. :)

jpivarski (approved):
(I fixed the merge conflicts.)

I'd be interested to see what the time is now that we're only getting ROOT and XRootD from conda.

But this is a good thing. We used to use mamba to speed up testing in Travis, and often had to revert it to conda because it didn't always work. Nowadays, mamba is a lot more trustworthy (I use it all the time on my desktop).

I'll enable squash-and-merge.

jpivarski:
@ianna, this is a good model to look at for installing ROOT in CI. (For the ROOT CI job in Awkward for testing your RDataFrame bridge.)

PR #684, opened 2022-08-24
fix: depend on packaging, not setuptools vendored packaging

henryiii:
This should not be pulled from setuptools vendoring, it's a standalone package. And `importlib.metadata.version` is the correct way to get the version (or from the backport `importlib_metadata`).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ packages = find:
 install_requires =
     awkward>=1.9.0rc1
     numpy
-    setuptools
+    packaging
 python_requires = >=3.6
 package_dir =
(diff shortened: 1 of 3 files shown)


jpivarski:
For the sake of cross-referencing, this seems to have the same goal as https://github.com/scikit-hep/awkward/pull/1562, but it's less complicated.

jpivarski (approved):
`:)`

henryiii:
I have a sneaky objective in this one, unlike https://github.com/scikit-hep/awkward/pull/1562 where I'm completely honest and above board. ;)

PR #685, opened 2022-08-24
ci: autocancel repeated runs

henryiii:
Saves some time for repeated pushes (including pre-commit fixes). Also allows Actions to run on a branch-to-branch PR, or manually.


jpivarski (approved):
Sounds good.

PR #688, opened 2022-08-24
feat: move to hatchling

henryiii:
Mostly an automated move using `pipx run hatch new --init`. I cleaned up a tiny bit after that, moved pytest (and expanded using the dev pages guidelines) and isort, moved flake8 to it's own sad little file, and deleted things that are no longer needed. (I'm guessing `build-uproot4.sh` is not used anymore).


henryiii:
(This was my secret motivation for #684)

jpivarski:
I see! Is this done; was it that easy?

henryiii:
🤷 All I could think of, at least. The only other thing is the conda-forge recipe will need an update after this gets released.

jpivarski:
Then perhaps I should include this and make an Uproot release. From `main`, I would only be releasing a v5 pre-release; maybe that doesn't count for what conda-forge picks up. (I've noticed that it doesn't follow pre-releases.)

jpivarski (approved):

PR #691, opened 2022-08-25
ci: use concurrency group for `semantic-pr-title`

agoose77:
The semantic-pr-title workflow does not currently run in a concurrency group. This means that subsequent pushes do not cancel in-progress actions. The following PR fixes this.

### DIFF
--- a/.github/workflows/semantic-pr-title.yml
+++ b/.github/workflows/semantic-pr-title.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: semantic-pr-title-${{ github.head_ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/semantic-pr-title.yml
```diff
@@ -8,6 +8,10 @@ on:
       - synchronize
   workflow_dispatch:
 
+concurrency:
+  group: semantic-pr-title-${{ github.head_ref }}
```
henryiii, line 12:
```suggestion
  group: ${{ github.workflow }}-${{ github.head_ref }}
```

Exactly the same, but more reusable.

PR #644, opened 2022-07-06
feat: Finalizing AwkwardForth reader for Uproot

aryan26roy:

PR #701, opened 2022-09-02
Add a test for a custom interpreted branch with forth

tamasgal:
This adds a tiny test on an existing sample ROOT file and increases the code coverage in `streamers.py` from 65% to 88%.

### DIFF
--- a/src/uproot/_awkward_forth.py
+++ b/src/uproot/_awkward_forth.py
@@ -5,6 +5,7 @@ This module defines utilities for adding components to the forth reader.
 """
 
 import numpy as np
+
 import uproot.containers
 
 symbol_dict = {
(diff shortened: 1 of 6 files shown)


jpivarski:
It's hard for me to scan, but it looks like the only non-formatting change is the addition of `test_82` in tests/test_0637-setup-tests-for-AwkwardForth.py.

Aha: that's what it is. pre-commit went crazy with formatting. I don't know why that would happen, since both this PR and #644 ought to be seeing the same pre-commit configurations.

But maybe @aryan26roy can just manually add the new test (commit f9e6254dea0c906a75bed9075c860920317b55b2)?

tamasgal:
Ah, yes the pre-commit went crazy, oh dear 😆 I pushed the second test. I created this branch off of `aryan-forth-reader-latest` so it should be fine, no idea what happened with the pre-commit.

Yes, adding them manually is fine with me.

tamasgal:
I will close this. The only relevant test is posted here: https://github.com/scikit-hep/uproot5/pull/644#issuecomment-1235917741

PR #702, opened 2022-09-03
docs: `uproot.dask` docs

kkothari2001:
Documentation for `uproot.dask` incuding changes to the getting started guide.


jpivarski:
Before you start working on this, remember to "Update branch".

kkothari2001:
Thanks @jpivarski. I'll need to do that. But since this PR is getting merged in December, I think we should keep this PR open for now and then eventually close it. I can do a separate PR later (with the same text copy-pasted from here). That'll save us from having to merge with main every time and also allow us to alter `_dask.py` without worrying about merge conflicts in the future. Since this is mainly just a doc update, I believe it will be easier to do that.

I also notice that I have not actually pushed any of the docs updates I have written, I'll just commit and push them shortly.

jpivarski:
@kkothari2001, I'm taking this out of draft mode to finish it up, so that it's ready for the 2.0 release. (That is, I'll take over the PR now; no need to push anything, but if you have any commits in your local git repo, let me know what they are by Slack so that I can merge them in.)

jpivarski:
It's done! Anyone looking at the Uproot docs will see `uproot.lazy` disappear and `uproot.dask` appear a week early, but everything else is unchanged. I think that's okay; it won't interrupt work the way that the much more significant changes in the Awkward docs would, if we were to release those early.

Oh, and thanks @kkothari2001 for getting this started!

Issue #697, opened 2022-09-01
`uproot.dask` raises error on empty branches


masonproffitt:
See example below. The exact file used is <https://github.com/iris-hep/func_adl_uproot/blob/61b35593b4512dfaf3a75054aa13ffe2fbc602b6/tests/empty_branches_tree_file.root>. It's a perfectly valid tree; there are just no entries, so I expect to get empty arrays rather than an error.

```python
>>> import uproot
>>> uproot.__version__
'5.0.0rc2'
>>> uproot.dask('tests/empty_branches_tree_file.root')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/miniconda3/envs/func_adl_uproot_rc/lib/python3.10/site-packages/uproot/_dask.py", line 143, in dask
    return _get_dak_array(
  File "/home/user/miniconda3/envs/func_adl_uproot_rc/lib/python3.10/site-packages/uproot/_dask.py", line 483, in _get_dak_array
    ].basket_entry_start_stop(0)
  File "/home/user/miniconda3/envs/func_adl_uproot_rc/lib/python3.10/site-packages/uproot/behaviors/TBranch.py", line 2112, in basket_entry_start_stop
    raise IndexError(
IndexError: branch 'int_branch' has 0 baskets; cannot get starting entry for basket 0
in file tests/empty_branches_tree_file.root
```

jpivarski:
That's right, and also this is handled correctly by `concatenate`, `arrays` and `array`:

```python
>>> uproot.concatenate('empty_branches_tree_file.root')
<Array [] type='0 * {int_branch: int32, float_branch: float32}'>
>>> uproot.open('empty_branches_tree_file.root:tree').arrays()
<Array [] type='0 * {int_branch: int32, float_branch: float32}'>
>>> uproot.open('empty_branches_tree_file.root:tree/int_branch').array()
<Array [] type='0 * int32'>
>>> next(uproot.open('empty_branches_tree_file.root:tree').iterate())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
StopIteration
```

So it's something in the `uproot.dask` logic, and @kkothari2001 should take a look. I'll put this file in [scikit-hep-testdata](https://github.com/scikit-hep/scikit-hep-testdata) so that we can add it to the tests.

Thanks!

jpivarski:
[...]

PR #700, opened 2022-09-02
fix: Fixes uproot.dask bug with empty branches

kkothari2001:
Fixes #697

PR #703, opened 2022-09-03
fix: Use `from_map` optimization for delayed numpy arrays and add tests with empty branches for the same

kkothari2001:
This PR tackles 2 very similar things together:
1. `from_map` optimisation for dask-numpy arrays with `open_files=False`.
2. Tests for empty arrays with delayed open for dask-numpy and dask-awkward. I added this because the chunking mechanism for delayed-file-open calls is handled differently and might not match the behaviour discussed in #697. These tests increase coverage and assure us the chunks are as expected.

This is PR is ready for review @jpivarski

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -327,6 +327,25 @@ class _UprootReadNumpy:
         )
 
 
+class _UprootOpenAndReadNumpy:
+    def __init__(self, custom_classes, allow_missing, real_options, key):
+        self.custom_classes = custom_classes
+        self.allow_missing = allow_missing
+        self.real_options = real_options
+        self.key = key
+
+    def __call__(self, file_path_object_path):
+        file_path, object_path = file_path_object_path
+        ttree = uproot._util.regularize_object_path(
+            file_path,
+            object_path,
+            self.custom_classes,
+            self.allow_missing,
+            self.real_options,
+        )
+        return ttree[self.key].array(library="np")
+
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
It looks good to me! But, to be honest, I only half-understand what's happening here. Of your "2 very similar things," it sounds like (1) was an optimization we were talking about from the meeting and (2) is an additional case not covered by #700? (Those tests didn't have `open_files=False`, which I know is significantly different from the `open_files=True` default.)

Is this something that you want @douglasdavis to check before merging? If you're comfortable with it, though, I trust your judgement.

kkothari2001:
We have 4 code paths each being handled by separate helper functions, combinatoric-ally 2 (library =np or ak) * 2 (open_files = True or False). Out of 4, 3 already used the `from_map` optimisation. The optimization for the fourth one is added over here. While doing this, I was wondering how `open_files=False` deals with chunks with empty arrays, so I added the tests here too.

I'd love if @douglasdavis  could have a look, feel free to! Although I'm not sure he'd be able to comment. Because all changes in this PR are only in the usage of core Uproot functions like `regularize_object_path` (internal helper used for opening root files) and `ttree.arrays()`.

jpivarski:
If it's using `from_map` (the part I don't know well) in a similar way as the other 3 code paths, then it should be fine because you understand how to use `from_map`. I'll merge it!

PR #710, opened 2022-09-13
refactor: refactor Forth generation

aryan26roy:


jpivarski (approved):
Great! We talked a lot on Zoom, and I have some requested changes, but as the next PR.

We can merge this one when the updated tests pass.

jpivarski:
This last pull from `main` should fix the test (from PR #719).

jpivarski:
Yes, it was just failing due to the bug that #719 fixed. The other update is not important; go ahead and merge as-is.

jpivarski:
I had been assuming that this was merged a month ago. I'm going to see how easy it is to get in now. Doesn't #749 need to be applied on top of these changes? Are they really independent?

PR #713, opened 2022-09-17
fix: use ctx manager to ensure resources are freed

### LINKED ISSUE #712: Threads staying open after `uproot.num_entries`
When using `uproot.num_entries` to read the number of entries for a lot of remote files, I eventually run into `RuntimeError: can't start new thread`. I cannot reproduce this with local files. This is tested with `uproot` version 4.3.5.
```python
import uproot
import threading

for _ in range(10):
    next(uproot.num_entries("https://github.com/scikit-hep/scikit-hep-testdata/"\
            "blob/main/src/skhep_testdata/data/uproot-HZZ.root?raw=true:events"))
    print(threading.active_count())
```
output:
```
2
3
4
5
6
7
8
9
10
11
```

agoose77:
Long term I think we should be dealing with a per-process pool rather than a per-source pool, but that's a bigger task.

Fixes #712 

@alexander-held would you be able to check that this works in your context w/o any changes to `HTTPSource.__del__`?

### DIFF
--- a/src/uproot/models/TTree.py
+++ b/src/uproot/models/TTree.py
@@ -968,6 +968,7 @@ def num_entries(paths):
         paths = [uproot._util.file_object_path_split(paths)]
 
     for i, (file_path, object_path) in enumerate(paths2):
-        yield paths[i][0], paths[i][1], uproot.open(
+        with uproot.open(
             {file_path: object_path}, custom_classes={"TTree": Model_TTree_NumEntries}
-        ).all_members["fEntries"][0]
+        ) as f:
+            yield paths[i][0], paths[i][1], f.all_members["fEntries"][0]


alexander-held (commented):
Thanks! I can confirm this fixes the issue I've been seeing (without changes to `HTTPSource.__del__`).

jpivarski (approved):
This looks like what was needed. Go ahead and merge it if you're done!

Ah, but first, we need to take care of the test failures. @douglasdavis described the fix: a module was moved inside `dask_awkward`. Let me see if there's another PR open with this, which we can merge in here.

jpivarski:
It was this (from @douglasdavis on Slack):

> dask_awkward.testutils has moved to dask_awkward.lib.testutils: https://github.com/ContinuumIO/dask-awkward/pull/76

I'll open a PR for just this, which we can merge against. @Moelf was also seeing this issue, so we need a separate commit that can be merged into all the open PRs.

jpivarski:
It's in #714.

agoose77:
Do we need `myst-nb` or can be use just the `myst-parser` subdependency?

PR #715, opened 2022-09-19
fix: ReadOnlyDirectory should provide the largest abs(cycle) when cycle is unspecified, not the largest cycle.

jpivarski:


jpivarski:
@agoose77, you've already weighed in on what you think in the issue comments, but I'm going to always ask for reviews from now on, even if they're easy. If you agree with this fix, you can also merge it. (What I'm saying here is that I won't be making any more changes, which is what "request to review" is supposed to mean.)

jpivarski:
I just remembered that I'll need to make a corresponding v4 version of this PR. I'll do that soon.

jpivarski:
This passed tests, but I think it was because it ran before the boost-histogram update. Therefore, we'll do the same procedure:

  1. #719 gets approved and merged into `main`.
  2. This PR updates to `main`.
  4. This PR gets approved and merged into `main`.

agoose77:
If #721 gets merged, I'm happy with this being merged.

jpivarski:
I'll take that as an approval of both #721 and #715 and enable auto-merge.

src/uproot/reading.py
```diff
@@ -2047,7 +2047,7 @@ def key(self, where):
                 return key
             elif cycle is None and last is None:
                 last = key
-            elif cycle is None and last.fCycle < key.fCycle:
+            elif cycle is None and abs(last.fCycle) < abs(key.fCycle):
```
jpivarski, line 2050:
```suggestion
            # Follow ROOT's behaviour in comparing negative fCycle values
            elif cycle is None and abs(last.fCycle) < abs(key.fCycle):
```

PR #717, opened 2022-09-19
fix: regularize ROOT type aliases to C fundamental type names.

jpivarski:

PR #718, opened 2022-09-19
fix: regularize ROOT type aliases to C fundamental type names (v4).

### LINKED ISSUE #707: Cannot read `TParameter<Long64_t>`
```py
>>> uproot.__version__
'4.3.4'
>>> v = uproot.open("bb-pdf-cables-cables_all-Co60.root")["NumberOfPrimariesEdep"] # <Unknown TParameter<Long64_t> at 0x7fae30e27e50>
>>> v.members
{}
```
So I cannot obviously ask for `fVal`. Bug? ([bb-pdf-cables-cables_all-Co60.root](https://github.com/scikit-hep/uproot5/files/9506174/bb-pdf-cables-cables_all-Co60.root.txt))

jpivarski:
This PR is the "v4" version of #717, which ultimately fixes #707.

### DIFF
--- a/src/uproot/model.py
+++ b/src/uproot/model.py
@@ -105,7 +105,10 @@ def reset_classes():
     reload(uproot.models.TMatrixT)
 
 
-_classname_regularize = re.compile(r"\s*(<|>|::)\s*")
+_classname_regularize = re.compile(r"\s*(<|>|,|::)\s*")
+_classname_regularize_type = re.compile(
+    r"[<,](Bool_t|Char_t|UChar_t|Short_t|UShort_t|Int_t|UInt_t|Long_t|ULong_t|Long64_t|ULong64_t|Size_t|Float_t|Double_t|LongDouble_t)[>,]"
+)
 _classname_encode_pattern = re.compile(rb"[^a-zA-Z0-9]+")
 _classname_decode_antiversion = re.compile(rb".*_([0-9a-f][0-9a-f])+_v([0-9]+)$")
 _classname_decode_version = re.compile(rb".*_v([0-9]+)$")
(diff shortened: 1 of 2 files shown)


jpivarski:
I think this passed accidentally because the new boost-histogram isn't getting picked up by the `main-v4` branch. Nevertheless, let's keep the two PRs in step with each other.

  1. #720 gets approved and merged into `main-v4`.
  2. This PR updates to `main-v4`.
  4. This PR gets approved and merged into `main-v4`.

src/uproot/model.py
```diff
@@ -105,7 +105,10 @@ def reset_classes():
     reload(uproot.models.TMatrixT)
 
 
-_classname_regularize = re.compile(r"\s*(<|>|::)\s*")
+_classname_regularize = re.compile(r"\s*(<|>|,|::)\s*")
+_classname_regularize_type = re.compile(
+    r"[<,](Bool_t|Char_t|UChar_t|Short_t|UShort_t|Int_t|UInt_t|Long_t|ULong_t|Long64_t|ULong64_t|Size_t|Float_t|Double_t|LongDouble_t)[>,]"
+)
```
agoose77, line 111: Should we generate the regex from the translation table?
```suggestion
_root_primitive_to_c_primitive = {
    "Bool_t": "bool",
    "Char_t": "char",
    "UChar_t": "unsigned char",
    "Short_t": "short",
    "UShort_t": "unsigned short",
    "Int_t": "int",
    "UInt_t": "unsigned int",
    "Long_t": "long",
    "ULong_t": "unsigned long",
    "Long64_t": "long long",
    "ULong64_t": "unsigned long long",
    "Size_t": "size_t",
    "Float_t": "float",
    "Double_t": "double",
    "LongDouble_t": "long double",
}

_classname_regularize = re.compile(r"\s*(<|>|,|::)\s*")
_classname_regularize_type = re.compile(
    r"[<,]("
    + "|".join([re.escape(p) for p in _root_primitive_to_c_primitive])
    + r")[>,]"
)
```

src/uproot/model.py
```diff
@@ -130,10 +133,51 @@ def classname_regularize(classname):
...
+            elif token == "Size_t":
+                replacement = "size_t"
+            elif token == "Float_t":
+                replacement = "float"
+            elif token == "Double_t":
+                replacement = "double"
+            elif token == "LongDouble_t":
+                replacement = "long double"
+            else:
+                raise AssertionError
```
agoose77, line 175:
```suggestion
            replacement = _root_primitive_to_c_primitive[token]
```

jpivarski:
I see the point and the value of this suggested change, but we'll have to manually make the same change to #717. And note that #717 is the forward-looking PR that goes into `main` (v5); this is a backport to `main-v4` (v4).

If you apply this to both PRs, I'll take it. Otherwise, we can leave it in both.

(By the way, I'm assuming that your refactor is readability-motivated, and I agree that it's an improvement. If it's performance-motivated, I wouldn't worry about it because it's inside of an unlikely `if` branch.)

agoose77:
Yes, it's primarily safety - we will never have the regex go out of sync with the if branch, but closely followed by readability.

jpivarski:
Thanks! #717 and #718 are in sync, so I'll merge them both.

PR #719, opened 2022-09-19
test: adjust for boost-histogram 1.3.2's _storage_type deprecation.

jpivarski:
This is needed for all PRs targeting `main`.

It should be fast-tracked into `main` so that it can be used in all those PRs.


jpivarski:
Specifically, it's for this:

https://github.com/scikit-hep/boost-histogram/blob/0c8659c202b0e4c5c2a061d26a4de6d526b01676/src/boost_histogram/_internal/hist.py#L617-L625

Okay sure, the Python 3.6 case fails because it's getting an older boost-histogram. I'm going to add a fallback to this and #720.

henryiii (approved):
Same comment as #720, this is the ideal way to handle it. I'm glad to see it was test only. :)

PR #720, opened 2022-09-19
test: adjust for boost-histogram 1.3.2's _storage_type deprecation (v4).

jpivarski:
This is needed for all PRs targeting `main-v4`.

It should be fast-tracked into `main-v4` so that it can be used in all those PRs.


henryiii (approved):
This looks like the correct way to handle this for now.

FYI, you can also list the warning as one you ignore, but better this way since you'll be ready if it gets dropped. (Full deprecation in 1.4, removal in 1.5 most likely)

PR #721, opened 2022-09-19
fix: ReadOnlyDirectory should provide the largest abs(cycle) when cycle is unspecified, not the largest cycle (v4).

jpivarski:
This is the "v4" version of #715.

Like the other in-flight PRs, it will follow this procedure:

  1. #720 gets approved and merged into `main-v4`.
  2. This PR updates to `main-v4`.
  4. This PR gets approved and merged into `main-v4`.

### DIFF
--- a/src/uproot/reading.py
+++ b/src/uproot/reading.py
@@ -2047,7 +2047,7 @@ class ReadOnlyDirectory(Mapping):
                 return key
             elif cycle is None and last is None:
                 last = key
-            elif cycle is None and last.fCycle < key.fCycle:
+            elif cycle is None and abs(last.fCycle) < abs(key.fCycle):
                 last = key
 
         if last is not None:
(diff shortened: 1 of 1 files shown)


src/uproot/reading.py
```diff
@@ -2047,7 +2047,7 @@ def key(self, where):
                 return key
             elif cycle is None and last is None:
                 last = key
-            elif cycle is None and last.fCycle < key.fCycle:
+            elif cycle is None and abs(last.fCycle) < abs(key.fCycle):
```
agoose77, line 2050:
```suggestion
            # Follow ROOT's behaviour in comparing negative fCycle values
            elif cycle is None and abs(last.fCycle) < abs(key.fCycle):
```

agoose77: Just adding a comment for future ease of understanding.

(yes, we can git-blame, but this is low-cost high-readability).

jpivarski:
I'll take that as an approval of both #721 and #715 and enable auto-merge.

agoose77:
Kicking CI

PR #734, opened 2022-09-23
feat: Use awkward pandas, instead of the existing code that explodes Pandas Dataframes

kkothari2001:


src/uproot/interpretation/library.py
```diff
@@ -840,105 +839,22 @@ def imported(self):
...
-            index = pandas.MultiIndex.from_arrays(
-                array.parents_localindex(entry_start, entry_stop),
-                names=["entry", "subentry"],
-            )
-            arrays = []
-            columns = []
-            _strided_to_pandas(
-                (), array.content.interpretation, array.content.array, arrays, columns
+            awkward_pandas = uproot.extras.awkward_pandas()
+            ak_lib = Awkward()
```
jpivarski, line 850: These objects should be treated as singletons.

```suggestion
            ak_lib = _libraries[Awkward.name]
```

src/uproot/interpretation/library.py
```diff
@@ -840,105 +839,22 @@ def imported(self):
...
-            )
-            out.leaflist = maxlen != 1
-            return out
-
-        elif isinstance(array, uproot.interpretation.jagged.JaggedArray) and isinstance(
-            array.content, uproot.interpretation.objects.StridedObjectArray
+        if (
+            (not (isinstance(array, numpy.ndarray)))
+            or (array.dtype.names is not None)
+            or (len(array.shape) != 1)
```
jpivarski, line 847: Unnecessary parentheses.

```suggestion
            not (isinstance(array, numpy.ndarray))
            or array.dtype.names is not None
            or len(array.shape) != 1
```

But also, this `if`-`then`-`else` is a little hard to read because it's defined in terms of a negative. The simple case is in the `else` block and the `if` predicate is "not the simple case."

It would be more readable like this:

```python
if (
    isinstance(array, numpy.ndarray)
    and array.dtype.names is None
    and len(array.shape) == 1
):
    return pandas.Series(array, index=index)
else:
    # more complicated thing with awkward-pandas
```

(Fixing that would override the suggestion I made above. But still, you don't need the extra parentheses.)

src/uproot/interpretation/library.py
```diff
@@ -955,130 +871,131 @@ def group(self, arrays, expression_context, how):
...
-                        df.reindex(out.index),
-                        out,
-                        how=how,
-                        left_index=True,
-                        right_index=True,
-                    )
-                return out
+            return _pandas_memory_efficient(pandas, arrays, names)
+
+            # if any(isinstance(x, tuple) for x in names):
```
jpivarski, line 876: As usual, you should delete commented-out code by the end of the PR.

src/uproot/interpretation/library.py
```diff
@@ -955,130 +871,131 @@ def group(self, arrays, expression_context, how):
 
         elif uproot._util.isstr(how) or how is None:
             arrays, names = _pandas_only_series(pandas, arrays, expression_context)
```
jpivarski, line 873: The `_pandas_only_series` function might also simplify because now you know that `arrays` contains _only_ Series, no DataFrames.

It still does a `_rename` thing that I think might be necessary. (I don't remember what it's for. That would be good to check: what puts a `"rename"` key in a `context`?)

src/uproot/interpretation/library.py
```diff
@@ -955,130 +871,131 @@ def group(self, arrays, expression_context, how):
...
-                    )
-                    out = pandas.merge(
-                        df.reindex(out.index),
-                        out,
-                        how=how,
-                        left_index=True,
-                        right_index=True,
-                    )
-                return out
+            return _pandas_memory_efficient(pandas, arrays, names)
```
jpivarski, line 874: Now `_pandas_memory_efficient` is the _only_ way DataFrames are made. Is it still necessary?

Calling `gc.collect()` is not a good thing to do in production code; I only acquiesced because a user was having real trouble with the intermediate memory use of building a whole DataFrame at once. I would _love_ to be able to justify removing `gc.collect()` from the codebase: that is, ensuring that we're combining Series that all have a common RangeIndex into a DataFrame without a lot of copying or shuffling. It seems to me like the DataFrame should just adopt the memory used by the Series. If there's some flag in Pandas to make that happen, it would be ideal. We know, for instance, that the original Series are not going to be used anymore; the new DataFrame does not need to copy the Series for safety.

jpivarski:
I see that this is not quite ready for review yet; there are still some tests to work through, and I made some suggestions in the code.

Another thing I should point out is that we no longer need a `_libraries_lazy` because we no longer have an `uproot.lazy`:

https://github.com/scikit-hep/uproot5/blob/366bbe85d29bfc8064a35fb5edd3d2da39a597f2/src/uproot/interpretation/library.py#L1121-L1160

This was for the set of libraries that can be used in `uproot.lazy`; only Awkward. There are no longer any references to `_libraries_lazy` in the codebase, so you can safely delete it.

src/uproot/interpretation/library.py
kkothari2001: https://github.com/scikit-hep/uproot5/blob/main/src/uproot/behaviors/TBranch.py#L2900 does, I guess I'll just keep that in

src/uproot/interpretation/library.py
kkothari2001: Done

src/uproot/interpretation/library.py
kkothari2001: Changed as suggested 👍🏻

jpivarski:
I attached issue #668 to this PR because this PR would fix that issue.

kkothari2001:
2 tests remain, I will solve them once awkard-pandas has a new release.
1 test (relating to HTTP sockets) was failing on my local machine, I don't think it is because of something I did. Could be that the service is temporarily down.

cc @jpivarski

jpivarski:
As a quick-fix (without looking into it), I restarted the CI jobs.

I'll return to this PR soon.

kkothari2001:
Forgot to add awkward-pandas to dependancies. Solving this right now.

jpivarski (approved):
I had been following the development of this PR, but I forgot to do the final review. (I use the "request review" mechanism in GitHub as a reminder.)

I just checked it over again, and it looks great! If the tests pass, we're ready to go. Also, this is an important part of Uproot 5.0.0, which will be released at the beginning of next week.

A few tests failed because awkward-pandas only supports Python 3.8 and above, while we have some Python 3.7 tests. That should be fixed now.

jpivarski:
The failure now is because this PR branch needs to be brought up to date with `main`, specifically #784.

PR #736, opened 2022-09-26
feat: use the v2 main release

agoose77:
This can't be merged until we have an `awkward==2.0.0rc0` out.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
-    "awkward>=1.9.0rc1",
+    "awkward>=2.0.0rc0",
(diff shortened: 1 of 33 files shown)


agoose77:
Huh, I thought I'd installed pre-commit on this repo already. Oops!

jpivarski:
I think this PR and #741 are both attempting to do the same thing. We both dropped the "`_v2`" everywhere, but I think our other commits differ.

agoose77:
I think we can close this; I started work on dropping the v2 prefix, but I wasn't sure if the plan was to make uproot5 require `awkward >= 2.0`. :)

jpivarski:
> I wasn't sure if the plan was to make uproot5 require `awkward >= 2.0`

It will require it. I probably need to make the pyproject.toml changes you made here in #741.

PR #741, opened 2022-09-29
refactor: prepare for Awkward 2.0.0rc1

jpivarski:
Starting by adapting to API changes and removing the "`_v2`" submodule.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -58,6 +58,9 @@ jobs:
           mamba install xrootd
           conda list
 
+      - name: "TEMPORARY!"
+        run: python -m pip install -U 'git+https://github.com/scikit-hep/awkward.git@main'
+
       - name: "Pip install the package"
         run: python -m pip install .[test,dev]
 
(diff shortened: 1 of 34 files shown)


jpivarski:
c054c716f048a99ab0174ff90a1939068861c6c8 is the one thing that was in #736 but not this PR. Now this one can supercede it.

agoose77:
@jpivarski oh, simple - we just need #742. I was just re-implementing that.

agoose77:
I need to make one PR to be included in this RC. I'll link it here in the next couple of minutes

jpivarski:
3d37ed74a9689e941222f6af950c627b75a6b33b installs Awkward from source to give this PR a fair test. (It's supposed to demonstrate testing against Awkward `main`.) That CI step, marked "TEMPORARY!" should be removed before this PR is done.

The 3.11 tests are failing because they can't find a suitable version of Awkward Array _before_ the latest wheel is replaced by the "TEMPORARY!" CI step. I think we can just ignore those.

PR #742, opened 2022-09-29
Drop Python 3.6.

jpivarski:


henryiii:
Here's the issue that forced me to drop macOS Python 3.11 testing:

https://github.com/actions/setup-python/issues/512

henryiii:
IMO, I believe that's a OpenSSL GitHub Actions bug, not a Python 3.11 one; if you want to make the SSL tests skippable then `pytest.mark.ssl` + add `ssl` as a mark + `-m "not ssl"` would be a better way to skip them.

PR #743, opened 2022-10-03
ci: remove GHA workaround for macOS Python 3.11

henryiii:
Added in #742, was tracking https://github.com/actions/python-versions/pull/189, and fixed in https://github.com/actions/setup-python/issues/512.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -67,7 +67,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Thanks, that was quick! When I noticed that MacOS was excluded from the `platform` matrix, I thought, "then the `pytest.mark.skip` lines were unnecessary." I see that you've removed them here, so this should be everything. If you're ready, go ahead and enable auto-merge.

PR #745, opened 2022-10-05
chore: some cleanup inpsired by refurb

henryiii:
I've added a few cleanups based on `pipx run refurb src`. I didn't work on else after return (there's a lot of them), or on ones that were not clearly an improvement, but I liked these.

- chore: remove extra returns
- chore: using copy for dicts (better readability)
- chore: nicer strings and comparisons
- chore: use contextlib.suppress

### DIFF
--- a/src/uproot/_awkward_forth.py
+++ b/src/uproot/_awkward_forth.py
@@ -53,7 +53,6 @@ class ForthGenerator:
         self.form_keys = []
         self.var_set = var_set
         self.count_obj = count_obj
-        return
 
     def traverse_aform(self):
         self.aform = self.aform.content
@@ -316,21 +315,17 @@ class ForthGenerator:
             self.forth_code[ref] = {}
         self.forth_code[ref][str(ref) + "pre"] = forth_exec_pre
         self.forth_code[ref][str(ref) + "post"] = forth_exec_post
-        return
 
     def add_to_final(self, code):
         if not isinstance(code, list):
             raise TypeError
         self.final_code.extend(code)
-        return
 
     def add_to_header(self, code):
         self.final_header.append(code)
-        return
 
     def add_to_init(self, code):
         self.final_init.append(code)
-        return
 
 
 class GenHelper:
--- a/src/uproot/exceptions.py
+++ b/src/uproot/exceptions.py
@@ -78,14 +78,7 @@ class KeyInFileError(KeyError):
             in_object = f"\nin object {self.object_path}"
 
         if self.cycle == "any":
-            return """not found: {} (with any cycle number){}{}{}{}""".format(
-                repr(self.key), because, with_keys, in_file, in_object
-            )
-        elif self.cycle is None:
-            return """not found: {}{}{}{}{}""".format(
-                repr(self.key), because, with_keys, in_file, in_object
-            )
-        else:
-            return """not found: {} with cycle {}{}{}{}{}""".format(
-                repr(self.key), self.cycle, because, with_keys, in_file, in_object
-            )
+            return f"not found: {self.key!r} (with any cycle number){because}{with_keys}{in_file}{in_object}"
+        if self.cycle is None:
+            return f"not found: {self.key!r}{because}{with_keys}{in_file}{in_object}"
+        return f"not found: {self.key!r} with cycle {self.cycle}{because}{with_keys}{in_file}{in_object}"
--- a/src/uproot/exceptions.py
+++ b/src/uproot/exceptions.py
@@ -79,6 +79,7 @@ class KeyInFileError(KeyError):
 
         if self.cycle == "any":
             return f"not found: {self.key!r} (with any cycle number){because}{with_keys}{in_file}{in_object}"
-        if self.cycle is None:
+        elif self.cycle is None:
             return f"not found: {self.key!r}{because}{with_keys}{in_file}{in_object}"
-        return f"not found: {self.key!r} with cycle {self.cycle}{because}{with_keys}{in_file}{in_object}"
+        else:
+            return f"not found: {self.key!r} with cycle {self.cycle}{because}{with_keys}{in_file}{in_object}"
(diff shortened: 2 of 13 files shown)


src/uproot/exceptions.py
```diff
@@ -78,14 +78,7 @@ def __str__(self):
...
-                repr(self.key), because, with_keys, in_file, in_object
-            )
-        else:
-            return """not found: {} with cycle {}{}{}{}{}""".format(
-                repr(self.key), self.cycle, because, with_keys, in_file, in_object
-            )
+            return f"not found: {self.key!r} (with any cycle number){because}{with_keys}{in_file}{in_object}"
+        if self.cycle is None:
+            return f"not found: {self.key!r}{because}{with_keys}{in_file}{in_object}"
+        return f"not found: {self.key!r} with cycle {self.cycle}{because}{with_keys}{in_file}{in_object}"
```
jpivarski, line 84: The formatting improvements are very good, but I want to keep match-case logic in explicit if-elif-else chains. The result is the same because all code branches return, but I think this expresses the intent more clearly.

```suggestion
        elif self.cycle is None:
            return f"not found: {self.key!r}{because}{with_keys}{in_file}{in_object}"
        else:
            return f"not found: {self.key!r} with cycle {self.cycle}{because}{with_keys}{in_file}{in_object}"
```

src/uproot/interpretation/objects.py
```diff
@@ -376,16 +377,14 @@ def simplify(self):
         ``self``.
         """
         if self._branch is not None:
-            try:
+            with contextlib.suppress(CannotBeStrided):
```
jpivarski, line 380: This one is new to me and expresses the intent very well!

src/uproot/source/xrootd.py
```diff
@@ -111,7 +111,7 @@ def _xrd_error(self, status):
 
         # https://github.com/xrootd/xrootd/blob/8e91462e76ab969720b40fc324714b84e0b4bd42/src/XrdCl/XrdClStatus.hh#L47-L103
         # https://github.com/xrootd/xrootd/blob/250eced4d3787c2ac5be2c8c922134153bbf7f08/src/XrdCl/XrdClStatus.cc#L34-L74
-        if status.code == 101 or status.code == 304 or status.code == 400:
+        if status.code in {101, 304, 400}:
```
jpivarski, line 114: Elsewhere, we have `if ___ in <tuple>`, so why should this one be a set?

```suggestion
        if status.code in (101, 304, 400):
```

You convinced me at one point that Python is smart enough to avoid the performance penalty that would come from constructing the tuple of constants just to see if some value is in it. That probably only applies to tuples of constants, not mutable types like a set (though it could, if a static code analyzer sees that this set doesn't escape the expression). Also, 3 is a small enough number of items that the hashmap-lookup might even be slower than iterating over the options.

But, of course, the main reason is for consistency. All of the other "is this a member of?" cases are tuples, not sets.

src/uproot/writing/writable.py
```diff
@@ -109,7 +109,7 @@ def recreate(file_path, **options):
         "initial_streamers_bytes", create.defaults["initial_streamers_bytes"]
     )
     uuid_function = options.pop("uuid_function", create.defaults["uuid_function"])
-    if len(options) != 0:
+    if options:
```
jpivarski, line 112: This is one of the few things that I'm self-consciously against the grain with in Python: the truthiness rules that turn any Python object into a boolean when it's in the position of a predicate seem very dangerous to me. I would personally prefer to be explicit in all of these cases.

```suggestion
    if len(options) != 0:
```

But since the Pythonic thing is to rely on truthiness, this is negotiable.

src/uproot/writing/writable.py
```diff
@@ -157,7 +157,7 @@ def update(file_path, **options):
         "initial_directory_bytes", create.defaults["initial_directory_bytes"]
     )
     uuid_function = options.pop("uuid_function", create.defaults["uuid_function"])
-    if len(options) != 0:
+    if options:
```
jpivarski, line 160: Same here.

```suggestion
    if len(options) != 0:
```

src/uproot/_awkward_forth.py
```diff
@@ -53,7 +53,6 @@ def __init__(self, aform=None, count_obj=0, var_set=False):
         self.form_keys = []
         self.var_set = var_set
         self.count_obj = count_obj
-        return
```
jpivarski, line 56: @aryan26roy and I are still refactoring this code, and unnecessary `return` statements like this would be on the chopping block, but thanks for catching it!

jpivarski (approved):
I think these are a lot of improvements. I have a few counter-suggestions, all of which are clickable as "±" suggested changes.

src/uproot/source/xrootd.py
henryiii: This actually is stored as a frozen set in the byte code since it's inline :) But there's not much difference for a small one like this. To me, "in" is natural on a set, and bolted on for lists/tuples. But there are some things that can't go into sets. I'm not sure what the performance profile is, so no problem here, I just had to pick one and happy to pick the other.

src/uproot/exceptions.py
henryiii: I am not fond of adding extra "elif"/"else" that serve no purpose other than to confuse the reader into thinking they are meaningful, adding extra indentation, and making a more complicated AST. It's also more visually apparent to me that the final fall through case (else) is special, and that all possible branches are covered. When these are nested, it's even more important. 

But I wasn't making this change, this just slipped in because I was already editing the lines.

src/uproot/interpretation/objects.py
henryiii: Python 3.4+

src/uproot/writing/writable.py
henryiii: IMO, since the type is fixed (I checked for any of these I changed), then this reads better IMO ("if there are options" is how I'd read it in English), and is potentially faster too, since it's calling fewer functions.

Once you start statically typing, this truthiness rule is much less bothersome. If you know the type, it's easy to safely use truthiness.

henryiii: If you still want these, feel free to self-accept the suggestions, it's your project :)

jpivarski (approved):
I would take this as-is (including the `if options`, since the type is very clear in that context).

I'll turn on auto-merge (squash).

Thanks!

Issue #556, opened 2022-02-08
Implement non-memberwise std::map found in a histogram


jpivarski:
In this file: http://submit08.mit.edu/~freerc/?dir=Jim

there's a histogram in

```
DQMData/Run 315270/DT/Run summary/02-Segments/Wheel2/Sector9/Station4/hResDist_W2_St4_Sec9_SL3
```

that is raising

```python
/cvmfs/sft.cern.ch/lcg/views/LCG_101swan/x86_64-centos7-gcc8-opt/lib/python3.9/site-packages/uproot/containers.py in read(self, chunk, cursor, context, file, selffile, parent, header)
   1136 
   1137         else:
-> 1138             raise NotImplementedError(
   1139                 """non-memberwise serialization of {0}
   1140 in file {1}""".format(

NotImplementedError: non-memberwise serialization of AsMap
in file DQM_V0001_R000315270__SingleMuon__Run2018A-12Nov2019_UL2018-v2__DQMIO.root
```

when you try to read it. This will be a good case to use to figure out non-memberwise std::map deserialization.

PR #746, opened 2022-10-05
feat: implemented NON-memberwise deserialization for AsMap.

jpivarski:
For all Interpretations other than AsMap, the non-memberwise case is implemented, and the memberwise case is implemented for only very few (and partially, at that). That's because when I tried making all combinations of STL containers, `std::map` consistently got serialized in a memberwise way, so I had no examples of what non-memberwise serialization looked like.

Earlier this year, however, I found an example: #556, and earlier today, a user encountered it: #744. Apparently, maps of parameter names and values are serialized as memberwise `std::map` in TFormula, which is part of a TF1 that can be attached to a TH1 histogram. So adding this implementation enables reading histograms with attached fit functions.

There's no indication yet that memberwise `std::map` appears in TTrees, so it's not yet clear that we'll need an AwkwardForth implementation. Furthermore, the two examples (#556 and #744) are in histograms, not TTrees, so we wouldn't be able to test an AwkwardForth implementation.

The memberwise `std::map` turns out to be very simple, though: just

   1. (4 bytes) length of map
   2. deserialize one key
   3. deserialize one value
   4. GOTO 2, until end of map

I also don't have any files that are small enough to add to scikit-hep-testdata, so no tests, unfortunately.

### DIFF
--- a/src/uproot/containers.py
+++ b/src/uproot/containers.py
@@ -1522,14 +1522,15 @@ class AsMap(AsContainer):
                 instance_version,
                 is_memberwise,
             ) = uproot.deserialization.numbytes_version(chunk, cursor, context)
-            cursor.skip(6)
-            if helper_obj.is_forth():
-                temp_jump = cursor._index - start_cursor._index
-                helper_obj.add_to_pre(f"{temp_jump} stream skip\n")
         else:
             is_memberwise = False
-        # raise NotImplementedError
+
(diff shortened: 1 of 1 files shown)


agoose77 (approved):

jpivarski:
Thanks! I'll take that as approval of the backport (#747) as well.

Issue #750, opened 2022-10-06
DeserializationError after hadd 


JohanWulff:
I have written 2 .root files with uproot and both of them can be accessed perfectly fine individually:
```
In [3]: egamma_b = uproot.open("./EGamma_B.root")

In [4]: egamma_b['tout']['Electron_eta'].array()
Out[4]: <Array [-0.656, -0.725, ... -1.67, 0.241] type='148 * float64'>

In [6]: egamma_c = uproot.open("./EGamma_C.root")

In [7]: egamma_c['tout']['Electron_eta'].array()
Out[7]: <Array [-0.613, 0.572, ... 0.957, -0.981] type='164 * float64'>

```

After hadding them like this:

```
hadd Data.root EGamma_B.root EGamma_C.root
```
accessing the resulting file 
```
In [8]: data = uproot.open("./Data.root")

In [9]: data['tout']['Electron_eta'].array()

```

raises a deserialisation error:
```
---------------------------------------------------------------------------
DeserializationError                      Traceback (most recent call last)
Cell In [9], line 1
----> 1 data['tout']['Electron_eta'].array()

File ~/miniconda3/envs/shep/lib/python3.10/site-packages/uproot/behaviors/TBranch.py:2210, in TBranch.array(self, interpretation, entry_start, entry_stop, decompression_executor, interpretation_executor, array_cache, library)
   2204             for (
   2205                 basket_num,
   2206                 range_or_basket,
   2207             ) in branch.entries_to_ranges_or_baskets(entry_start, entry_stop):
   2208                 ranges_or_baskets.append((branch, basket_num, range_or_basket))
-> 2210 _ranges_or_baskets_to_arrays(
   2211     self,
   2212     ranges_or_baskets,
   2213     branchid_interpretation,
   2214     entry_start,
   2215     entry_stop,
   2216     decompression_executor,
   2217     interpretation_executor,
   2218     library,
   2219     arrays,
   2220     False,
   2221 )
   2223 _fix_asgrouped(
   2224     arrays, expression_context, branchid_interpretation, library, None
   2225 )
   2227 if array_cache is not None:

File ~/miniconda3/envs/shep/lib/python3.10/site-packages/uproot/behaviors/TBranch.py:3495, in _ranges_or_baskets_to_arrays(hasbranches, ranges_or_baskets, branchid_interpretation, entry_start, entry_stop, decompression_executor, interpretation_executor, library, arrays, update_ranges_or_baskets)
   3492     pass
[...]

PR #751, opened 2022-10-06
fix: avoid empty TBasket issue in embedded TBasket

jpivarski:
The issue is that these files apparently don't have a second TKey. See https://github.com/scikit-hep/uproot5/issues/750#issuecomment-1270469267.

### DIFF
--- a/src/uproot/models/TBasket.py
+++ b/src/uproot/models/TBasket.py
@@ -272,7 +272,10 @@ class Model_TBasket(uproot.model.Model):
             cursor.skip(self._members["fKeylen"])
 
             self._raw_data = None
-            self._data = cursor.bytes(chunk, self.border, context)
+            if self.border == 0:
+                self._data = numpy.empty(0, dtype=numpy.uint8)
+            else:
+                self._data = cursor.bytes(chunk, self.border, context)
 
         else:
             compressed_bytes = self._members["fNbytes"] - self._members["fKeylen"]
--- a/tests/test_0750-avoid-empty-TBasket-issue.py
+++ b/tests/test_0750-avoid-empty-TBasket-issue.py
@@ -1,0 +1,17 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot4/blob/main/LICENSE
+
+import json
+import sys
+
+import numpy
+import pytest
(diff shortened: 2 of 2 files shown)


jpivarski:
@agoose77, if you approve either one of these (#751 or #752), I'll consider them both approved and will auto-merge them both. Thanks!

src/uproot/models/TBasket.py
```diff
@@ -272,7 +272,10 @@ def read_members(self, chunk, cursor, context, file):
             cursor.skip(self._members["fKeylen"])
 
             self._raw_data = None
-            self._data = cursor.bytes(chunk, self.border, context)
+            if self.border == 0:
+                self._data = numpy.empty(0, dtype=numpy.uint8)
```
agoose77, line 276: Is there a reason to special case this outside of `cursor.bytes`? I.e., is there a time when we'd want to call `cursor.bytes` with `self.border = 0`?

jpivarski: The signature for `cursor.bytes` is `chunk, length, context`, so this is a special case for `length=0`.

We could have `cursor.bytes` always short-circuit like this whenever the requested length is zero, but then we wouldn't find out about issues like this one. It was only after I found out about the missing second TKey in this file that I was comfortable ignoring the fact that `cursor` is outside of the `chunk`. If it happens in other circumstances, I'd want to know why because it might be an indication of a bigger misunderstanding about the ROOT format.

That's the only reason.

agoose77: OK, so this isn't fixing a broken `bytes()` for zero bytes, it's that the conditions for using `bytes()` are not met in this context?  Fab.

agoose77 (approved):

src/uproot/models/TBasket.py
jpivarski: That's right: the real issue here is the

```python
cursor.skip(self._members["fKeylen"])
```

just above it. The whole story is here: https://github.com/scikit-hep/uproot5/issues/750#issuecomment-1270469267.

agoose77: Ah, reading that again, that's exactly what you're saying! Great.

PR #752, opened 2022-10-06
fix: avoid empty TBasket issue in embedded TBasket (v4)

jpivarski:
This is the backport version of #751.

### DIFF
--- a/tests/test_0750-avoid-empty-TBasket-issue.py
+++ b/tests/test_0750-avoid-empty-TBasket-issue.py
@@ -14,4 +14,4 @@ def test():
     with uproot.open(skhep_testdata.data_path("uproot-issue-750.root"))[
         "tout/Electron_eta"
     ] as branch:
-        branch.array()
+        branch.array(library="np")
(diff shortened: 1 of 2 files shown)


tests/test_0750-avoid-empty-TBasket-issue.py
```diff
@@ -0,0 +1,17 @@
...
+
+import numpy
+import pytest
+import skhep_testdata
+
+import uproot
+
+
+def test():
+    with uproot.open(skhep_testdata.data_path("uproot-issue-750.root"))[
```
agoose77, line 14:
```suggestion
def test():
    pytest.importorskip("awkward")
    with uproot.open(skhep_testdata.data_path("uproot-issue-750.root"))[
```

agoose77: In `main-v4`, Awkward isn't installed by default, and the CI skips it for windows. Perhaps we still want to run this on windows. If so, is there a way to read but not interpret?

jpivarski: Both of these should be using `library="np"`!

PR #753, opened 2022-10-06
fix: don't use Awkward in test_0751 that doesn't need it

jpivarski:
This is a quick fix-up of #751, which shouldn't have been using Awkward in the test.

It's also interesting that it wasn't caught in `main`, only `main-v4`. In principle, there ought to be a test in our suite that doesn't install Awkward.


jpivarski:
@agoose77, since this is an add-on for #751, I'll just be squash-and-merging this.

PR #725, opened 2022-09-21
feat: (POC) Add column projection for reading to dask_awkward

douglasdavis:
This PR shows how to add column projection to `uproot.dask` when `library="ak"`. For this to work properly right now it would require the changes that are in ContinuumIO/dask-awkward#80 (still WIP). I've made this draft PR just to share the planned interface for making I/O column-project-able. Since `uproot.dask(..., library="ak")` already uses `dask_awkward.from_map(fn: Callable, ...)`, all `fn` needs is a `project_columns` method.

PR #755, opened 2022-10-11
feat: Added column_projection optimization

kkothari2001:
Include the `project_columns` method in the existing callable classes as described in #725.

Ready for review. cc @jpivarski @douglasdavis 
I wasn't sure how to test this though, any suggestions?

### DIFF
--- a/tests/test_0755-dask-awkward-column-projection.py
+++ b/tests/test_0755-dask-awkward-column-projection.py
@@ -9,7 +9,6 @@ import uproot
 dask = pytest.importorskip("dask")
 dask_awkward = pytest.importorskip("dask_awkward")
 
-pytest.importorskip("pyarrow")  # dask_awkward.lib.testutils needs pyarrow
 from dask_awkward.lib.testutils import assert_eq
 
 
--- a/tests/test_0755-dask-awkward-column-projection.py
+++ b/tests/test_0755-dask-awkward-column-projection.py
@@ -1,0 +1,25 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot4/blob/main/LICENSE
+
+import numpy
+import pytest
+import skhep_testdata
+
+import uproot
+
+dask = pytest.importorskip("dask")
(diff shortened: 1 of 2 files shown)


douglasdavis:
Since the logic is handled on the `dask-awkward` side I think a simple sanity check would suffice. Just do a read followed by a `getitem` and see that you get what is expected
```python
data = uproot.dask(..., library="ak")
data = data[["branch1", "branch2"]]
result = data.compute()
assert ... # check that result is equal to the same thing but not dask
```
This doesn't test that column projection takes place, but it may potentially catch future bugs in the workflow that column projection targets (read followed by getitem)

tests/test_0755-dask-awkward-column-projection.py
```diff
@@ -0,0 +1,25 @@
...
+import numpy
+import pytest
+import skhep_testdata
+
+import uproot
+
+dask = pytest.importorskip("dask")
+dask_awkward = pytest.importorskip("dask_awkward")
+
+pytest.importorskip("pyarrow")  # dask_awkward.lib.testutils needs pyarrow
```
jpivarski, line 12: As we talked about in the meeting, this is no longer necessary.

```suggestion
```

Now we'll see if pytest agrees!

jpivarski (approved):
This is good; the only thing to change is the `importorskip` below.

After this PR, we should add a test that verifies that the undesired columns are not being read. That can be done with the `cache` option of array-reading functions (like [TTree.arrays](https://uproot.readthedocs.io/en/latest/uproot.behaviors.TTree.TTree.html#arrays)); see also [this tutorial](https://uproot.readthedocs.io/en/latest/basic.html#caching-and-memory-management). When you provide a dict as `cache`, it will be filled with unique string keys and array references as values. The unique strings include the TBranch name, along with a lot of other stuff. To allow that key format to evolve, only require that the TBranch name is a substring of the key, not that it appears at a particular position. (E.g. `"px1" in key`.)

Issue #672, opened 2022-08-11
Support boost_histogram/hist's categorical axes


jpivarski:
Inspired by #659 (@jlabounty). Currently, the boost-histogram/hist → ROOT TH* conversion takes a categorical axis with `n` categories as a continuous regular axis of `n` bins from `0` to `n-1` (inclusive). ROOT TH* objects have no notion of categorical axes, but they do have `fLabels`, which can be set to a list of strings. When this value is set to `n` strings, ROOT draws these strings under each tick, effectively simulating a categorical axis.

This feature request would be (1) to have boost-histogram/hist → ROOT TH* set `fLabels`, somewhere here (may have to be expanded to a regular `for` loop):

https://github.com/scikit-hep/uproot5/blob/c7d728207878edaf9d017595d3aba864ad58ecf3/src/uproot/writing/identify.py#L320-L331

And (2) to interpret a TAxis with `fLabels is not None and len(fLabels) != 0` as a boost-histogram/hist categorical axis when transforming the other way:

https://github.com/scikit-hep/uproot5/blob/c7d728207878edaf9d017595d3aba864ad58ecf3/src/uproot/behaviors/TH1.py#L17-L40

We probably won't be able to preserve the int-ness of IntCategorical (`fLabels` are strictly strings), but maybe if all strings don't raise ValueError when cast as a Python `int`, that would be a good indicator.

Issue #762, opened 2022-10-24
Problem writing TList of TObjString


lobis:
I have noticed a problem when trying to write a `TList` of `TObjString` objects.

**I woud like to reproduce the following file created by root**:

```c++
const std::vector<const char*> strings = {"this", "is", "a", "test"};

auto f = TFile::Open("list.root", "RECREATE");

auto l = new TList();

for (const auto& s: strings) {
    l->Add(new TObjString(s));
}

l->Write("listOfStrings", TObject::kSingleKey); // if TObject::kSingleKey is not specified, each item on the list is written as a key to the root file

f->Close();
```

<details>
  <summary>Opening the file with root:</summary>

```
root@2a578f2dc0f6:~# root list.root
   ------------------------------------------------------------------
  | Welcome to ROOT 6.26/06                        https://root.cern |
  | (c) 1995-2021, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Sep 25 2022, 12:37:00                 |
  | From tags/v6-26-06@v6-26-06                                      |
  | With c++ (Ubuntu 11.2.0-19ubuntu1) 11.2.0                        |
  | Try '.help', '.demo', '.license', '.credits', '.quit'/'.q'       |
   ------------------------------------------------------------------

root [0]
Attaching file list.root as _file0...
(TFile *) 0x5572bae81430
root [1] .ls
TFile**         list.root
 TFile*         list.root
  KEY: TList    listOfStrings;1 Doubly linked list
root [2] listOfStrings->Print(nullptr)
Collection name='TList', class='TList', size=4
 TObjString = this
 TObjString = is
 TObjString = a
 TObjString = test
root [3]
```
</details>

My attempt to do this using `uproot` makes use of the `uproot.writing.identify.to_TList` method which according to its description should be able to write a `TList` of root objects into a root file.

Python to write `TList` into file:

```python
import uproot
print(f"{uproot.__version__=}")

strings = ["this", "is", "a", "test"]

listOfStrings = uproot.writing.identify.to_TList([uproot.writing.identify.to_TObjString(s) for s in strings])
# listOfStrings = uproot.writing.identify.to_TList([]) # this works

print(f"{listOfStrings=}")

filename = "list.root"
key = "listOfStrings"

with uproot.recreate(filename) as f:
    f[key] = listOfStrings
```

```
uproot.__version__='5.0.0rc4'
[...]

Issue #659, opened 2022-08-03
Calling to_hist() and to_numpy() gives error when opening a CategoryAxis which has been saved to a root file


jlabounty:
Hello,

I have been using uproot to save/load hist objects from root files. I have noticed that calling `to_hist()` on a `THn` object which contains a category axis yields a `ValueError: could not broadcast input array from shape (N,...) into shape (N+1,...)`. This snippet of code reproduces the issue:

```python
import hist
print(f'{hist.__version__=}')
import uproot
print(f'{uproot.__version__=}')
import numpy

cat_axis = hist.axis.IntCategory([1,2,3], label='Category')
reg_axis = hist.axis.Regular(100,0,100, label='Random')
h = hist.Hist(cat_axis)
h2 = hist.Hist(cat_axis, reg_axis)
h.fill(
    numpy.random.randint(1,4,1000)
)
h2.fill(
    numpy.random.randint(1,4,1000), numpy.random.normal(20,5,1000)
)

with uproot.recreate("bug_output.root") as f:
    f['h'] = h
    f['h2'] = h2

    print(f['h'])
    print(f['h2'])

with uproot.open("bug_output.root") as f:
    h_opened = f['h'] 
    h2_opened = f['h2']

print(h_opened)
print(h2_opened)

#this fails 
print(h_opened.to_hist())
print(h2_opened.to_hist())

#this also fails
print(h_opened.to_numpy()) #works, but the bin edges are wrong
print(h2_opened.to_numpy()) #fails
```

```bash
(gm2) jlab@SB3:~/github/g2_analysis/analysis$ python hist_uproot_category_bug.py 
hist.__version__='2.6.1'
uproot.__version__='4.3.3'
<TH1D (version 3) at 0x7fa5dc5650c0>
<TH2D (version 4) at 0x7fa5db88e1a0>
<TH1D (version 3) at 0x7fa5eb986710>
<TH2D (version 4) at 0x7fa5eb793400>
Traceback (most recent call last):
  File "/home/jlab/github/g2_analysis/analysis/hist_uproot_category_bug.py", line 39, in <module>
    print(h_opened.to_hist())
  File "/home/jlab/miniconda3/envs/gm2/lib/python3.10/site-packages/uproot/behaviors/TH1.py", line 206, in to_hist
    self.to_boost(metadata=boost_metadata, axis_metadata=boost_axis_metadata)
[...]

PR #764, opened 2022-10-24
feat: support categorical axes on boost histograms

lobis:
Closes https://github.com/scikit-hep/uproot5/issues/672.

The logic for setting the `fLabels` on the axis is pretty straightforward, however to achieve this we need to serialize `THashList` which is not currently implemented.

While working on this I found out a problem when serializing `TList` (which `THashList` inherits directly from), which is pointed out in https://github.com/scikit-hep/uproot5/issues/762 and is fixed by https://github.com/scikit-hep/uproot5/pull/763.

Another issue related to weighted histograms was also found and fixed at https://github.com/scikit-hep/uproot5/pull/774.

Summary of changes:
- Implement `THashList` serialization and `to_HashList` factory method.
- Add tests for `THashList` writing. Compare serialization (bytes) to `TList` as they should be the same except for class name.
- The name (`fName`) of all axes is now `xaxis`, `yaxis` or `zaxis` regardless of user definition, as this field should not be modified by the user since it causes problems on histogram drawing. (The text of the axis is `fTitle`). Tests have been updated accordingly.
- Add tests for histograms with categorical axes.


src/uproot/writing/identify.py
```diff
@@ -659,6 +660,31 @@ def _fXbins_maybe_regular(axis, boost_histogram):
...
 
 
+def _fLabels_maybe_categorical(axis, boost_histogram):
+    if boost_histogram is None:
+        return None
+
+    if not isinstance(axis, boost_histogram.axis.IntCategory) and not isinstance(
+        axis, boost_histogram.axis.StrCategory
+    ):
+        return None
```
agoose77, line 670:
```suggestion
    if not isinstance(
            axis, (boost_histogram.axis.IntCategory, boost_histogram.axis.StrCategory)
        ):
            return None
```

jpivarski:
I have to admit that I haven't looked into the code, but are the `fLabels` that you're adding converted into strings? ROOT's `fLabels` can only be strings. If they're `"\x01"`, `"\x02"`, `"\x03"`, then those are all invisible characters. You'll want them to be `"1"`, `"2"`, `"3"`, etc.

(That's a shot in the dark as to why they're not showing up in the plot.)

As a diagnostic method, instead of going directly for the TCanvas plot, what are the `TAxis::GetBinLabel(bin)` values? Or the equivalent in Uproot, `Model_TAxis.member("fLabels")`?

lobis:
> I have to admit that I haven't looked into the code, but are the `fLabels` that you're adding converted into strings? ROOT's `fLabels` can only be strings. If they're `"\x01"`, `"\x02"`, `"\x03"`, then those are all invisible characters. You'll want them to be `"1"`, `"2"`, `"3"`, etc.
> 
> (That's a shot in the dark as to why they're not showing up in the plot.)
> 
> As a diagnostic method, instead of going directly for the TCanvas plot, what are the `TAxis::GetBinLabel(bin)` values? Or the equivalent in Uproot, `Model_TAxis.member("fLabels")`?

I think they are being saved correctly as strings. `Model_TAxis.member("fLabels")` returns correct values etc.

The serialization of the root histogram also appears to be correct, which why I thought there must be some style problem misplacing the labels etc.

```
  "fXaxis" : {
    "_typename" : "TAxis",
    "fUniqueID" : 0,
    "fBits" : 8,
    "fName" : "",
    "fTitle" : "Category",
    "fNdivisions" : 510,
    "fAxisColor" : 1,
    "fLabelColor" : 1,
    "fLabelFont" : 42,
    "fLabelOffset" : 0.005,
    "fLabelSize" : 0.035,
    "fTickLength" : 0.03,
    "fTitleOffset" : 1,
    "fTitleSize" : 0.035,
    "fTitleColor" : 1,
    "fTitleFont" : 42,
    "fNbins" : 3,
    "fXmin" : 0,
    "fXmax" : 3,
    "fXbins" : [0, 1, 2, 3],
    "fFirst" : 0,
    "fLast" : 0,
    "fBits2" : 0,
    "fTimeDisplay" : false,
    "fTimeFormat" : "",
    "fLabels" : {
      "_typename" : "THashList",
      "name" : "THashList",
      "arr" : [{
        "_typename" : "TObjString",
        "fUniqueID" : 0,
        "fBits" : 0,
        "fString" : "1"
      }, {
        "_typename" : "TObjString",
        "fUniqueID" : 0,
        "fBits" : 0,
        "fString" : "2"
      }, {
        "_typename" : "TObjString",
        "fUniqueID" : 0,
        "fBits" : 0,
        "fString" : "3"
      }],
      "opt" : ["", "", ""]
    },
    "fModLabs" : null
  },
```

However the `h->GetXaxis()->GetBinLabel(bin)` always returns `""` so there must be a problem there. `h->GetXaxis()->GetLabels()` however appears to return the correct object, with size `3` and the correct values for entries. However when doing `h->GetXaxis()->SetBinLabel(bin, label);` an additional label is inserted into the `THashList`, instead of being replaced, so there must be something wrong.

Otherwise I have tried playing with the `THashList` object written by uproot from the root prompt and everything looks fine, so I will keep looking.

src/uproot/behaviors/TH1.py
```diff
@@ -21,7 +21,15 @@ def _boost_axis(axis, metadata):
...
-        out = boost_histogram.axis.StrCategory([str(x) for x in axis.member("fLabels")])
+        try:
+            out = boost_histogram.axis.IntCategory(
+                [int(x) for x in axis.member("fLabels")]
+            )
+        except ValueError:
+            # not all labels can be interpreted as integers
+            out = boost_histogram.axis.StrCategory(
+                [str(x) for x in axis.member("fLabels")]
+            )
```
jpivarski, line 32: Try-except logic is dangerous in that it can hide more exceptions than the one you want. We don't know whether `axis.member("fLabels")` raised the ValueError, `int(x)` raised it, or `IntCategory`. They have to be more focused:

```suggestion
        fLabels = axis.member("fLabels")
        try:
            labels = [int(x) for x in fLabels]
            category_cls = boost_histogram.axis.IntCategory
        except ValueError:
            labels = [str(x) for x in fLabels]
            category_cls = boost_histogram.axis.StrCategory
        out = category_cls(labels)
```

src/uproot/writing/identify.py
```diff
@@ -659,6 +660,34 @@ def _fXbins_maybe_regular(axis, boost_histogram):
...
+    labels = [str(label) for label in axis]
+    if isinstance(axis, boost_histogram.axis.IntCategory):
+        # Check labels are valid integers (this may be redundant)
+        for label in labels:
+            try:
+                int(label)
+            except ValueError:
+                raise ValueError(
+                    f"IntCategory labels must be valid integers. Found {label} on axis {axis}"
+                )
```
jpivarski, line 681: You're right; this might be redundant with https://github.com/scikit-hep/uproot5/pull/764/files#r1006165021

pre-commit was complaining about not specifying whether you want exception-chaining on the second ValueError. Although I said we default to exception-chaining, this is such a short-distance case that chaining would be redundant. I'd do `from None` here:

```suggestion
        # Check labels are valid integers (this may be redundant)
        for label in labels:
            try:
                int(label)
            except ValueError:
                raise ValueError(
                    f"IntCategory labels must be valid integers. Found {label!r} on axis {axis!r}"
                ) from None
```

I'm also adding `!r` to the error message so that it prints their `repr` strings, rather than their `str` strings. The `label` will be quoted, which is good for an error message. (One of us, debugging: "Are the spaces included in `label` or am I confusing that with the rest of the error message?")

lobis:
After some debugging I found the problem of not placing the labels correctly is due to the `fName` of the axis, which must be `xaxis` (in the case of x axis). When the axis is created by boost such as in this case, the name field is empty.

jpivarski:
Don't be surprised about the tests failing soon; I have to do a tricky two-package code change, explained in https://github.com/scikit-hep/uproot5/pull/770#issuecomment-1295652662.

When Awkward 2.0.0rc2 is released and #770 is merged into `main`, you'll be able to pick up `main` and get your tests working again.

(Sorry!)

jpivarski:
I think that a TAxis's `fName` _has to be_ `"xaxis"`, `"yaxis"`, or `"zaxis"`, but the `fTitle` can be anything. Is it the `fTitle` that gets displayed when plotting on ROOT's canvas?

lobis:
There seems to be something wierd with the tests:

Only Python 3.8 is failing, and it seems that some of my changes didn't go through to this specific build?

On run https://github.com/scikit-hep/uproot5/actions/runs/3356458514/jobs/5561538637#step:7:829, which is run for commit https://github.com/scikit-hep/uproot5/pull/764/commits/22cdbf79580eb2539cda649e06e7ff89b0d5f51a, the changes from commit https://github.com/scikit-hep/uproot5/pull/764/commits/70e70083f75e30eed6a2add0b830309794b42a47 should already be applied, but the `assert` insde the run log appears to point out that these changes didn't get done specifically only on python 3.8 🤔

lobis:
> I think that a TAxis's `fName` _has to be_ `"xaxis"`, `"yaxis"`, or `"zaxis"`, but the `fTitle` can be anything. Is it the `fTitle` that gets displayed when plotting on ROOT's canvas?

Yes, `fTitle` is what is displayed. On commit https://github.com/scikit-hep/uproot5/pull/764/commits/43affea01ddc2baedf225e1886d45ae7f314b82e I made axis name be one of theese three values and I also updated the tests cases that checked for this on [70e7008](https://github.com/scikit-hep/uproot5/pull/764/commits/70e70083f75e30eed6a2add0b830309794b42a47).

lobis:
I think this is mostly ready to go!

However there is a problem I could not manage to figure out. In [5370ac9](https://github.com/scikit-hep/uproot5/pull/764/commits/5370ac9e3281f6dca75c8f931dfeb978e908afb8) and d56eba3cec8b2fff22f7d661dad53905361f08d8 I introduced a dirty fix to bypass the `if` branch for weighted histogram. This needs to be properly addressed.

If this change is not introduced, the tests we failing, specifically `test_0422-hist-integration::test_issue_0659` (https://github.com/scikit-hep/uproot5/issues/659). The `view[...] =` call fails if the correct storage type is not selected? Also I cannot understand how 

```
view.value = values
view.variance = sumw2
```

are supposed to work. Any advice is welcome.

jpivarski (changes requested):
This all looks good to me, apart from the "`and False`" predicates called out below. I can't tell what was intended and if "temporary" was supposed to extend beyond this PR into a follow-up PR or not.

src/uproot/behaviors/TH1.py
```diff
@@ -292,14 +309,20 @@ def to_numpy(self, flow=False, dd=False):
...
+
         boost_histogram = uproot.extras.boost_histogram()
 
         values = self.values(flow=True)
 
         sumw2 = self.member("fSumw2", none_if_missing=True)
 
-        if sumw2 is not None and len(sumw2) == self.member("fNcells"):
+        # TODO: this is a temporary fix
+        if sumw2 is not None and len(sumw2) == self.member("fNcells") and False:
```
jpivarski, line 325: (I had missed this one.)

```suggestion
        # TODO: this is a temporary fix
        if sumw2 is not None and len(sumw2) == self.member("fNcells"):
```

jpivarski:
Okay, then I'm pretty sure you just want to drop all of the "`and False`" because the consequent of this predicate handles the "weights exist" case:

https://github.com/scikit-hep/uproot5/blob/5370ac9e3281f6dca75c8f931dfeb978e908afb8/src/uproot/behaviors/TH1.py#L324-L333

`sumw2 is None` would happen if the corresponding ROOT member were `nullptr`, but ROOT uses `len(sumw2) == 0` if the sumw2 is not defined (i.e. [TH1::Sumw2](https://root.cern.ch/doc/master/classTH1.html#aefa4ee94f053ec3d217f3223b01fa014) is never called). "`fNcells`" is the number of bins; there are several ways of getting that.

I don't know of a way to distinguish between a zero-bin histogram with sumw2 and a zero-bin histogram without sumw2, but that case might not be possible, anyway.

> However if we should not remove it we should atleast add some additional test. **Can we remove this code?**

We definitely can't remove the code, since there are histograms (somewhere) with sumw2 defined. They need to have the corresponding `sumw2` array shaped in the right way and converted to native endian (they're initially big-endian, and need to become little-endian on most platforms). Also, the corresponding boost-histograms need to have `Weight` storage for these.

Isn't it possible to make example ROOT histograms with `sumw2` by either calling the [TH1::Sumw2](https://root.cern.ch/doc/master/classTH1.html#aefa4ee94f053ec3d217f3223b01fa014) method directly or [TH1::Fill](https://root.cern.ch/doc/master/classTH1.html#a894833e678f283b6849e777c635c059d) with weights?

src/uproot/behaviors/TH3.py
```diff
@@ -134,15 +135,22 @@ def to_boost(self, metadata=boost_metadata, axis_metadata=boost_axis_metadata):
...
-        if isinstance(zaxis, boost_histogram.axis.StrCategory):
+        if isinstance(
+            zaxis, (boost_histogram.axis.IntCategory, boost_histogram.axis.StrCategory)
+        ):
             values = values[:, :, 1:]
 
         view = out.view(flow=True)
-        if sumw2 is not None and len(sumw2) == len(values):
+        # TODO: this is a temporary fix
+        if sumw2 is not None and len(sumw2) == len(values) and False:
```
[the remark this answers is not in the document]
lobis: Yes, this made me think the `to_boost` are pretty similar between each other (`TH1`, `TH2`, `TH3`) so I made them into one on the `Histogram` parent class in b0c940d.

jpivarski (approved):
It all looks good and I approve. Thank you very much for the effort you put into this!

One thing now looks odd to me: you've removed the three instances of

```python
    def to_boost(self, metadata=None, axis_metadata=None):
        return super().to_boost(metadata=metadata, axis_metadata=axis_metadata)
```

which bypass the `no_inherit` mechanism, which prevents normal inheritance because of the ROOT histogram inheritance tree. So how is TH1D getting a `to_boost` method?

I can see that it is because these tests pass:

https://github.com/lobis/uproot5/blob/c294f8f5ccd67ff1d75799074e2621736c5abaab/tests/test_0046-histograms-bh-hist.py#L2821-L2842

(I was about to add them, but they were already there.)

jpivarski:
Nevermind—I think I see it: `no_inherit` prevents Python classes from being applied to C++ objects in which the C++ class hierarchy has a superclass of that name, but it doesn't do anything to Python inheritance itself (which is good, because that would be confusing).

https://github.com/lobis/uproot5/blob/c294f8f5ccd67ff1d75799074e2621736c5abaab/src/uproot/deserialization.py#L78-L83

Tthe Python inheritance is

```mermaid
graph TB
    A((Histogram))-->B((TH1))
    A-->C((TH2))
    A-->D((TH3))
    A-->H((Profile))
    H-->E((TProfile))
    H-->F((TProfile2D))
    H-->G((TProfile3D))
```

and the C++ inheritance is



When we deserialize a `TH3F`, for instance, we see that it has a C++ superclass named `TH3`, and we have a registered Python behavior for `TH3`, so its methods and properties get mixed into the histogram Model class (`Model_TH3F_v#`). Since `TH3` is set to "`no_inherit`" from `TH1`, we don't mix our `TH1` methods and properties into that object. However, our Python `TH3` is a Python subclass of `Histogram`, so it gets all of the `Histogram` methods, including `to_boost`.

So this PR makes complete sense to me. I'll squash-and-merge it.

Thanks again!

tests/test_0352-write-THashList.py
```diff
@@ -0,0 +1,44 @@
...
+
+    tlist = uproot.writing.identify.to_TList(entries)
+    thashlist = uproot.writing.identify.to_THashList(entries)
+
+    with uproot.recreate(filename) as f1:
+        f1["tlist"] = tlist
+        f1["thashlist"] = thashlist
+
+    with uproot.open(filename) as f2:
+        assert f2["tlist"] != f2["thashlist"]
```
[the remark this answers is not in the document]
lobis: You are right, this does nothing!

lobis:
I will merge (squash!) once the tests pass.

PR #765, opened 2022-10-25
chore: update Uproot to require Awkward 2.0.0rc1.

jpivarski:
This PR won't pass tests until Awkward 2.0.0rc1 is actually released on PyPI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ If you need to develop Awkward Array as well, see its [installation for develope
 
 **Awkward Array is highly recommended.** It is not a strict dependency, but is included in the dependencies to help most users; install without dependencies (`--no-deps` with pip) if you want to avoid `awkward`. If you're using Uproot without Awkward Array, you'll have to use the `library="np"` option or globally set `uproot.default_library` to return arrays as NumPy arrays (see documentation).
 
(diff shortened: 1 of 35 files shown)


jpivarski:
Thanks for checking that. We can do that as a first test, but eventually we'll want a post-2.0.0rc1 awkward-dask release. (This is all a big chicken-and-egg problem.)

douglasdavis:
dask-awkward `2022.10a0` is now on PyPI 👍

jpivarski:
Yes, I completely forgot that I had already done it.

I just scanned over the two diffs, and they (the diff of diffs) are essentially the same—that is, I understand the differences. Not wrapping `ForthMachine64.output` in `np.asarray` is a little better (this PR), and this PR also doesn't have the hack of pulling in Awkward from `main`, unnecessary since yesterday. Also, the diff in this PR is fresher; it's on top of other improvements that have gone into Uproot in the past 26 days, so I closed the old one.

jpivarski:
This last triggering of the tests is a sanity check: the test of be23aec13d3134db67cfef3d6570275c58845cfd (which was successful) should have been enough. I just want to see it go from start to finish.

Other than that, I think it's done. If you (@agoose77) give the approval, I'll merge it and that unblocks Coffea tests.

PR #770, opened 2022-10-28
chore: remove Identifier and `"uproot"` parameter.

jpivarski:
The tests for this PR should pass after scikit-hep/awkward#1845 has been added to a release (probably 2.0.0rc2).


jpivarski:
The purpose of the `"uproot"` parameters in the Forms was for Awkward to do the ROOT deserialization, based on the information they contained. That is, this was a pre-AwkwardForth idea, so it really must go away.

jpivarski:
Note: tests against Awkward `main` still work.

jpivarski:
The tests won't pass until `awkward>=2.0.0rc2` is done: https://github.com/scikit-hep/awkward/actions/runs/3349263292

Making coordinated changes across git repos is not great because there must be these in-between times when there's no combination of packages that will work. However, that's why it has to happen _now_ while these are still in pre-release.

jpivarski:
There will be no more changes to this PR, except if something is needed to get the tests to work. The tests may need to be manually triggered after Awkward 2.0.0rc2 is done.

It's the other half of scikit-hep/awkward#1845.

PR #772, opened 2022-10-31
perf: streamline metadata handling for TBranch name lookup and uproot.dask

jpivarski:
Here's the problem (nano_dy.root has 1499 branches):

```python
>>> import uproot
>>> import time
>>> starttime = time.time(); tree = uproot.open("nano_dy.root:Events"); time.time() - starttime
0.4795074462890625
>>> starttime = time.time(); dask_tree = uproot.dask({"nano_dy.root": "Events"}); time.time() - starttime
2.7875990867614746
```

Why does `uproot.dask` take so much more time to process metadata? (Neither of the above actually read any arrays.)

The problem is that there are _O(n²)_ walks over TBranches to check their names. One of these is for simple name-lookup, which affects non-Dask as well as Dask, and the other is that the Dask implementation is using public API functions that do more regularization tests on the given list of TBranch names than is necessary (by construction, the list we provide is in the proper internal format).


jpivarski:
The first commit (above) makes lookup faster for Dask and non-Dask alike. The bottom line is somewhat improved:

```python
>>> starttime = time.time(); tree = uproot.open("nano_dy.root:Events"); time.time() - starttime
0.4882633686065674
>>> starttime = time.time(); dask_tree = uproot.dask({"nano_dy.root": "Events"}); time.time() - starttime
1.1381452083587646
```

Subbranch-lookup already had a cache, `HasBranches._lookup` (dict), but this was populated on demand and could include full paths with slashes. Thus, the first time we look up some end-of-list key `"z-z-top"` (diboson top production), it has to walk through the whole list to find it. Looking for all _n_ keys in order would be an _O(n²)_ lookup.

Instead, I populated the `HasBranches._lookup` as soon as the `HasBranches` is constructed (in `HasBranches.postprocess` for all of the subclasses of `HasBranches`: `TTree` and `TBranch`). This is _slower_ if the user only ever wanted one branch, and it was at the beginning of the list (`"a-a-axion"`) because a full loop over all branches must now be done once. But _only ever_ once; never more than once now.

Consequently, it is important that `HasBranches._lookup` is considered immutable after construction. We should not be putting new `where` strings into it anymore.

jpivarski:
This last edit avoids the high-level function `HasBranches.num_entries_for`. Now `uproot.dask` is not calling `_regularize_expressions` anywhere, though the final time is not noticeably better than above.

Now I'm going to ask @kkothari2001 to review this PR. I've only replaced `HasBranches.arrays` and `HasBranches.num_entries_for` in the one spot that the performance test checks; there may be others elsewhere that should also be replaced. If it's a lot of places, then these solutions shouldn't be copied verbatim but wrapped up in helper functions. I've already done that with the One Weird Trick because it's so weird that it should probably be replaced by a public function from Awkward.

When you're happy with it, @kkothari2001, you could ask for another review from @agoose77. (I shouldn't review PRs that I wrote part of.)

src/uproot/_dask.py
```diff
@@ -647,11 +649,14 @@ def real_filter_branch(branch):
...
-        else:
-            entry_step = ttree.num_entries_for(step_size, expressions=common_keys)
+
+        branchid_interpretation = {}
+        for key in common_keys:
+            branch = ttree[key]
+            branchid_interpretation[branch.cache_key] = branch.interpretation
+        entry_step = _regularize_step_size(
+            ttree, step_size, entry_start, entry_stop, branchid_interpretation
+        )
```
kkothari2001, line 659: Quick question, @jpivarski

The above code is called for each ttree, and I've noticed that different trees with the same branch name and type often have different step-sizes. Why does step-size depend on the ttree even when the branch name and data type are the same?

Just curious.

kkothari2001: Oh right, understood. Thanks

kkothari2001:
I know the above commit will not pass 2 tests due to change in our chunking policy (as we discussed in the meet). I only pushed to ask a question.

src/uproot/_dask.py
```diff
@@ -431,6 +432,21 @@ def real_filter_branch(branch):
...
+
+        branchid_interpretation = {}
+        for key in common_keys:
+            branch = ttree[key]
+            branchid_interpretation[branch.cache_key] = branch.interpretation
+        entry_step = _regularize_step_size(
+            ttree, step_size, entry_start, entry_stop, branchid_interpretation
+        )
+        start_stop_step_map[i] = (entry_start, entry_stop, entry_step)
```
kkothari2001, line 448: If we have a `hasbranches` list and we want to map a tree to certain properties (like an entry start,stop and step here) what would be the most appropriate way to hash the ttree for the map? I have used the order in which `enumerate(ttree)` loops over them, but I'm not sure if that's the best practise. Please suggest.

kkothari2001: cc @jpivarski @agoose77

jpivarski: The `entry_start`, `entry_stop` data does not need to come from

```python
        entry_start, entry_stop = _regularize_entries_start_stop(
            ttree.num_entries, None, None
        )
```

Since we're doing whole TTrees, `entry_start = 0` and `entry_stop = ttree.num_entries`. A TTree's `num_entries` is already a lookup; creating a new lookup table for it won't optimize anything. (Though removing the `_regularize_entries_start_stop` could save a bit of time.)

It's also not necessary to have fine-tuned `step_size` per file. They differ by file because each file has slightly different numbers of items per entry, but a set of ROOT files almost always represents an [independent and identically distributed](https://en.wikipedia.org/wiki/Independent_and_identically_distributed_random_variables) dataset, so the `step_size` of each file will be roughly equal. Also, chunk sizes don't need to be exact, since their purpose is to protect from memory issues on one side and overhead on the other, which is an order-of-magnitude thing.

So you can average the `step_size` values of all the files, round it to the nearest integer and ensure that it's at least 1. (Probably not necessary because the `step_size` of each file is already guaranteed to be at least 1, so their average will be as well.) The `step_size` for a single file is already an average over TBaskets.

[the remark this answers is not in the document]
jpivarski: Also keep in mind that `entry_start` is always `0`, `entry_stop` is a number attached to each `ttree`: `ttree.num_entries`. The only potentially variable thing is `step_size`. If it was important to let that vary across files, it should also be important to let it vary across TBaskets, which we don't. (Just arguing from consistency.)

jpivarski:
I ran the profile test again, and it's still fine:

```python
>>> starttime = time.time(); tree = uproot.open("nano_dy.root:Events"); time.time() - starttime
0.4969170093536377
>>> starttime = time.time(); dask_tree = uproot.dask({"nano_dy.root": "Events"}); time.time() - starttime
1.1085560321807861
```

Twice (second time in the same process as the first, so it excludes any library-loading):

```python
>>> starttime = time.time(); tree = uproot.open("nano_dy.root:Events"); time.time() - starttime
0.5592877864837646
>>> starttime = time.time(); dask_tree = uproot.dask({"nano_dy.root": "Events"}); time.time() - starttime
1.124070405960083
```

Issue #722, opened 2022-09-20
Cannot convert `TH1` to hist via `.to_boost()`


francesco-curcio:
I'm trying to read the histogram "hist" in this file [file.root.zip](https://github.com/scikit-hep/uproot5/files/9604957/file.root.zip) and I can indeed use `uproot.open`, but when trying to convert it to a boost histogram or a hist with `to_boost()` or `to_hist()` I get
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In [312], line 2
      1 f = uproot.open("file.root")
----> 2 f['hist'].to_hist().plot()

File ~/opt/anaconda3/envs/ml/lib/python3.10/site-packages/uproot/behaviors/TH1.py:206, in Histogram.to_hist(self, metadata, axis_metadata)
    195 def to_hist(self, metadata=boost_metadata, axis_metadata=boost_axis_metadata):
    196     """
    197     Args:
    198         metadata (dict of str \u2192 str): Metadata to collect (keys) and
   (...)
    203     Converts the histogram into a ``hist`` object.
    204     """
    205     return uproot.extras.hist().Hist(
--> 206         self.to_boost(metadata=boost_metadata, axis_metadata=boost_axis_metadata)
    207     )

File ~/opt/anaconda3/envs/ml/lib/python3.10/site-packages/uproot/behaviors/TH1.py:325, in TH1.to_boost(self, metadata, axis_metadata)
    323     view.variance = sumw2
    324 else:
--> 325     view[...] = values
    327 return out

File ~/opt/anaconda3/envs/ml/lib/python3.10/site-packages/boost_histogram/_internal/view.py:51, in View.__setitem__(self, ind, value)
     49     super().__setitem__(ind, array)  # type: ignore[no-untyped-call]
     50 else:
---> 51     raise ValueError("Needs matching ndarray or n+1 dim array")

ValueError: Needs matching ndarray or n+1 dim array
```
the `to_numpy()` method however works just fine.

Best,
Francesco

agoose77:
@jpivarski do you know why we slice the values for string categories here:

https://github.com/scikit-hep/uproot5/blame/092bc679b215d565652c23b04a782dbf6ba38e3a/src/uproot/behaviors/TH1.py#L317-L318

I assume this is the reason for this error; the subsequent branch for weight / non-weight handling erroneously chooses the non-weight path:

[...]

PR #774, opened 2022-10-31
fix: histogram weights not handled correctly in hist / boost conversion

lobis:
Currently the `fSumw2` (weights) are not being correctly read/written when performing conversion to/from `hist`/`boost` histograms.

This is causing issues in https://github.com/scikit-hep/uproot5/issues/672 and solving this will also close https://github.com/scikit-hep/uproot5/issues/722.

- We now use `hist` `storage_type` property to check wether it's a weighted histogram or not, instead of the `variances` which is always defined and was cuasing false positives.
- Add two new tests, to check the conversion with a histogram with weights and another one without weights.
- Updated slicing of values / variances for categorical axes.
- Refactored `weighted` property and `to_boost` method to use inheritance.
- Write test for 2D, 3D histograms with weights / labels.

### DIFF
--- a/src/uproot/behaviors/TH1.py
+++ b/src/uproot/behaviors/TH1.py
@@ -110,7 +110,11 @@ class Histogram:
         True if the histogram has weights (``fSumw2``); False otherwise.
         """
         sumw2 = self.member("fSumw2")
-        return sumw2 is not None and len(sumw2) == self.member("fNcells")
+        return (
+            sumw2 is not None
+            and len(sumw2) > 0
+            and len(sumw2) == self.member("fNcells")
+        )
 
     @property
     def kind(self):
--- a/src/uproot/behaviors/TH1.py
+++ b/src/uproot/behaviors/TH1.py
@@ -293,8 +293,8 @@ class TH1(Histogram):
 
     @property
     def weighted(self):
-        sumw2 = self.member("fSumw2", none_if_missing=True)
-        return sumw2 is not None and len(sumw2) == self.member("fNcells")
+        sumw2 = self.member("fSumw2")
+        return len(sumw2) > 0 and len(sumw2) == self.member("fNcells")
 
     @property
     def kind(self):
--- a/src/uproot/behaviors/TH1.py
+++ b/src/uproot/behaviors/TH1.py
@@ -201,7 +201,7 @@ class Histogram:
         values = self.values(flow=True)
 
         sumw2 = None
-        if self.weighted:   # ensures self.member("fSumw2") exists
+        if self.weighted:  # ensures self.member("fSumw2") exists
             sumw2 = self.member("fSumw2")
             sumw2 = numpy.asarray(sumw2, dtype=sumw2.dtype.newbyteorder("="))
             sumw2 = numpy.reshape(sumw2, values.shape)
@@ -231,7 +231,7 @@ class Histogram:
             ):
                 continue
             # TODO: simplify this code (save multidim slice into a variable?)
-            ix = (numpy.s_[:],)*i + (numpy.s_[1:],)
+            ix = (numpy.s_[:],) * i + (numpy.s_[1:],)
             values = values[ix]
             if sumw2 is not None:
                 sumw2 = sumw2[ix]
--- a/src/uproot/behaviors/TH1.py
+++ b/src/uproot/behaviors/TH1.py
@@ -110,7 +110,7 @@ class Histogram:
         True if the histogram has weights (``fSumw2``); False otherwise.
         """
         sumw2 = self.member("fSumw2")
-        return len(sumw2) > 0 and len(sumw2) == self.member("fNcells")
+        return sumw2 is not None and len(sumw2) == self.member("fNcells")
 
     @property
     def kind(self):
@@ -200,9 +200,8 @@ class Histogram:
 
         values = self.values(flow=True)
 
-        # 'fSumw2' will never be missing, if weights are not defined it is an array of length 0
         sumw2 = None
-        if self.weighted:
+        if self.weighted:   # ensures self.member("fSumw2") exists
             sumw2 = self.member("fSumw2")
             sumw2 = numpy.asarray(sumw2, dtype=sumw2.dtype.newbyteorder("="))
             sumw2 = numpy.reshape(sumw2, values.shape)
@@ -232,18 +231,10 @@ class Histogram:
             ):
                 continue
             # TODO: simplify this code (save multidim slice into a variable?)
-            if i == 0:
-                values = values[1:]
-                if sumw2 is not None:
-                    sumw2 = sumw2[1:]
-            elif i == 1:
-                values = values[:, 1:]
-                if sumw2 is not None:
-                    sumw2 = sumw2[:, 1:]
-            elif i == 2:
-                values = values[:, :, 1:]
-                if sumw2 is not None:
-                    sumw2 = sumw2[:, :, 1:]
+            ix = (numpy.s_[:],)*i + (numpy.s_[1:],)
+            values = values[ix]
+            if sumw2 is not None:
+                sumw2 = sumw2[ix]
 
         view = out.view(flow=True)
         if sumw2 is not None:
--- a/src/uproot/behaviors/TH1.py
+++ b/src/uproot/behaviors/TH1.py
@@ -109,7 +109,8 @@ class Histogram:
         """
         True if the histogram has weights (``fSumw2``); False otherwise.
         """
-        raise NotImplementedError(repr(self))
+        sumw2 = self.member("fSumw2")
+        return len(sumw2) > 0 and len(sumw2) == self.member("fNcells")
 
     @property
     def kind(self):
@@ -293,8 +294,7 @@ class TH1(Histogram):
 
     @property
     def weighted(self):
-        sumw2 = self.member("fSumw2")
-        return len(sumw2) > 0 and len(sumw2) == self.member("fNcells")
+        return super().weighted()
 
     @property
     def kind(self):
--- a/src/uproot/behaviors/TH1.py
+++ b/src/uproot/behaviors/TH1.py
@@ -201,13 +201,13 @@ class Histogram:
         values = self.values(flow=True)
 
         # 'fSumw2' will never be missing, if weights are not defined it is an array of length 0
-        sumw2 = self.member("fSumw2")
-        if len(sumw2) > 0 and len(sumw2) == self.member("fNcells"):
+        sumw2 = None
+        if self.weighted:
+            sumw2 = self.member("fSumw2")
             sumw2 = numpy.asarray(sumw2, dtype=sumw2.dtype.newbyteorder("="))
             sumw2 = numpy.reshape(sumw2, values.shape)
             storage = boost_histogram.storage.Weight()
         else:
-            sumw2 = None
             if issubclass(values.dtype.type, numpy.integer):
                 storage = boost_histogram.storage.Int64()
             else:
--- a/src/uproot/behaviors/TH1.py
+++ b/src/uproot/behaviors/TH1.py
@@ -294,7 +294,7 @@ class TH1(Histogram):
 
     @property
     def weighted(self):
-        return super().weighted()
+        return super().weighted
 
     @property
     def kind(self):
--- a/src/uproot/behaviors/TH1.py
+++ b/src/uproot/behaviors/TH1.py
@@ -287,10 +287,6 @@ class TH1(Histogram):
             raise ValueError("axis must be 0 (-1) or 'x' for a TH1")
 
     @property
-    def weighted(self):
-        return super().weighted
-
-    @property
     def kind(self):
         return "COUNT"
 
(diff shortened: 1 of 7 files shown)


tests/test_0167-use-the-common-histogram-interface.py
```diff
@@ -143,3 +148,11 @@ def test_boost_2():
         # assert f["cutflow"].to_boost().title == "dijethad"
         # assert f["cutflow"].to_boost().axes[0].name == "xaxis"
         # assert f["cutflow"].to_boost().axes[0].title == ""
+
+
+@pytest.mark.skip(reason="test file not yet available")
```
lobis, line 153:
```suggestion
```

This must be removed when the testing file is available.

src/uproot/behaviors/TH1.py
```diff
@@ -109,7 +109,8 @@ def weighted(self):
         """
         True if the histogram has weights (``fSumw2``); False otherwise.
         """
-        raise NotImplementedError(repr(self))
+        sumw2 = self.member("fSumw2")
+        return len(sumw2) > 0 and len(sumw2) == self.member("fNcells")
```
jpivarski, line 113: Elsewhere, I have been assuming that `fSumw2` can be `nullptr`, probably because it's possible for some ROOT files (some versions of ROOT?).

```suggestion
        return sumw2 is not None and len(sumw2) == self.member("fNcells")
```

src/uproot/behaviors/TH1.py
```diff
@@ -190,9 +191,73 @@ def to_boost(self, metadata=boost_metadata, axis_metadata=boost_axis_metadata):
...
+        if metadata is None:
+            metadata = boost_metadata
+
+        boost_histogram = uproot.extras.boost_histogram()
+
+        values = self.values(flow=True)
+
+        # 'fSumw2' will never be missing, if weights are not defined it is an array of length 0
+        sumw2 = None
+        if self.weighted:
```
jpivarski, line 205: The `self.weighted` property actually checks for `nullptr`, as well as `fSumw2` just not being defined in the TStreamerInfo (which I don't think has been the case in any 21st century version of ROOT).

https://github.com/scikit-hep/uproot5/blob/10d87e242c13a16131875563863369ae4b253433/src/uproot/behaviors/TH1.py#L230-L233

```suggestion
        sumw2 = None
        if self.weighted:   # ensures self.member("fSumw2") exists
```

tests/test_0167-use-the-common-histogram-interface.py
jpivarski, line 153: [scikit-hep-testdata 0.4.23](https://pypi.org/project/scikit-hep-testdata/0.4.23/) has been released, which contains uproot-issue-722.root.

```suggestion
```

src/uproot/behaviors/TH1.py
jpivarski: I just noticed that you've reimplemented the `weighted` property, but I set the implementation back to the old definition.

It would be possible to prove that checking for `fSumw2 == nullptr` is unnecessary by testing files made by a suite of old ROOT versions. That's something that I did at the beginning of Uproot's development: I selected popular ROOT versions, which had been released approximately once every 6 months, going back 5 years before the start of Uproot development, which was 2017. Then I installed each on a virtual machine and generated data in each: TTrees and (I _think_!) histograms, then checked what each of these generated. I don't remember very well the exact outcome of that study or where to find it, and if I've misremembered, I might be following a superstition to still be checking for `fSumw2 == nullptr`.

The study could be repeated for a set of ROOT versions we care about today, but it hardly seems worth it; we can just be a little overcautious, especially when it's wrapped up in a function call.

src/uproot/behaviors/TH1.py
```diff
@@ -190,9 +191,73 @@ def to_boost(self, metadata=boost_metadata, axis_metadata=boost_axis_metadata):
...
+                if sumw2 is not None:
+                    sumw2 = sumw2[1:]
+            elif i == 1:
+                values = values[:, 1:]
+                if sumw2 is not None:
+                    sumw2 = sumw2[:, 1:]
+            elif i == 2:
+                values = values[:, :, 1:]
+                if sumw2 is not None:
+                    sumw2 = sumw2[:, :, 1:]
```
jpivarski, line 246: ~~Sometimes, repetition is better than clever ways of avoiding repetition. The three explicit cases below could be replaced by a~~

```suggestion
            slicer = (Ellipsis,) * i + (slice(1, None),)
            values = values[slicer]
            if sumw2 is not None:
                sumw2 = sumw2[slicer]
```

Okay, that is actually simpler. Nevermind.

src/uproot/behaviors/TH1.py
```diff
@@ -229,8 +294,7 @@ def axis(self, axis=0):  # default axis for one-dimensional is intentional
 
     @property
     def weighted(self):
-        sumw2 = self.member("fSumw2", none_if_missing=True)
-        return sumw2 is not None and len(sumw2) == self.member("fNcells")
+        return super().weighted
```
jpivarski, line 297: Since the subclass's implementation is exactly the same as the superclass's, the whole definition can be skipped. Some linters would probably require that.

I can't delete this in a suggestion through the web form because it includes deleted lines, but I recommend it. I'll use the tag "unnecessary subclass method implementation" for all cases like this one.

src/uproot/behaviors/TH1.py
```diff
@@ -292,36 +356,5 @@ def to_numpy(self, flow=False, dd=False):
...
-        view = out.view(flow=True)
-        if sumw2 is not None and len(sumw2) == len(values):
-            view.value = values
-            view.variance = sumw2
-        else:
-            view[...] = values
-
-        return out
+    def to_boost(self, metadata=None, axis_metadata=None):
+        return super().to_boost(metadata=metadata, axis_metadata=axis_metadata)
```
jpivarski, line 360: unnecessary subclass method implementation

```suggestion
```

src/uproot/behaviors/TH2.py
```diff
@@ -34,8 +33,7 @@ def axis(self, axis):
 
     @property
     def weighted(self):
-        sumw2 = self.member("fSumw2", none_if_missing=True)
-        return sumw2 is not None and len(sumw2) == self.member("fNcells")
+        return super().weighted
```
jpivarski, line 36: unnecessary subclass method implementation

src/uproot/behaviors/TH2.py
```diff
@@ -101,39 +99,5 @@ def to_numpy(self, flow=False, dd=False):
...
-        view = out.view(flow=True)
-        if sumw2 is not None and len(sumw2) == len(values):
-            view.value = values
-            view.variance = sumw2
-        else:
-            view[...] = values
-
-        return out
+    def to_boost(self, metadata=None, axis_metadata=None):
+        return super().to_boost(metadata=metadata, axis_metadata=axis_metadata)
```
jpivarski, line 103: unnecessary subclass method implementation

```suggestion
```

src/uproot/behaviors/TH3.py
```diff
@@ -38,8 +37,7 @@ def axis(self, axis):
 
     @property
     def weighted(self):
-        sumw2 = self.member("fSumw2", none_if_missing=True)
-        return sumw2 is not None and len(sumw2) == self.member("fNcells")
+        return super().weighted
```
jpivarski, line 40: unnecessary subclass method implementation

src/uproot/behaviors/TH3.py
```diff
@@ -110,42 +108,5 @@ def to_numpy(self, flow=False, dd=False):
...
-        view = out.view(flow=True)
-        if sumw2 is not None and len(sumw2) == len(values):
-            view.value = values
-            view.variance = sumw2
-        else:
-            view[...] = values
-
-        return out
+    def to_boost(self, metadata=None, axis_metadata=None):
+        return super().to_boost(metadata=metadata, axis_metadata=axis_metadata)
```
jpivarski, line 112: unnecessary subclass method implementation

```suggestion
```

jpivarski (changes requested):
Very nice! All of my requested changes are inline.

src/uproot/behaviors/TH1.py
lobis: Unless I am missing something, there is some error in this code.

```python
import numpy as np

x = np.array(
    [[[1, 2, 3], [4, 5, 6]], [[7, 8, 9], [10, 11, 12]], [[-1, -2, -3], [-4, -5, -6]], [[-7, -8, -9], [-10, -11, -12]]])
print(f"{x.shape=}")

try:
    assert np.array_equal(x[1:], x[(Ellipsis,) * 0 + (slice(1, None),)])
    assert np.array_equal(x[:, 1:], x[(Ellipsis,) * 1 + (slice(1, None),)])
    assert np.array_equal(x[:, :, 1:], x[(Ellipsis,) * 2 + (slice(1, None),)])
except:
    print("Not working")

assert np.array_equal(x[1:], x[(numpy.s_[:],) * 0 + (numpy.s_[1:],)])
assert np.array_equal(x[:, 1:], x[(numpy.s_[:],) * 1 + (numpy.s_[1:],)])
assert np.array_equal(x[:, :, 1:], x[(numpy.s_[:],) * 2 + (numpy.s_[1:],)])
```

There is no need to correct it, I will use @agoose77 solution from https://github.com/scikit-hep/uproot5/pull/774#discussion_r1010787702, but it would be nice to see a way to achieve the same without numpy!

src/uproot/behaviors/TH1.py
lobis: I would also add a check for `len(sumw2) > 0` to be considered weighted. Probably it makes no difference since having `fNcells` being zero would mean an empty histogram but it should not be considered weighted imho.

src/uproot/behaviors/TH1.py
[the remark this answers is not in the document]
lobis: really useful, thanks!

jpivarski: The reason is that @agoose77 was correct and I was wrong! (And I didn't change my original comment in time for you to get it, apparently.)

```python
try:
    assert np.array_equal(x[1:], x[(Ellipsis,) * 0 + (slice(1, None),)])
    assert np.array_equal(x[:, 1:], x[(Ellipsis,) * 1 + (slice(1, None),)])
    assert np.array_equal(x[:, :, 1:], x[(Ellipsis,) * 2 + (slice(1, None),)])
except:
    print("Not working")
```

should fail because it's `slice(None)` (equivalent to `np.s_[:]`) that you want, not `Ellipsis`.

src/uproot/behaviors/TH1.py
lobis: > Since the subclass's implementation is exactly the same as the superclass's, the whole definition can be skipped. Some linters would probably require that.
> 
> I can't delete this in a suggestion through the web form because it includes deleted lines, but I recommend it. I'll use the tag "unnecessary subclass method implementation" for all cases like this one.

This was my first approach but it failed, I think due to `no_inherit = (uproot.behaviors.TH1.TH1,)`. Probably because I was at first implementing the methods in `TF1`, I assumed it wouldn't also work if I implemented them in `Histogram` but I was wrong, thanks!

jpivarski: Oh, right! I forgot about that. The issue is that ROOT's histogram inheritance tree is... surprising... and we used to have a problem with inherited methods being the wrong implementation. (So, for instance, if we hadn't thought about some subclass like TProfile yet, all of these methods would appear to be implemented, but they would do the wrong things.) The `no_inherit` mechanism is to prevent inheritance, which we opt into intentionally. (So an unimplemented subclass appears to be unimplemented.)

Calling `super().function(all, of, the, arguments)` is probably the best way to do it. Go ahead and resolve all of those issues.

src/uproot/behaviors/TH1.py
```diff
@@ -109,7 +109,12 @@ def weighted(self):
         """
         True if the histogram has weights (``fSumw2``); False otherwise.
         """
-        raise NotImplementedError(repr(self))
+        sumw2 = self.member("fSumw2")
+        return (
+            sumw2 is not None
+            and len(sumw2) > 0
```
lobis, line 115: This is the only addition after the initial review, I added it to make sure empty histograms are treated as non weighted. Probably it makes no difference. Let me know if you want this changed.

lobis:
Looking at the code of `TH{1,2,3}.py` I see it can be refactored in a similar way to what is done in `to_boost` in this PR (implement `to_numpy`, etc in `Histogram`). This would leave only `axes` property and perhaps the `axis` method on `TH{1,2,3}.py`.

I don't think the complexity of the code increases enough to discourage this and it would reduce code duplication (triplication!), but I would love to hear some thoughts @jpivarski @agoose77.

src/uproot/behaviors/TH1.py
jpivarski: I don't even know if zero-bin histograms are allowed in ROOT. If there is no bin data, I suppose it can't matter whether it's weighted or not, though I guess there's the underflow and overflow bins.

jpivarski (approved):

PR #780, opened 2022-11-11
fix: ensure AwkwardForth fallback path is tested without history.

jpivarski:
Swapped for loops for pytest fixtures, so that the with-AwkwardForth code path and the without-AwkwardForth code path are fully tested.

Switching the order with-AwkwardForth and without-AwkwardForth in `test_00` revealed a bug in the fallback path that hadn't been tested (also fixed in this PR).

This PR also removes "FIXME" messages for the old path: it won't be removed because it's a fallback.

I'll ask @aryan26roy to check this over.

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -583,9 +583,6 @@ def awkward_form(model, file, context):
         return model.awkward_form(file, context)
 
 
-# FIXME: Until we get Awkward reading these bytes directly, rather than
-# going through ak.from_iter, the integer dtypes will be int64 and the
-# floating dtypes will be float64 because that's what ak.from_iter makes.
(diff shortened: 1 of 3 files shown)


aryan26roy (approved):
Nothing seems to be changed for Forth generation except for the tests. I can't see why this would be a problem.

jpivarski:
It wasn't the Forth generation that was a problem; it was the non-Forth path. By doing Forth first and then non-Forth with the same objects, something was set such that it didn't actually run the non-Forth path.

The thing that was broken in the non-Forth path was adjusting for `"__array__": "sorted_map"` on the RecordForm instead of the ListOffsetForm that contains the RecordForm. (Remember how we changed that?) It was correct in Forth but not in non-Forth.

But thanks for the check!

PR #781, opened 2022-11-11
refactor: remove now-unused lazy libraries dict (no uproot.lazy).

jpivarski:
The `_libraries_lazy` dict was the set of libraries (and their names) that could be used with `uproot.lazy`, which is now gone.

The `_libraries` dict is now the only such dict.

After #734 is done, I'll look around to see if there's any more dead code that can be removed.


kkothari2001:
I do believe this section is already being removed in the current awkward pandas PR.


https://github.com/scikit-hep/uproot5/commit/85c8be583deb7c76ef835ca51ba9346490b44050

jpivarski:
Great! In that case, let's not do it twice, so that there are no merge issues.

PR #782, opened 2022-11-11
docs: fix TRef.py doc urls

veprbl:


src/uproot/models/TRef.py
```diff
@@ -15,8 +15,8 @@
 
 _tref_format1 = struct.Struct(">xxIxxxxxx")
 
-# Note: https://github.com/root-project/root/blob/master/io/doc/TFile/TRef.txt
-#       https://github.com/root-project/root/blob/master/io/doc/TFile/TRefArray.txt
+# Note: https://github.com/root-project/root/blob/master/io/doc/TFile/tref.md
+#       https://github.com/root-project/root/blob/master/io/doc/TFile/trefarray.md
```
jpivarski, line 19: That's not a bad idea, although these are just internal code comments. If they're out of date, a developer can search around and find the right ones.

But anyway, here's what _these particular_ comments would become.

```suggestion
# Note: https://github.com/root-project/root/blob/v6-26-08/io/doc/TFile/tref.md
#       https://github.com/root-project/root/blob/v6-26-08/io/doc/TFile/trefarray.md
```

(A recent tag name is more readable than a commit hash. I don't think ROOT will be deleting tags, since they correspond to released versions of ROOT.)

jpivarski (approved):

PR #787, opened 2022-11-28
feat: made 'very optional' arguments keyword-only

jpivarski:
This is similar to https://github.com/scikit-hep/awkward/pull/1905, but for Uproot.

Multi-array-fetching functions (`HasBranches.arrays`, `HasBranches.iterate`, `uproot.iterate`, `uproot.concatenate`, `uproot.dask`) can take `expressions` and `cut` as positional, but `filter_name` onward are all keyword-only. Similarly, `common_entry_offsets` takes `filter_name` onward as keyword-only (no positional arguments).

The single-array fetching function (`TBranch.array`) can take `interpretation`, `entry_start`, and `entry_stop` as positional, but `decompression_executor` onward are all keyword-only.

`uproot.open` and `ReadOnlyFile` can take `file_path` as positional, but the rest are keyword-only.

`ReadOnlyDirectory` and `HasBranches` (`TTree` and `TBranch`) have mapping-like functions (`keys`, `values`, `items`, `iterkeys`, `itervalues`, `iteritems`, and the special `classnames`, `iterclassnames`, `typenames`, `itertypenames`), which take a bunch of boolean flag arguments, all now keyword-only.

`WritableDirectory.mkdir` takes a positional `name`, but keyword-only `initial_directory_bytes`.

`WritableDirectory.mktree` takes positional `name`, `branch_types`, and `title`, but all of the advanced options are now keyword-only.

`HasBranches.show` and `WritableTTree.show` now have no positional arguments; all arguments from `filter_name` onward are keyword-only.

### DIFF
--- a/src/uproot/__init__.py
+++ b/src/uproot/__init__.py
@@ -187,3 +187,4 @@ from uproot._util import no_filter
 from uproot._dask import dask
 
 from uproot.pyroot import from_pyroot
+from uproot.pyroot import to_pyroot
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -9,6 +9,7 @@ from uproot.behaviors.TBranch import HasBranches, TBranch, _regularize_step_size
 
 def dask(
     files,
+    *,
     filter_name=no_filter,
     filter_typename=no_filter,
     filter_branch=no_filter,
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -571,6 +571,7 @@ class HasBranches(Mapping):
 
     def show(
         self,
+        *,
         filter_name=no_filter,
         filter_typename=no_filter,
         filter_branch=no_filter,
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -48,6 +48,7 @@ def iterate(
     files,
     expressions=None,
     cut=None,
+    *,
     filter_name=no_filter,
     filter_typename=no_filter,
     filter_branch=no_filter,
@@ -226,6 +227,7 @@ def concatenate(
     files,
     expressions=None,
     cut=None,
+    *,
     filter_name=no_filter,
     filter_typename=no_filter,
     filter_branch=no_filter,
@@ -672,6 +674,7 @@ class HasBranches(Mapping):
         self,
         expressions=None,
         cut=None,
+        *,
         filter_name=no_filter,
         filter_typename=no_filter,
         filter_branch=no_filter,
@@ -896,6 +899,7 @@ class HasBranches(Mapping):
         self,
         expressions=None,
         cut=None,
+        *,
         filter_name=no_filter,
         filter_typename=no_filter,
         filter_branch=no_filter,
@@ -1133,6 +1137,7 @@ class HasBranches(Mapping):
 
     def keys(
         self,
+        *,
         filter_name=no_filter,
         filter_typename=no_filter,
         filter_branch=no_filter,
@@ -1170,6 +1175,7 @@ class HasBranches(Mapping):
 
     def values(
         self,
+        *,
         filter_name=no_filter,
         filter_typename=no_filter,
         filter_branch=no_filter,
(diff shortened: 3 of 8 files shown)


jpivarski:
How do we query everyone to ask if these are "sensible" places? Well, we can always back off—move the keyword-only `*` later in the list—that direction is backward compatible. Maybe we'll start here and if the first adopters of Uproot 5.0.0 complain, we'll move them back for 5.0.1 and subsequent.

alexander-held (commented):
As a user of the high-level API, these changes all look very sensible to me. I cannot see any argument that is made keyword-only here which I would rather provide as a positional argument; doing so would be compromising readability of the code.

When it comes to reading code, I would personally prefer also seeing `expressions` and `cut` written as kwarg (the latter more so than the former). Users are of course still free to do so.

Another approach to determine where positional-only arguments should start could be making anything that has a default value positional-only. I do not see a significant gain in that approach compared to the current implementation though.

jpivarski:
Thanks for your comments! The reason I didn't make `expressions` keyword-only is because

```python
arrays = tree.arrays(["branch1", "branch2", "branch3"])
```

is a pretty common pattern.

An argument for making people write

```python
arrays = tree.arrays(expressions=["branch1", "branch2", "branch3"])
```

would be to avoid the confusion that these are, in fact, expressions, so `"branch*"` does not match all branches; it's a syntax error because `*` is multiplication and there's nothing on the right hand side. (A user who does this really wants `filter_names`.)

An argument against it is that if `expressions` were keyword-only, the error message would say "the 'arrays' method takes 0 positional arguments" and that would nudge newbies to write

```python
arrays = tree.arrays()
```

which means "read all branches," which has bad performance consequences. I wish the Python error message would list the names of the keyword-only arguments; maybe some new version of Python does or some future version will.

If `expressions` is positional, then `cut` should be positional as well, because they're similar sorts of things. (The first is a mapper, the second is a filter.) And there's a long tradition of expressions and cuts being the first two positional arguments. In [PAW](https://cdsweb.cern.ch/record/2296392), you'd write

```fortran
nt/plot 10.sqrt(px**2+py**2) abs(eta)<2.4.and.isolation>0.5 ! ! ! ! 100
```

to plot the expression `sqrt(px**2+py**2)` from an ntuple named `10`, filtered by `abs(eta)<2.4.and.isolation>0.5`, and fill a histogram named `100`. The expressions and the cuts are the first two arguments; the histogram-to-fill is the seventh argument in a language without keywords so the `!`s skip the middle four.

Anyway, point being that the order

   1. expression
   2. cut
   3. other stuff

is a long-standing convention. I'm afraid that making people type these names out will start a revolt.

alexander-held:
Thanks for all those additional details! I agree with your conclusion, the choice here seems very well motivated.

eduardo-rodrigues:
Hi @jpivarski, seeing your comment above on Python error messages. Have you considered the `(*, )` argument? The following seems understandable to me:
```
In [1]: def test1(x=0):
   ...:     print(x)
   ...:
In [2]: def test2(*, x=0):
   ...:     print(x)
   ...:
In [3]: test1()
0
In [4]: test2()
0
In [5]: test1(5)
5
In [6]: test2(5)
TypeError                                 Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 test2(5)

TypeError: test2() takes 0 positional arguments but 1 was given

In [7]: test1(x=5)
5

In [8]: test2(x=5)
5
```
Just a quick thought.

eduardo-rodrigues:
OK.

jpivarski:
Yeah, that's what I meant. I wish the error message wasn't


```
TypeError: arrays() takes 2 positional arguments but 3 were given
```

but

```
TypeError: arrays() takes 2 positional arguments but 3 were given; non-positional arguments are

    'filter_name', 'filter_typename', 'filter_branch', 'aliases', 'language', 'entry_start', 'entry_stop',
    'decompression_executor', 'interpretation_executor', 'array_cache', 'library', 'ak_add_doc', 'how'
```

which tells the user they weren't wrong for thinking the function might take more arguments; they just did it in the wrong way. That list of names would also act as a prompt that it's worth looking this function up in the documentation to find out what some of those other arguments do.

PR #788, opened 2022-11-28
feat: adjust for name change in scikit-hep/awkward#1919.

jpivarski:
The tests in this PR will fail until the change in https://github.com/scikit-hep/awkward/pull/1919 makes its way into a release, 2.0.0rc5 (probably). Once that happens, these tests will pass and all others will fail.

Follow the plan described in https://github.com/scikit-hep/awkward/pull/1919#pullrequestreview-1196679809.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
-    "awkward>=2.0.0rc2",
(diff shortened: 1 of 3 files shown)


jpivarski:
@agoose77 Now that the name change has propagated to an Awkward release, this Uproot PR should follow. (And then an Uproot RC release.)

PR #806, opened 2022-12-17
fix: uproot.dask: Protect against `branches=None` in `project_columns`

douglasdavis:
In `uproot.dask`'s `project_columns` implementations we need to make sure `branches` is not `None` before looping over the argument. (The optimization in `dask-awkward` will run project_columns with `None` branches if the the optimization determines that all original branches are necessary). more info at https://github.com/dask-contrib/dask-awkward/issues/134 and https://github.com/dask-contrib/dask-awkward/pull/131

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -597,9 +597,11 @@ class _UprootRead:
         )
 
     def project_columns(self, branches):
+        if branches is not None:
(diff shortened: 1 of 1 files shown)


kkothari2001 (approved):
LGTM!

jpivarski:
I agree. I'll merge it!

PR #705, opened 2022-09-05
feat: Support writing of RNTuple

Moelf:

PR #812, opened 2023-01-06
fix: an uproot.dask test was wrong; revealed by new dask-awkward.

jpivarski:
The `da[da.int_branch < 0]` expression should return an array with all fields, not just `"int_branch"`. An update in dask-awkward (2023.1.0) got it right, and older version (2022.12a3.dev6+g7fe448a) got it wrong, and presumably the old test was modeled on the old dask-awkward result.

There's still an error in these tests, which comes from scikit-hep/scikit-hep-testdata#111. A ROOT file was updated that the Uproot RNTuple test can no longer read. @Moelf?

### DIFF
--- a/src/uproot/const.py
+++ b/src/uproot/const.py
@@ -4,6 +4,7 @@
 This module defines integer constants used by serialization and deserialization routines.
 """
 
+import struct
 
 import numpy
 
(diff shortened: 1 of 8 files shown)


Moelf:
yeah, the `array<>` was not handled before, so need #705

Moelf (approved):

PR #819, opened 2023-01-26
fix: separate AwkwardForth machine for each TBranch context.

jpivarski:


aryan26roy (approved):

aryan26roy:
@jpivarski I just looked through it and the changes are exactly what we talked about in the meeting. Just to be sure I tried to reproduce the error locally and it ran without any errors.

PR #825, opened 2023-02-16
chore: move to Ruff

henryiii:
- chore: move to using Ruff
- style: refactor remaining issues


src/uproot/behaviors/TBranch.py
```diff
@@ -2817,7 +2805,7 @@ def _regularize_expression(
                 arrays,
                 expression_context,
                 branchid_interpretation,
-                symbol_path + (symbol,),
+                (*symbol_path, symbol),
```
jpivarski, line 2808: That's cute: I didn't know about that. This is available in Python 3.7? I guess it is, since those tests passed.

src/uproot/interpretation/identify.py
```diff
@@ -110,9 +110,8 @@ def _from_leaves_one(leaf, title):
     m = _title_has_dims.match(title)
     if m is not None:
         dims = tuple(int(x) for x in re.findall(_item_dim_pattern, title))
-        if dims == ():
-            if leaf.member("fLen") > 1:
-                dims = (leaf.member("fLen"),)
+        if dims == () and leaf.member("fLen") > 1:
+            dims = (leaf.member("fLen"),)
```
jpivarski, line 114: Sometimes, nested `if` instead of a single predicate with `and` is for code symmetry. I don't see that in this case, though.

jpivarski (approved):

src/uproot/behaviors/TBranch.py
henryiii: I think it was new in 3.0 (splat operators work in more places) - it was at least there by 3.4, for sure. It's also why `{**a, **b}` is a Python 3 way to merge dictionaries (though losing custom classes, unlike 3.9 or 3.10 with `|` on dicts).

henryiii:
Once this goes in I'll try turning on the test folder and seeing what the flake8-pytest plugin finds.

PR #827, opened 2023-02-16
fix: `_awkward_forth.symbol_dict` was missing `np.dtype(">i1")`

jpivarski:
I also sorted the dict, to make it easier to spot when something is missing.


jpivarski:
I'm going to merge this PR because these are two correct fixes—improvements—even if the bug reported in #826 is not completely fixed yet.

PR #820, opened 2023-01-26
fix: separate ZstdDecompressor for each thread.

jpivarski:
Does this fix the

```python
>>> fut.compute(scheduler="synchronous")
<Array [[469, 459, 188, 17.8], ..., [414, ...]] type='203939 * var * float32'>
>>> fut.compute(scheduler="threads")
Traceback (most recent call last):
...
zstd.ZstdError: decompression error: Corrupted block detected
```

issue, @nsmith-?

PR #828, opened 2023-02-16
fix: separate ZstdDecompressor for each thread (re-do)

nsmith-:
Continuation of #820


jpivarski (approved):
Ah, right! The `_DecompressZSTD` class is only created in one thread and used in the other threads. So it won't have the `obj` yet.

Have you tested against this? I think it's correct (more correct than the previous attempt).

nsmith-:
Yes, with this change my test works.

PR #830, opened 2023-02-17
feat: allow `uproot.dask` to re-map forms at the data source

lgray:
This PR implements the interface (to be merged) defined in https://github.com/douglasdavis/dask-awkward/pull/3 to implement form remapping in `uproot.dask`.

I have confirmed this so far with local testing, ideas for a minimal test would be appreciated!

@jpivarski

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -21,6 +21,7 @@ def dask(
     custom_classes=None,
     allow_missing=False,
     open_files=True,
+    form_mapping=None,
     **options,
 ):
     """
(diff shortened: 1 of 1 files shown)


src/uproot/_dask.py
```diff
@@ -63,6 +64,9 @@ def dask(
             the dask collections. If False, the opening of the files and reading the
             metadata is also delayed till the compute call. In this case, branch-names
             are inferred by opening only the first file.
+        form_mapping (Callable[awkward.forms.form] -> awkward.forms.form | None): If not none
```
jpivarski, line 67:
```suggestion
        form_mapping (Callable[awkward.forms.Form] -> awkward.forms.Form | None): If not none
```

jpivarski: `awkward` is not loaded here (and it can't be, not at module level). However, I don't think this docstring depends on that.

jpivarski (approved):
It looks good: I see that the public API of `uproot.dask` is increased by one argument, `form_mapping`. There's no equivalent for it in eager array functions, but I don't think there needs to be.

lgray:
For eager mode I don't think there's a need for encapsulating this concept.

jpivarski:
I was on the fence about whether this should be considered a fix or a feature. I guess you're right that it's a feature. It's not something that we had promised before.

jpivarski:
@all-contributors please add @lgray for code

PR #831, opened 2023-02-17
fix: Uproot should be able to run without Awkward.

jpivarski:
I know, I know, we'll have to add a test runner without Awkward installed to test this. For now, I'm just getting it back into a can-run-without-Awkward state and we'll add the test runner later.


agoose77 (approved):
LGTM!

PR #832, opened 2023-02-17
fix: ak_add_doc should add docs to both the lazy and the materialized array.

jpivarski:
This is essentially completing #784, which added docstrings to materialized arrays, but not the typetracers that Dask uses.


jpivarski:
Actually, it's better to reuse `interp_options`, to have fewer "moving parts" (i.e. multiple representations of the same data that could, in principle, become inconsistent with each other). So thanks for catching that!

PR #834, opened 2023-02-18
fix: let form mappings apply a behavior

lgray:


lgray:
Ah - sorry this one requires things to get merged in dask-awkward, it is presently not possible in uproot to do it.

src/uproot/_dask.py
```diff
@@ -772,7 +772,11 @@ def _get_meta_array(
     if form_mapping is not None:
         form = form_mapping(form)
     empty_arr = awkward.from_buffers(
-        form, 0, {"": b"\x00\x00\x00\x00\x00\x00\x00\x00"}, buffer_key=""
+        form,
+        0,
+        {"": b"\x00\x00\x00\x00\x00\x00\x00\x00"},
+        buffer_key="",
+        behavior=None if form_mapping is None else form_mapping.behavior,
     )
```
jpivarski, line 780: Nowadays, we can replace this whole thing with

```python
empty_arr = form.length_zero_array(behavior=None if form_mapping is None else form_mapping.behavior)
```

[defined here](https://github.com/scikit-hep/awkward/blob/4e01cc40e033cb9d06eee869cca73c4be3b0d53c/src/awkward/forms/form.py#L534-L547).

lgray:
Looks like the new release of dask-awkward broke some existing tests.

lgray:
@jpivarski could you re-run the tests on this branch, it should pass now?

jpivarski (approved):

PR #835, opened 2023-02-18
feat: pass tree title as a docstr when remapping

lgray:
Define top-level docstring for tree by kwarg.

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -770,7 +770,7 @@ def _get_meta_array(
         common_keys,
     )
     if form_mapping is not None:
-        form = form_mapping(form)
+        form = form_mapping(form, docstr=ttree.title)
     empty_arr = awkward.from_buffers(
         form, 0, {"": b"\x00\x00\x00\x00\x00\x00\x00\x00"}, buffer_key=""
     )


src/uproot/_dask.py
```diff
@@ -770,7 +770,7 @@ def _get_meta_array(
         common_keys,
     )
     if form_mapping is not None:
-        form = form_mapping(form)
+        form = form_mapping(form, docstr=ttree.title)
```
jpivarski, line 773: With #831, if `ak_add_doc=True`, then the docstrings would already be incorporated into the `form` that `form_mapping` is given; it can just preserve them.

Oh! This is the title of the whole `ttree`, not just the branches. How about if we handle that the same way as the branches? That is, `ak_add_doc=True` would mean

  * each record field's `__doc__` is set by the corresponding `TBranch.title`
  * the whole record's `__doc__` is set by the `TTree.title`

and these would already be in the `form` that is passed to `form_mapping`, so `form_mapping` can just be careful to preserve them. (Instead of having more "moving parts.")

lgray: Yes this would work well.

PR #836, opened 2023-02-18
Add ttree title to __doc__ of top level record

lgray:
for ```x = uproot.dask(ttree, ak_add_doc=True``` this will cause x to have parameter `__doc__` to be set to `ttree.title`


src/uproot/behaviors/TBranch.py
```diff
@@ -892,10 +892,16 @@ def get_from_cache(branchname, interpretation):
...
-        return _ak_add_doc(
+        out = _ak_add_doc(
             library.group(output, expression_context, how), self, ak_add_doc
         )
 
+        if ak_add_doc and type(out).__module__ == "awkward.highlevel":
+            out.layout.parameters["__doc__"] = self.title
+            out.__doc__ = self.title
+
+        return out
```
jpivarski, line 903: The `_ak_add_doc` function that has just been applied does exactly what these new lines do:

https://github.com/scikit-hep/uproot5/blob/27b3956a431907a3860e6262fa6ecff00b5af75e/src/uproot/behaviors/TBranch.py#L539-L542

except for adding the Python dunder `__doc__`. If that's necessary, we could add it to the `_ak_add_doc` function... but it looks like `ak.Array` already does that, too:

https://github.com/scikit-hep/awkward/blob/4e01cc40e033cb9d06eee869cca73c4be3b0d53c/src/awkward/highlevel.py#L231-L233

In principle, the entire highlighted change should do nothing. Did you see a difference in your output?

src/uproot/behaviors/TBranch.py
```diff
@@ -1123,6 +1129,10 @@ def iterate(
                     ak_add_doc,
                 )
 
+                if ak_add_doc and type(out).__module__ == "awkward.highlevel":
+                    out.layout.parameters["__doc__"] = self.title
+                    out.__doc__ = self.title
+
```
jpivarski, line 1135: Same comment as above. (If I could highlight 3 more lines, it would include the part where `_ak_add_doc` is already applied to `out`.)

src/uproot/_dask.py
```diff
@@ -775,6 +775,8 @@ def _get_meta_array(
         contents.append(content_form)
 
     form = awkward.forms.RecordForm(contents, common_keys)
+    if ak_add_doc:
+        form = form.copy(parameters={"__doc__": ttree.title})
```
jpivarski, line 779:
```suggestion

    parameters = {"__doc__": ttree.title} if ak_add_doc else None

    form = awkward.forms.RecordForm(contents, common_keys, parameters=parameters)
```

jpivarski (changes requested):
Two of these changes shouldn't have any effect (see below)[^1]. Did you observe one?

[^1]: I'm not saying that it didn't—bugs and oversights happen.

src/uproot/behaviors/TBranch.py
lgray: I'm pretty sure I wasn't seeing the parameter show up correctly until this one. I will check again.

lgray:
OK - was able to remove all the extra bits of paranoia, it works fine without them after a rebase locally.

jpivarski (approved):

Issue #833, opened 2023-02-17
Error in untested AwkwardForth case


jpivarski:
### Discussed in https://github.com/scikit-hep/uproot5/discussions/826

<div type='discussions-op-text'>

<sup>Originally posted by **renyhp** February 16, 2023</sup>
Hi, I have just installed uproot, but I can't get it to read a tree.

My code is very simple:

```python3
import uproot
f = uproot.reading.open("myfile.root")
t = f.get("aTree")
t['TheOnlyBranch'].array()
```

This just crashes with this stacktrace:
```text
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/venv/lib/python3.10/site-packages/uproot/behaviors/TBranch.py", line 834, in arrays
    _ranges_or_baskets_to_arrays(
  File "/tmp/venv/lib/python3.10/site-packages/uproot/behaviors/TBranch.py", line 3124, in _ranges_or_baskets_to_arrays
    uproot.source.futures.delayed_raise(*obj)
  File "/tmp/venv/lib/python3.10/site-packages/uproot/source/futures.py", line 36, in delayed_raise
    raise exception_value.with_traceback(traceback)
  File "/tmp/venv/lib/python3.10/site-packages/uproot/behaviors/TBranch.py", line 3066, in basket_to_array
    basket_arrays[basket.basket_num] = interpretation.basket_array(
  File "/tmp/venv/lib/python3.10/site-packages/uproot/interpretation/objects.py", line 139, in basket_array
    output = self.basket_array_forth(
  File "/tmp/venv/lib/python3.10/site-packages/uproot/interpretation/objects.py", line 206, in basket_array_forth
    output = self._discover_forth(
  File "/tmp/venv/lib/python3.10/site-packages/uproot/interpretation/objects.py", line 286, in _discover_forth
    output[i] = self._model.read(
  File "/tmp/venv/lib/python3.10/site-packages/uproot/containers.py", line 609, in read
    return uproot.deserialization.read_object_any(
  File "/tmp/venv/lib/python3.10/site-packages/uproot/deserialization.py", line 274, in read_object_any
    obj = cls.read(chunk, cursor, context, file, selffile, parent)
  File "/tmp/venv/lib/python3.10/site-packages/uproot/model.py", line 1372, in read
    versioned_cls.read(
  File "/tmp/venv/lib/python3.10/site-packages/uproot/model.py", line 867, in read
    self.read_members(chunk, cursor, context, file)
[...]

PR #838, opened 2023-02-20
fix: complain about CannotBeAwkward earlier, before reading data.

jpivarski:
This came up in Discussion #826 (Issue #833): the AwkwardForth errors occur on an Interpretation that could never have been Awkward anyway.

An early hint was that it was going through [read_object_any](https://github.com/scikit-hep/uproot5/blob/c24159917ea5732ada3bb0b03615707823194197/src/uproot/deserialization.py#L184-L319), which means arbitrary-typed pointers, circular references, and a whole host of things that are beyond the scope of Awkward Array, but the clearest indication was that we got the "Cannot be Awkward" error in TTrees with zero entries—if it could have gotten past the AwkwardForth read, it would have encountered this error, anyway. (It didn't get past the AwkwardForth read because we don't generate AwkwardForth inside `read_object_any`.)

So this PR

  * takes the check out of `Library.finalize`
  * puts it into `_ranges_or_baskets_to_arrays`, on the main thread, before any reading starts[^1]
  * fixes an unrelated error in `_util`, but it was uncovered by this PR.

[^1]: Well, actually, `Source.chunks` has already been called, so the bytes have been read from local files or a remote server has been queried, but we haven't waited for any data to come back yet and won't be decompressing it or interpreting it if it does.

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -570,6 +570,14 @@ def awkward_form(model, file, context):
                 _primitive_awkward_form[model] = awkward.forms.from_json('"float32"')
             elif model == numpy.dtype(numpy.float64):
                 _primitive_awkward_form[model] = awkward.forms.from_json('"float64"')
+            elif model.fields is not None:
+                fields = []
+                contents = []
+                for field, (dtype, _) in model.fields.items():
+                    fields.append(field)
+                    contents.append(awkward_form(dtype, file, context))
+                # directly return; don't cache RecordForms in _primitive_awkward_form
+                return awkward.forms.RecordForm(contents, fields)
             else:
                 raise AssertionError(f"{model!r}: {type(model)}")
 
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -2479,6 +2479,26 @@ in file {}""".format(
 
         return self
 
+    def _awkward_check(self, interpretation):
+        try:
+            interpretation.awkward_form(self.file)
+        except uproot.interpretation.objects.CannotBeAwkward as err:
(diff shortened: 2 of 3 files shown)


src/uproot/_util.py
```diff
@@ -570,6 +570,14 @@ def awkward_form(model, file, context):
...
             elif model == numpy.dtype(numpy.float64):
                 _primitive_awkward_form[model] = awkward.forms.from_json('"float64"')
+            elif model.fields is not None:
+                fields = []
+                contents = []
+                for field, (dtype, _) in model.fields.items():
+                    fields.append(field)
+                    contents.append(awkward_form(dtype, file, context))
+                # directly return; don't cache RecordForms in _primitive_awkward_form
+                return awkward.forms.RecordForm(contents, fields)
```
jpivarski, line 580: This is unrelated to the main purpose of the PR, but calling the `awkward_forth` method at this point in the code revealed that the leaf-list case isn't handled. Somehow, `test_0033-more-interpretations-2.py::test_leaflist_awkward` was passing before because something must have figured out that it needs a `RecordArray` without calling `awkward_forth`. (I'm not sure which code path that is, but now, at least, this is consistent with that.)

ioanaif (approved):
LGTM! Should tests be added for this issue?

jpivarski:
The file that revealed the issue is private—otherwise, that would have been used as a test.

However, every existing test with `library="ak"` goes through this new code path, so it is being tested in the sense of having full coverage. In fact, the unrelated bug (highlighted above) was discovered because it went through this code path, and in the old system, it somehow managed to avoid the old (late) code path. I'd take that as an indication that coverage has increased.

Issue #814, opened 2023-01-10
uproot fails to read branch of TVector3 created with RDF::Snapshot


davehadley:
The following example script generates two files with the same data in slightly different ways. `uproot` can successfully read the first file. But it fails to read the second file (created with `RDF::Snapshot`).
```python
import ROOT
import uproot

def make_file_1_with_root():
    tfile = ROOT.TFile("tfile_with_tvector3_1.root", "RECREATE")
    tree = ROOT.TTree("tree", "tree")
    tvector3 = ROOT.TVector3()
    tree.Branch("tvector3", tvector3)
    for x in range(10):
        tvector3.SetX(x)
        tree.Fill()
    tree.AutoSave()
    tfile.Close()

def make_file_2_with_root():
    tfile = ROOT.TFile("tfile_with_tvector3_1.root", "READ")
    tree = tfile.Get("tree")
    rdf = ROOT.RDataFrame(tree)
    branchlist = ROOT.std.vector(ROOT.std.string)()
    branchlist.push_back("tvector3")
    rdf.Snapshot("tree", "tfile_with_tvector3_2.root", branchlist)
    return

def read_file_1_with_uproot():
    print("reading file 1")
    uproot.open("tfile_with_tvector3_1.root:tree").arrays()

def read_file_2_with_uproot():
    print("reading file 2")
    uproot.open("tfile_with_tvector3_2.root:tree").arrays()


make_file_1_with_root()
make_file_2_with_root()
read_file_1_with_uproot() # uproot successfully reads file 1
read_file_2_with_uproot() # uproot fails to read file 2
```

`uproot.open("tfile_with_tvector3_2.root:tree").arrays()` fails with the error message:

```
Traceback (most recent call last):
  File "/home/dave/work/reproduce_uproot_bug/reproduce_uproot_bug.py", line 36, in <module>
    read_file_2_with_uproot() # uproot fails to read file 2
  File "/home/dave/work/reproduce_uproot_bug/reproduce_uproot_bug.py", line 30, in read_file_2_with_uproot
    uproot.open("tfile_with_tvector3_2.root:tree").arrays()
  File "/home/dave/work/reproduce_uproot_bug/env/uprootbug/lib/python3.10/site-packages/uproot/behaviors/TBranch.py", line 834, in arrays
    _ranges_or_baskets_to_arrays(
[...]

PR #841, opened 2023-02-23
fix: models should not be wrapped in a ListOffsetForm

ioanaif:
Fix for #814


jpivarski (approved):
(The PR review is on Zoom. We talked about all of this, and it looks great!)

jpivarski:
@all-contributors please add @ioanaif for code

PR #843, opened 2023-02-28
chore: don't repeat awkward dependency

agoose77:
This is a code cleanup PR — we already define `awkward` as a core dependency, so we don't need to repeat this in the other dependency groups now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,6 @@ dynamic = [
 
 [project.optional-dependencies]
 dev = [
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Because it's on line 41. 👍

jpivarski:
But something happened in the tests. I'm going to see if `main` is affected: https://github.com/scikit-hep/uproot5/actions/runs/4307341032

PR #846, opened 2023-03-01
fix: update licence url for tests

ioanaif:


jpivarski (approved):
Thanks! There is a redirect from https://github.com/scikit-hep/uproot4 to https://github.com/scikit-hep/uproot5, but it's better to have non-confusing URLs.

If you do a multi-file grep, you'll probably find more (such as the license URLs for source code files).

PR #851, opened 2023-03-02
fix: replaced incorrect AwkwardForth `var_set` logic with a check to see if the Form is complete

### LINKED ISSUE #850: Very slow reading of some (not all) vector<vector<int>> variables
Dear uproot,

One of my ROOT file contains `vector<vector<int>>` variables which are all produced in a very similar way.
Out of 4, 2 of them are extremely slow to read with uproot (5.0.3) : 
```python
t= uproot.open("test.root:gnnjet")
>>> timeit.timeit("a=t.arrays(['nnTruthReco' ])", globals=globals(), number=1)
6.898946930188686
>>> timeit.timeit("a=t.arrays(['nnRecoReco' ])", globals=globals(), number=1)
0.03529514092952013
>>> timeit.timeit("a=t.arrays(['nnRecoTruth' ])", globals=globals(), number=1)
5.870497648837045
>>> timeit.timeit("a=t.arrays(['nnTruthTruth' ])", globals=globals(), number=1)
0.06448978511616588
```

As you can see there seem to be a factor>100 in the read time for 2 of the variables. 
This is a  test file with 1000 entries which can be found at : 
https://cernbox.cern.ch/s/rvpBMjWLCZfhatG

My complete files have ~100k entries  so they are practically unreadable...
Is there something wrong with these files ?  Or is the slowness coming from uproot ?

Thanks for any hint !

P-A

```python
# python version is 3.10.9
>>> import uproot
>>> uproot.__version__
'5.0.3'
```

jpivarski:
@mpad saw factors of 200× speed difference between some TBranches in #850. It turned out to depend on whether any `std::vectors` are empty or not: TBranches that contain any empty `std::vectors` in all entries were slow, those that didn't were fast.

It's because the AwkwardForth code-generation step has to consider, at the end of each entry, whether the code that it has is complete or if it's not complete because it's never observed a given list as occupied (and therefore could learn how many bytes its headers are, if there had been any objects in it). The logic, as-written, would mark an entry incomplete (`var_set = True`) if any list anywhere within it is empty. However, data like this:

```
[[[]], [[]], [[]], [[]], [[]], [[123]]]
```

or this:

```
[[[123]], [[]], [[]], [[]], [[]], [[]]]
```

should not be considered incomplete because the walk over data reached all parts of the corresponding type tree _at some point_. Rather than asking if there is _any_ empty list, the appropriate question would be whether they're _all_ empty lists.

But there's more than that: this determination must be made separately at each level of depth (and in multiple branches, possibly at the same level, if there are any C++ classes/Awkward record arrays). It needs to reach all points of the _type_ tree for any part of the _value_ tree. To extend the current logic, `var_set` would have to become a container with as many items as list nodes in the type tree, the walk over values would start pessimistic and would toggle to a good value if any list value for a given list type node has non-zero length, and the final check would require all items to be good values.

Perhaps it could have been solved by using `ForthGenerator.get_keys(1)` to identify the current list node in the type tree during the walk over values, though I would first need to know the full set of keys to expect. (It could be done by examining the Form, generated before reading any data using `AsObjects.awkward_form(TBranch.file)`.)

However, I think there's an easier way, implemented in this PR. The same walk that generates AwkwardForth code generates an Awkward Form. If we don't get to a particular node to discover whether how many bytes its header has (to put the right `skip` command into the AwkwardForth code), then we also don't get to make the corresponding Form node. They're built in tandem. (Right?) So instead of checking `var_set`, I added a check to the current state of the Form (which is made of Python dicts and lists), to see if there are any missing nodes. Missing nodes are all labeled by a string, `"NULL"`. (Right?) If the Form does not contain any `"NULL"`, then the Forth code must be complete, too, and we can consider the code-generation phase to be complete.

Also, this means we can remove the `var_set` entirely, which didn't have a good name, anyway. (It wasn't clear what it was supposed to mean by just looking at it.)

@aryan26roy, could you check my assumptions? This PR fixes #850 and all of the tests still run. I also verified that one of the tests would have `var_set is False` after the first entry, so the new logic is being exercised to some degree. But if you can see any mistaken assumptions that wouldn't hold in general, I should fix them.

### DIFF
--- a/src/uproot/interpretation/objects.py
+++ b/src/uproot/interpretation/objects.py
@@ -270,9 +270,28 @@ class AsObjects(uproot.interpretation.Interpretation):
 
         return output
 
+    def _any_NULL(self, form):
+        if form == "NULL":
+            return True
+        else:
+            content = form.get("content")
+            if content is not None:
+                return self._any_NULL(content)
+            contents = form.get("contents")
+            if isinstance(contents, list):
+                for content in contents:
+                    if self._any_NULL(content):
+                        return True
+            elif isinstance(contents, dict):
+                for content in contents.values():
+                    if self._any_NULL(content):
+                        return True
+            return False
+
     def _discover_forth(self, data, byte_offsets, branch, context, cursor_offset):
         output = numpy.empty(len(byte_offsets) - 1, dtype=numpy.dtype(object))
         context["cancel_forth"] = False
+
         for i in range(len(byte_offsets) - 1):
             byte_start = byte_offsets[i]
             byte_stop = byte_offsets[i + 1]
@@ -281,8 +300,10 @@ class AsObjects(uproot.interpretation.Interpretation):
             cursor = uproot.source.cursor.Cursor(
                 0, origin=-(byte_start + cursor_offset)
             )
+
             if "forth" in context.keys():
-                context["forth"].gen.var_set = False
+                context["forth"].gen.count_obj = 0
+
             output[i] = self._model.read(
                 chunk,
                 cursor,
@@ -291,11 +312,15 @@ class AsObjects(uproot.interpretation.Interpretation):
                 branch.file.detached,
                 branch,
             )
+
             if context["cancel_forth"] and "forth" in context.keys():
                 del context["forth"]
+
             if "forth" in context.keys():
                 context["forth"].gen.awkward_model = context["forth"].gen.top_node
-                if not context["forth"].gen.var_set:
+
+                discovered_form = context["forth"].gen.top_form
+                if not self._any_NULL(discovered_form):
                     context["forth"].prereaddone = True
                     self._assemble_forth(
                         context["forth"].gen, context["forth"].gen.top_node["content"]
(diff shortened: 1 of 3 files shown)


src/uproot/interpretation/objects.py
```diff
@@ -281,8 +300,10 @@ def _discover_forth(self, data, byte_offsets, branch, context, cursor_offset):
             cursor = uproot.source.cursor.Cursor(
                 0, origin=-(byte_start + cursor_offset)
             )
+
             if "forth" in context.keys():
-                context["forth"].gen.var_set = False
+                context["forth"].gen.count_obj = 0
```
jpivarski, line 305: This is an unrelated fix. I noticed that when we do get to the next entry, the `count_obj` keeps increasing so `ForthGenerator.get_keys` returns numbers for keys that don't exist. For instance, the keys are

```python
["node0-offsets", "node1-offsets", "node2-data"]
```

and `get_keys` was returning `0` and `1` for offsets in the first entry, `3` and `4` in the second entry, etc. I could see that the integer was being used specifically to build references to these keys in the AwkwardForth code, so I'm quite certain that this is the right thing to do.

But if it doesn't look right to you, let me know.

src/uproot/interpretation/objects.py
```diff
@@ -291,11 +312,15 @@ def _discover_forth(self, data, byte_offsets, branch, context, cursor_offset):
...
+
             if context["cancel_forth"] and "forth" in context.keys():
                 del context["forth"]
+
             if "forth" in context.keys():
                 context["forth"].gen.awkward_model = context["forth"].gen.top_node
-                if not context["forth"].gen.var_set:
+
+                discovered_form = context["forth"].gen.top_form
+                if not self._any_NULL(discovered_form):
```
jpivarski, line 323: This is the essence of the change. We no longer look at `var_set` (or even construct it), but instead ask if there are any `"NULL"` values within the Form.

aryan26roy (approved):

PR #852, opened 2023-03-07
fix: strided interpretation for data with extra offsets

ioanaif:


ioanaif:
pre-commit.ci autofix

ioanaif:
I compared the last issue of the non-primitive dtypes being passed on to awkward in the test added by this PR with `tests/test_0034-generic-objects-in-ttrees.py::test_jagged_strided_awkward`

I tracked where the byte order changes. This happens in `interpretation.jagged.final_array()`:   

```
print("content dtype ", cnt['fP/fX'].dtype) => content dtype  >f8
content = numpy.empty((before,), self.content.to_dtype)
before = 0
for cnt in contents:
    content[before : before + len(cnt)] = cnt
    before += len(cnt)

print(“content dtype after “, content['fP/fX'].dtype) => content dtype after  float64

```

For the test I added, in `interpretation.numerical.final_array()` this transformation doesn’t take place so the byteorder is preserved and thus awkward complains about the fact that it gets a non-native dtype. This is solved by adding `output = output.newbyteorder('=')` before `self.hook_before_library_finalize`.

tests/test_0852-fix-strided-interp-extra-offsets.py
```diff
@@ -4,8 +4,6 @@
 import uproot
 import skhep_testdata
 
-ROOT = pytest.importorskip("ROOT")
```
jpivarski: Right, because the file already exists. This will allow the test to be tested on more systems. Thanks for catching this!

PR #870, opened 2023-04-05
feat: have unknown type become float64 when in array context

ioanaif:


src/uproot/writing/_cascadetree.py
```diff
@@ -324,6 +324,8 @@ def __init__(
         self._key = None
 
     def _branch_ak_to_np(self, branch_datashape):
+        if type(branch_datashape).__name__ == "UnknownType":
+            return numpy.dtype("float64")
         if type(branch_datashape).__name__ == "NumpyType":
```
jpivarski, line 329: This type check is part of the `if`-`elif`-`else` chain that follows, so let's make that explicit.

```suggestion
        elif type(branch_datashape).__name__ == "NumpyType":
```

A lot of these code patterns exist because I like the match-case structure from other languages, now in Python as of 3.10. It will be a long time before 3.10 becomes the minimum version, but then there would be an opportunity to turn a lot of these into explicit match-case chains.

jpivarski (approved):
Looks good! I'd just like you to apply the one suggestion below (`if` → `elif`), and I'll let you merge it when you think it's ready.

PR #871, opened 2023-04-06
feat: add support for pandas is_numeric API change

ioanaif:


jpivarski (approved):
Awesome! And I just verified that the full path to this available without having to explicitly import submodules within Pandas (i.e. it didn't work "by accident" this time and would fail later).

```python
>>> import pandas
>>> pandas.__version__
'2.0.0'
>>> pandas.api.types.is_any_real_numeric_dtype
<function is_any_real_numeric_dtype at 0x7fd0a7fe0ea0>
```

Also,

```python
>>> import pandas
>>> pandas.__version__
'1.5.3'
>>> pandas.api.types.is_any_real_numeric_dtype
AttributeError: module 'pandas.api.types' has no attribute 'is_any_real_numeric_dtype'
```

It could just check `AttributeError`, not both `AttributeError` and `ImportError`, but this is okay.

You can merge when ready. (Looks like you're already ready.)

jpivarski (approved):
These are all improvements! Go ahead and merge. I think your other PR needs this, and then after that's done, I can start a release. Am I forgetting anything?

ioanaif:
Just these two PRs #871  and #870.

Issue #79, opened 2020-08-26
uproot4 unable to open ROOT file with colons in the name that uproot3 can


matthewfeickert:
Hi. I've come across an issue where if a ROOT file has a colon in the name of it `uproot3` can open the file but `uproot4` fails. I can't give you the file, but I can show you a minimal failing example and then a reproducible example with public files.

## Minimal Failing Example

```
$ tree .
.
├── data-tree
│   └──  data16_13TeV:data16_13TeV.periodA.physics_Main.PhysCont.DAOD_JETM1.grp16_v01_p4061.root
├── issue.py
├── requirements.txt

1 directory, 3 files
$ cat requirements.txt 
uproot
uproot4
$ docker run --rm -it -v $PWD:/data -w /data python:3.8 /bin/bash
root@510598a7f4e8:/data# python -m pip install --upgrade pip setuptools wheel
root@510598a7f4e8:/data# python -m pip install -r requirements.txt
root@510598a7f4e8:/data# python --version
Python 3.8.5
root@510598a7f4e8:/data# python -m pip list
Package        Version
-------------- -------
awkward        0.13.0
cachetools     4.1.1
numpy          1.19.1
pip            20.2.2
setuptools     49.6.0
uproot         3.12.0
uproot-methods 0.7.4
uproot4        0.0.18
wheel          0.35.1
root@510598a7f4e8:/data# cp data-tree/data16_13TeV\:data16_13TeV.periodA.physics_Main.PhysCont.DAOD_JETM1.grp16_v01_p4061.root data-tree/renamed.root
root@510598a7f4e8:/data# python
Python 3.8.5 (default, Aug  5 2020, 08:22:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import uproot as uproot3
>>> import uproot4
>>> from pathlib import Path
>>> 
>>> uproot3_file = uproot3.open(
...     "data-tree/data16_13TeV:data16_13TeV.periodA.physics_Main.PhysCont.DAOD_JETM1.grp16_v01_p4061.root"
... )
>>> print(f"uproot3 opens file as {uproot3_file}")
uproot3 opens file as <ROOTDirectory b'/home/feickert/workarea/submitDir/data-tree//data16_13TeV:data16_13TeV.periodA.physics_Main.PhysCont.DAOD_JETM1.grp16_v01_p4061.root' at 0x7fce6da5b5e0>
>>> 
[...]

PR #875, opened 2023-04-12
fix: Ignore semicolon in EOS token when separating file name and object name

bcouturi:
To use EOS tokens, the token must be appended to the file name, so the URLs are typically of the style:

root://eosserver//eos/path/file.root?xrd.wantprot=unix&authz=zteos64:BASE64TOKEN

Currently EOS does not accept the ":" to be escaped as %3A, and the semicolon is interpreted by uproot as the separator between the file name and the object name in the file.

This patch just ignores the semicolon if preceded by zteos64.

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -293,6 +293,13 @@ def file_object_path_split(path):
         ):
             return path, None
 
+        # Unfortunately EOS tokens start with zteos:, so we should ignore
+        # the : when preceded by that string as it is part of the URL, and
+        # not a seperator between the file name and the path in the file
+        # Future versions of EOS will accept to have that character escaped.
+        if file_path.endswith(("zteos", "zteos64")):
+            return path, None
+
         file_path = file_path.rstrip()
         object_path = object_path.lstrip()
 


jpivarski:
Sorry, but having a special rule for a string as specific as "`zteos64`" is a bad idea: it will fix this case and break others in a way that's hard to guess.

(I _wish_, I _wish_, I _wish_ we never introduced that colon-parsing rule!)

There are two ways around it:

  * use a `pathlib.Path` instead of a bare string, since a `pathlib.Path` is assumed to be entirely a file-path (although using a URL as a `pathlib.Path` is a little unusual)
  * use the `{"FULL-PATH": None}` syntax.

In your case, the latter would be to replace

```python
"root://eosserver//eos/path/file.root?xrd.wantprot=unix&authz=zteos64:BASE64TOKEN"
```

with

```python
{"root://eosserver//eos/path/file.root?xrd.wantprot=unix&authz=zteos64:BASE64TOKEN": None}
```

That's how we solidified all of the paths in Coffea: https://github.com/CoffeaTeam/coffea/pull/771

In that pull request, you can see a lovely list of 10 Issues, PRs, and Discussions related to the colon parsing. This would be the twelveth, counting the Coffea PR.

(I _wish_, I _wish_, I _wish_ we never introduced that colon-parsing rule! Let this be a lesson about adding high-level user conveniences that complicate interfaces for library developers. I should have stuck with this opinion: https://github.com/scikit-hep/uproot5/issues/79#issuecomment-680861697.)

Adding the new rule in this PR would be going further in that direction.

bcouturi:
Hi Jim, fair enough! We can use the {"FULL-PATH": None} syntax indeed, that will be a lot cleaner.

PR #876, opened 2023-04-12
feat: add in capability for blindly splitting files into chunks for dask

lgray:
Presently using the lazy file-opening mode of dask causes dask to map one file to one partition.
For the typical size of CMS NanoAOD files this is much too large and will exhaust memory on worker nodes.
This PR adds the possibility, in the case of lazy file opening, to blindly chunk the files and avoid this problem.

@jpivarski following up from slack chat.


jpivarski:
Moving a conversation from Slack: @lgray reminds me that `chunks_per_file` only _needs_ to be applied to the two `open_files=False` cases (NumPy and Awkward), since `step_size` can be used in the `open_files=True` cases.

That's right, but we want interfaces to be as consistent as possible. We now have two ways of specifying the chunking within a file, and that's bad because if there's a fight between them, who wins? The interface should be designed to (1) highlight the fact that they're both addressing the same thing and (2) be very clear about which one is active.

For (1), let's start by giving them the same names. `step_size` is used all of Uproot; we should keep that name, but make the new argument `steps_per_file` instead of `chunks_per_file`.

For (2), we can remove confusion by making the two arguments explicitly incompatible. That is, the defaults of both `step_size` and `steps_per_file` would be `unset`, where `unset` is defined as it is in Awkward:

https://github.com/scikit-hep/awkward/blob/ee387cbe4ad41fb7407a965f5a91cc39f9e28cc3/src/awkward/_util.py#L83-L89

The first thing to check when entering the `uproot.dask` function is the state of the two arguments:

  * If both are not `unset` (i.e. if both have been set to some values, even their defaults), then raise a `TypeError` with a message saying that you can only set one or the other, not both.
  * If `step_size` is set to some value,
    * if `open_files=True`, use it;
    * if `open_files=False`, raise a `TypeError` saying that `step_size` is inconsistent with `open_files=False`.
  * If `steps_per_file` is set to some value, use it. For `open_files=True`, the `steps_per_file` value can immediately be converted into an equivalent `step_size`, and that's what can get passed down into the machinery. That is, the `ichunk` and `nchunk` only need to be propagated to the `_NumpyOpenAndRead` and `_UprootOpenAndRead` classes.
  * If both are `unset`, use `steps_per_file=1` regardless of whether `open_files` is `True` or `False`. This is a change, but it makes the behavior consistent between the two `open_files` cases. If this increases memory use for any current users, the recourse is to set one of these arguments.

That should do it. And then there needs to be tests for each case, and you can test both NumPy and Awkward by using a `pytest.parameterize` to switch between them.

jpivarski (approved):
@ioanaif, I see that you propagated @lgray's `_UprootOpenAndRead` changes to `_UprootOpenAndReadNumpy` :heavy_check_mark:, that you added a complete suite of tests :heavy_check_mark:, and that you added `unset` sentinel-based logic to ensure that `step_size` and `steps_per_file` can't both be set :heavy_check_mark:.

I added a few things:

  * Explained the incompatibility of `step_size` and `steps_per_file` in the docstring.
  * Converted the case-handling into an `if`-`elif`-`else` chain, with nesting for `step_size`/`steps_per_file` checking versus `open_files` checking, even though this means we add useless `pass` statements, so that the logic is more readable. With a series of unconnected `if` statements, I was having trouble being 100% sure that all cases were being handled properly.
  * `unset` is a singleton; we can use it with `is`. We don't have to check it via its type (as though there could be other instances of `_Unset` around). After all, if somebody created their own `_Unset` instances (ignoring that privacy underscore) and passed it as an argument to `step_size` or `steps_per_file`, then they _have_ assigned those arguments. Using `is` instead of `isinstance` makes the logic a lot easier to read and therefore be sure that it's correct.
  * The `memory_size` helper can't convert `unset` into a numerical value, since `unset` might be a placeholder for different arguments with different defaults in different functions. `unset` has to be handled before anything gets passed to this regularizer.
  * The earliest point where `steps_per_file` can be converted into a `step_size` is in the `_get_dask_array` and `_get_dak_array` functions, after the TTree metadata of all files has been read. So `steps_per_file` has to be passed into these functions for the case in which `steps_per_file` is assigned, `step_size` is unset, and `open_files=True`.

I'd say it's good now. Any counter-counter edits?

Issue #883, opened 2023-05-03
Can't make a hist out of a Geant4 histogram


henryiii:
When trying to make a Hist out of an geant4 histogram, you get the following error:

```python
import uproot
uproot_file = uproot.open("../scikit-hep-testdata/src/skhep_testdata/data/uproot-from-geant4.root")
uproot_file["cot_diff"].to_hist()
```

```pytb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/henryschreiner/git/scikit-hep/uproot-browser/.venv/lib/python3.11/site-packages/uproot/behaviors/TH1.py", line 277, in to_hist
    self.to_boost(metadata=metadata, axis_metadata=axis_metadata)
  File "/Users/henryschreiner/git/scikit-hep/uproot-browser/.venv/lib/python3.11/site-packages/uproot/behaviors/TH1.py", line 226, in to_boost
    axes = [
           ^
  File "/Users/henryschreiner/git/scikit-hep/uproot-browser/.venv/lib/python3.11/site-packages/uproot/behaviors/TH1.py", line 227, in <listcomp>
    _boost_axis(self.member(name), axis_metadata)
  File "/Users/henryschreiner/git/scikit-hep/uproot-browser/.venv/lib/python3.11/site-packages/uproot/behaviors/TH1.py", line 23, in _boost_axis
    if axis.member("fLabels") is not None:
       ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/henryschreiner/git/scikit-hep/uproot-browser/.venv/lib/python3.11/site-packages/uproot/model.py", line 556, in member
    raise uproot.KeyInFileError(
uproot.exceptions.KeyInFileError: not found: 'fLabels' because uproot.dynamic.Model_TAxis_v6 has only the following members:

    '@fUniqueID', '@fBits', 'fName', 'fTitle', 'fNdivisions', 'fAxisColor', 'fLabelColor', 'fLabelFont', 'fLabelOffset', 'fLabelSize', 'fTickLength', 'fTitleOffset', 'fTitleSize', 'fTitleColor', 'fTitleFont', 'fNbins', 'fXmin', 'fXmax', 'fXbins', 'fFirst', 'fLast', 'fTimeDisplay', 'fTimeFormat'

in file ../scikit-hep-testdata/src/skhep_testdata/data/uproot-from-geant4.root
```

[...]

PR #884, opened 2023-05-03
fix: histograms from Geant4

henryiii:
Possible fix for #883. Uproot-browser can now view histograms from Geant4.


jpivarski (approved):
See https://github.com/scikit-hep/uproot5/issues/883#issuecomment-1534029085.

PR #894, opened 2023-05-22
refactor: use public typetracer API

agoose77:
This PR 
- changes use of private APIs for public ones.[^note]
- replaces use of dask-awkward's test-util with a call to `ak.almost_equal`
- bumps the minimum dask-awkward version to the near-latest release

[^note]: we don't explicitly pin awkward for these features, but they're used in a dask-awkward context, so we can rely on the pin imposed there. It's not ideal, but I don't think we need to worry too much.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -488,6 +488,24 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "natsukium",
+      "name": "OTABI Tomoya",
+      "avatar_url": "https://avatars.githubusercontent.com/u/25083790?v=4",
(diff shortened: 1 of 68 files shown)


ioanaif (approved):
Great, this will also fix the latest failing tests by avoiding the helpers depending on a `pyarrow` import.

jpivarski:
Oops; I had independently fixed this with #957, but didn't see that this one already fixed it. Well, I resolved the conflict.

jpivarski (approved):
Also, I approve switching to the public interface. It's a much better position to be in, with respect to flexibility to change the two projects independently.

PR #898, opened 2023-06-07
feat: chunk specification in uproot.dask

### LINKED ISSUE #891: uproot.dask should be able to receive pre-defined chunks.
Extend the uproot.dask interface to allow specification of chunks along the lines of:
```python3
files = {"/some/file/path.root" : {"objectpath": "Events", "chunks": [[0, 10000], [15000, 20000], ...]}, ...}
```
or
```python3
files = {"/some/file/path.root" : {"objectpath": "Events", "chunks": [0, 10000, 15000, 20000, ...]}, ...}
```

Where the first version allows discontinuous chunks and the second specifies chunk offsets.

Discussed with @jpivarski in the last two months or so.

I'll provide an implementation example soon if it helps!

lgray:
Fixes #891

Example implemented for delayed-open.

---------------

From @jpivarski: implemented for four cases, `open_files=False`,`True` and `library="np"`,`"ak"`.

If any user-specified steps (we're calling them "steps," not "chunks" now) are out of bounds, this raises a `ValueError` in all four cases (though the `open_files=False` cases raise errors later than `open_files=True`).

If steps are specified, all four cases set Dask array `chunks` and dask-awkward `divisions`.

Also new: the `open_files=True` and `library="ak"` case wasn't setting `divisions`, though it should have been.

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -445,8 +445,14 @@ class _UprootOpenAndReadNumpy:
         self.key = key
         self.interp_options = interp_options
 
-    def __call__(self, file_path_object_path_istep_nsteps):
-        file_path, object_path, istep, nsteps = file_path_object_path_istep_nsteps
+    def __call__(self, file_path_object_path_istep_nsteps_ischunk):
+        (
+            file_path,
+            object_path,
+            istep_or_start,
+            nsteps_or_stop,
+            ischunk,
+        ) = file_path_object_path_istep_nsteps_ischunk
         ttree = uproot._util.regularize_object_path(
             file_path,
             object_path,
@@ -455,10 +461,12 @@ class _UprootOpenAndReadNumpy:
             self.real_options,
         )
         num_entries = ttree.num_entries
-        events_per_steps = math.ceil(num_entries / nsteps)
-        start, stop = (istep * events_per_steps), min(
-            (istep + 1) * events_per_steps, num_entries
-        )
+        start, stop = istep_or_start, nsteps_or_stop
+        if not ischunk:
+            events_per_steps = math.ceil(num_entries / nsteps_or_stop)
+            start, stop = (istep_or_start * events_per_steps), min(
+                (istep_or_start + 1) * events_per_steps, num_entries
+            )
 
         return ttree[self.key].array(
             library="np",
@@ -654,28 +662,46 @@ def _get_dask_array_delay_open(
         else:
             dt, inner_shape = dt.subdtype
 
+        partitions = []
         partition_args = []
         for ifile_iobject_maybeichunks in files:
             ifile_path, iobject_path = ifile_iobject_maybeichunks[0:2]
-            if len(ifile_iobject_maybeichunks) == 3:
-                ifile_iobject_maybeichunks[2]
 
-            for istep in range(steps_per_file):
-                partition_args.append(
-                    (
-                        ifile_path,
-                        iobject_path,
-                        istep,
-                        steps_per_file,
+            chunks = None
+            if len(ifile_iobject_maybeichunks) == 3:
+                chunks = ifile_iobject_maybeichunks[2]
+
+            if chunks is not None:
+                partitions.extend([stop - start for start, stop in chunks])
+                for start, stop in chunks:
+                    partition_args.append(
+                        (
+                            ifile_path,
+                            iobject_path,
+                            start,
+                            stop,
+                            True,
+                        )
                     )
-                )
+            else:
+                partitions.extend([numpy.nan] * steps_per_file)
+                for istep in range(steps_per_file):
+                    partition_args.append(
+                        (
+                            ifile_path,
+                            iobject_path,
+                            istep,
+                            steps_per_file,
+                            False,
+                        )
+                    )
 
         dask_dict[key] = _dask_array_from_map(
             _UprootOpenAndReadNumpy(
                 custom_classes, allow_missing, real_options, key, interp_options
             ),
             partition_args,
-            chunks=((numpy.nan,) * len(files) * steps_per_file,),
+            chunks=(tuple(partitions),),
             dtype=dt,
             label=f"{key}-from-uproot",
         )
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -47,16 +47,16 @@ def dask(
             with slashes (``/``); otherwise, use the descendant's name as
             the field name.
         step_size (int or str): If an integer, the maximum number of entries to
-            include in each chunk; if a string, the maximum memory_size to include
-            in each chunk. The string must be a number followed by a memory unit,
+            include in each chunk/partition; if a string, the maximum memory_size to include
+            in each chunk/partition. The string must be a number followed by a memory unit,
             such as "100 MB". Mutually incompatible with steps_per_file: only set
             step_size or steps_per_file, not both. Cannot be used with
             ``open_files=False``.
         steps_per_file (int, default 1):
-            Subdivide files into the specified number of chunks. Mutually incompatible
+            Subdivide files into the specified number of chunks/partitions. Mutually incompatible
             with step_size: only set step_size or steps_per_file, not both.
             If both ``step_size`` and ``steps_per_file`` are unset,
-            ``steps_per_file``'s default value of 1 (whole file per chunk) is used,
+            ``steps_per_file``'s default value of 1 (whole file per chunk/partition) is used,
             regardless of ``open_files``.
         library (str or :doc:`uproot.interpretation.library.Library`): The library
             that is used to represent arrays. If ``library='np'`` it returns a dict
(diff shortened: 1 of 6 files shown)


tests/test_0603-dask-delayed-open.py
```diff
@@ -38,3 +38,29 @@ def test_multiple_delay_open():
...
+    files = {
+        filename1: {"object_path": "events", "chunks": chunks1},
+        filename2: {"object_path": "events", "chunks": chunks2},
+    }
+
+    assert uproot.dask(files, open_files=open_files)["px1"].divisions == (
+        None,
+        None,
+        None,
+    )
```
jpivarski, line 61: I'm confused about this. Although `open_files=False` would ordinarily give you unknown divisions, specifying them explicitly should make them known. Isn't that the whole point of this PR?

(What happens when the files are eventually do get opened and the chunks are impossible because they go beyond the number of entries in the file is another question—maybe it should raise an error, maybe it should trim like a Python `slice`, etc.)

This is weird: they're never set!

```python
>>> filename1 = skhep_testdata.data_path("uproot-Zmumu.root")
>>> filename2 = skhep_testdata.data_path("uproot-Zmumu-uncompressed.root")
>>> uproot.dask([filename1, filename2], open_files=True)["px1"].divisions
(None, None, None)
>>> uproot.dask([filename1, filename2], open_files=False)["px1"].divisions
(None, None, None)
```

And yet the property has meaning:

```python
>>> import awkward as ak
>>> import numpy as np
>>> import dask_awkward as dak
>>> a = dak.from_awkward(ak.Array(np.arange(100)), npartitions=3)
>>> a
dask.awkward<from-awkward, npartitions=3>
>>> a.divisions
(0, 34, 68, 100)
```

The `open_files=False` case is sensitive to the number of chunks:

```python
>>> uproot.dask({filename1: {"object_path": "events", "chunks": [0, 2305]}}, open_files=False)["px1"].divisions
(None, None)
>>> uproot.dask({filename1: {"object_path": "events", "chunks": [0, 1000, 2305]}}, open_files=False)["px1"].divisions
(None, None, None)
>>> uproot.dask({filename1: {"object_path": "events", "chunks": [0, 1000, 2000, 2305]}}, open_files=False)["px1"].divisions
(None, None, None, None)
```

and

```python
>>> uproot.dask({filename1: {"object_path": "events", "chunks": [[0, 2305]]}}, open_files=False)["px1"].divisions
(None, None)
>>> uproot.dask({filename1: {"object_path": "events", "chunks": [[0, 1000], [1000, 2305]]}}, open_files=False)["px1"].divisions
(None, None, None)
>>> uproot.dask({filename1: {"object_path": "events", "chunks": [[0, 1000], [1000, 2000], [2000, 2305]]}}, open_files=False)["px1"].divisions
(None, None, None, None)
```

The `open_files=True` case isn't, but that's my job.

I noticed that this is already a case in which the `chunks` are asking for more events than are available. The first file has only 2304 events, so the whole range would be `[0, 2304]`. It doesn't complain about too-large values:

```python
>>> uproot.dask({filename1: {"object_path": "events", "chunks": [[0, 5000]]}}, open_files=False)["px1"].compute()
<Array [-41.2, 35.1, 35.1, 34.1, ..., 32.4, 32.4, 32.5] type='2304 * float64'>
```

I _suspect_ that you'll want it to raise errors. If you're providing exact numbers of entries, I think you'd want to know if the numbers are wrong.

lgray: It shouldn't be too hard to add in the divisions in this case, even if we haven't opened the files. I think that's a constructor argument for the dask array, or we can set them?

I think there's a huge and useful distinction between hinting the chunks (when you're sure your files won't change underneath you) and needed to calculate them from the files.

In particular this is useful because it is not *always* the case that we can uniformly split files into chunks, and if we want ~uniform processing time per-partition we need to be able to say in what way we want chunks.

jpivarski: Suppose that a chunk's `start` and `stop` are both greater than the number of entries. That partition would be empty. Guessing wrong and getting lots of empty partitions isn't a problem, is it?

lgray: In terms of the system complaining about things not lining up. 

The behavior in steps_per_file mode does tail clipping, but I guess we're expecting the user to supply something truly exact there.  Sure, that makes sense to me.

jpivarski:
Also, I think we'll need to rename "chunk" → "partition." Uproot has a specific meaning for [Chunk](https://github.com/scikit-hep/uproot5/blob/2d24309dffcb8eccd15687c37bb11108a4490f9f/src/uproot/source/chunk.py#L217-L470) that we don't want to confuse this with.

Although dask-array uses the word "chunk," I'm not seeing that in dask-awkward, which has `npartitions` and `divisions`.

I can do that, since I've got the files open.

--------------

Unrelated: if you specify chunks/partitions in the `files`, it will have to complain about using `steps_per_file` or `step_size`. We can't have people guessing about which takes precedence.

jpivarski:
> "chunk" → "partition"

Ack! Everywhere else in `uproot.dask`, we call these things "steps." That's to align with the convention set up `uproot.iterate`. Since we want it to be clear that these things are mutually exclusive with `step_size` and `steps_per_file`, I guess a better replacement is "chunk" → "step".

src/uproot/_dask.py
```diff
@@ -886,7 +915,11 @@ def _get_dak_array(
     is_self = []
 
     count = 0
-    for file_path, object_path in files:
+    for file_object_maybechunks in files:
+        file_path, object_path = file_object_maybechunks[0:2]
+        if len(file_object_maybechunks) == 3:
+            file_object_maybechunks[2]
```
jpivarski, line 921: This doesn't do anything: the `file_object_maybechunks[2]` is accessed but not assigned. You must need this information further down. Let me see if you get it another way...

lgray: checking - this may have been a leftover from previous development.

lgray: Yes, indeed that should be:
```python3
chunks = file_object_maybechunks[2]
```

to assign the chunks - I may have added that in to the wrong function first by mistake and forgot to delete it.

You'll see the correct implementation further down.

jpivarski: Got it. Then I'll delete this reference without assignment, since the active part is in the `_get_*_delay_open` functions.

tests/test_0603-dask-delayed-open.py
jpivarski: I'll leave it as clipping, rather than raising errors.

jpivarski: Change of plan (after talking on Slack): if any user-provided steps (i.e. "chunks") go out of bounds for any TTree, it raises an error.

This strictness has a benefit—if we know that we'll get the entry ranges we ask for or die trying, the Dask array `chunks` or dask-awkward `divisions` can be known before `.compute()`. As a consequence, there will be fewer `unknown_lengths` in type-tracer arrays, so unknown values propagate less, and downstream situations that need known lengths or values won't have to give up. It may be that some workflow is impossible with `open_files=False` and no steps in `files`, but possible with them (which is a nice incentive).

jpivarski:
@ioanaif, I asked for your review on this PR because I contributed to it. (Reviewers shouldn't be authors.)

src/uproot/_util.py
```diff
@@ -809,21 +809,32 @@ def objectarray1d(items):
...
+    else:
+        raise TypeError(
+            "'files' argument's steps must be an iterable of integer offsets or a list of pairs of integer starts and stops."
         )
 
     if len(out.shape) == 1:
-        out = numpy.stack((out[:-1], out[1:]), axis=1)
+        out = numpy.stack((out[:-1], out[1:]), axis=1).tolist()
 
     return out
```
lgray, line 840: Shouldn't it just be the numpy array here rather than tolist()? out starts out as a numpy array.

jpivarski: Oh, that was to normalize the types to plain Python. It's metadata, and although it's possible to iterate over NumPy arrays, it's less surprising for it to be Python `list` and `int`.

Also, the arrayification doesn't determine which integer type it is. If, for instance, it's an unsigned type, subtraction doesn't always work as expected. If it's int8, it could roll over. All sorts of bad things could happen that we just don't have to think about if it's turned into genetic `int`.

But being able to accept very generic types from users is also a benefit.

That's why I did that.

lgray: Sure but now the code paths return different types.

jpivarski: Yikes—it's supposed to be `.tolist()` for everything. That should have been on the `return`.

Are you in a position where you'd be able to do that easily? (I'm on a phone and my computer is shut down.)

Anyway, @ioanaif, this should be fixed before we merge.

jpivarski (approved):
I just noticed that @ioanaif is on break, so I'll review this PR. (Bending the rules, but not by much/with good reason.)

This PR is ready to go, once the tests pass.

nsmith-:
An optional TODO would be to re-implement also the alignment along cluster boundaries as done in https://github.com/CoffeaTeam/coffea/blob/f8a4eb97137e84dd52474d26b8100174da196b57/src/coffea/processor/executor.py#L115

lgray:
@nsmith- IMO that kind of logic should go in user code or be provided by another library. I don't think uproot should be making those kinds of decisions.

Better to have some code that does it and makes the cluster-aligned list of chunks.

OTOH implemented as a free function it kinda makes sense here so people can apply it or have a reference implementation of a chunk calculator.

PR #901, opened 2023-06-09
feat: unproject_layout support

douglasdavis:


src/uproot/_dask.py
```diff
@@ -780,22 +781,33 @@ def __call__(self, i_start_stop):
...
-            array = self.ttrees[i].arrays(
-                self.common_keys,
-                entry_start=start,
-                entry_stop=stop,
-                ak_add_doc=self.interp_options["ak_add_doc"],
+
+            return awkward.Array(
+                dask_awkward.lib.unproject_layout.unproject_layout(
+                    self.rendered_form,
+                    array.layout,
```
lgray, line 788: As this is right now it drops the behavior that's possibly passed with the form mapping.

PR #905, opened 2023-06-19
fix: if using form remapping start off with full list of remapped columns

### LINKED ISSUE #904: uproot.dask form remapping does not work when dask-awkward column optimization turned off
Logging it here since we discussed it over zoom / on friday.

Replicating the text of where the problem is:
"""
so the issue is that the nanoevents keys do not get subbed in until project_columns is called https://github.com/scikit-hep/uproot5/blob/main/src/uproot/_dask.py#L789 and it's only called when we optimize columns https://github.com/dask-contrib/dask-awkward/blob/main/src/dask_awkward/lib/optimize.py#L124 (iiuc)
"""

So I think we can make it so that when uproot.dask creates the layer we make sure to pass the correct dask-awkward form keys.

lgray:
Fixes #904

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -1139,7 +1139,7 @@ which has {entry_stop} entries"""
     return dask_awkward.from_map(
         _UprootRead(
             ttrees,
-            common_keys,
+            common_keys=common_keys if form_mapping is None else form.columns(),
             common_keys,
             interp_options,
             form_mapping=form_mapping,
@@ -1230,7 +1230,7 @@ def _get_dak_array_delay_open(
             custom_classes,
             allow_missing,
             real_options,
-            common_keys,
+            common_keys=common_keys if form_mapping is None else form.columns(),
             common_keys,
             interp_options,
             form_mapping=form_mapping,


lgray:
@iasonkrom can you give this PR a spin and see if it fixes your tests without column-opt turned on?

iasonkrom:
Yup, they do pass! The only problem is that they take way too long without column optimization. Testing the analysis tools now went up from 1 min to 1 hr.  Can this be sped up by maybe reading way less with `NanoEventsFactory`?

lgray:
How big is the file you're testing on?

iasonkrom:
It's the `nano_dy` sample. It's crazy slow when `NanoEventsFactory` is in the way. With just `uproot.dask` it's fast

lgray:
Hmm it's not that slow in eager mode either. I'll take a closer look tomorrow, this seems to be some funny code path.

@jpivarski @ioanaif please don't merge this until we get decent performance out of it.

lgray:
If you can run a profile (with pyinstrument) to see where it's spending all its time, it'll help.

iasonkrom:
Letting it run and I'll send the output html file over.

lgray:
no, not really. I'll take a look tomorrow then.

lgray:
@iasonkrom I found the fix! `form.select_columns` seems to have quadratic or worse scaling. Now I just check if the list of request columns is the same as the full set of columns and use the un-projected form.

iasonkrom:
image

Yup, looks good!

lgray:
@jpivarski @ioanaif this is ready to merge

jpivarski (approved):
The changes look good to me and you've verified that it's a performance improvement, so I'll merge it. Thanks!

Issue #906, opened 2023-06-20
Exception on accessing `.interpretation` member for some non-numeric branches


nikoladze:
For some branches i get an Exception when `.interpretation` is accessed. Noticed this when working on the coffea PHYSLITE interpretation (the NanoEvents factory tries to check whether a branch is readable in uproot by looking at the interpretation). It also happens when i call `tree.show()`.

```python
>>> import uproot
>>> uproot.__version__
'5.0.8'
```

Example:

```python
from skhep_testdata import data_path
import uproot
filename = data_path("uproot-issue-798.root") # PHYSLITE example file
f = uproot.open(filename)
tree = f["CollectionTree"]

>>> tree["EventInfo"].interpretation
Traceback (most recent call last):
  File "/home/nikolai/python/uproot4/src/uproot/interpretation/identify.py", line 383, in interpretation_of
    from_dtype = _leaf_to_dtype(leaf, getdims=False).newbyteorder(">")
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nikolai/python/uproot4/src/uproot/interpretation/identify.py", line 102, in _leaf_to_dtype
    return numpy.dtype((_ftype_to_dtype(leaf.member("fType")), dims))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nikolai/python/uproot4/src/uproot/interpretation/identify.py", line 60, in _ftype_to_dtype
    raise NotNumerical()
uproot.interpretation.identify.NotNumerical

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nikolai/python/uproot4/src/uproot/behaviors/TBranch.py", line 1933, in interpretation
    self._interpretation = uproot.interpretation.identify.interpretation_of(
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nikolai/python/uproot4/src/uproot/interpretation/identify.py", line 460, in interpretation_of
    return out.simplify()
           ^^^^^^^^^^^^^^
  File "/home/nikolai/python/uproot4/src/uproot/interpretation/objects.py", line 450, in simplify
    return self._model.strided_interpretation(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #910, opened 2023-06-29
fix: issues with members when dealing with non-numeric branches issue #906

ioanaif:
This issue was created when adding a way to deal with data having extra offsets. 

For that, members were set to have a (None, None) item before headers were added ([L234](https://github.com/scikit-hep/uproot5/blob/a9f4374607b7434ef89ebec82c784d34bfc31413/src/uproot/streamers.py#L234))

Then, in `objects.py`, the `members` var was cleared of the `(None, None)`s. ([L600](https://github.com/scikit-hep/uproot5/blob/a9f4374607b7434ef89ebec82c784d34bfc31413/src/uproot/interpretation/objects.py#L600))

However, it was only tested in the case where a header is added after the Nones tuple, thus the while loop would stop at the first actual header and continue to clear the members var of Nones. In the file that caught this issue, no headers are appended and the `members` var in the end is equal only to (None, None) which causes this while loop to complain of index out of range. 

*** I used `range` to fix the issue, because of the way the condition is imposed.  If `enumerate` is used, pre-commit will complain of not using `first_value_loc` in the loop body.


jpivarski (approved):
> I used `range` to fix the issue, because of the way the condition is imposed. If `enumerate` is used, pre-commit will complain of not using `first_value_loc` in the loop body.

One code-tidying rule would have in one way ("Use `enumerate` if you use both the index and the value!") and another code-tidying rule would have it the other way ("Use `_` if you don't reference the index in the loop!") I guess the rules are in conflict here because using `break` to intentionally get the first index value that satisfies some condition and use it beyond the loop itself is breaking the abstraction of what a loop is for.



(But it's fine to use it here because this is imperative programming; `break` to find the first index satisfying a condition is a common pattern.)

We talked about this on Zoom and I remember the context; this is the right fix.

I'll merge it now because it looks like a simple PR, and it looks done.

Issue #880, opened 2023-04-17
Something wrong in deserialization of straightforward custom class


jpivarski:
### Discussed in https://github.com/scikit-hep/uproot5/discussions/879

<div type='discussions-op-text'>

<sup>Originally posted by **mattiasoldani** April 17, 2023</sup>
Howdy!

I'm using uproot 4.3.7 to open a ROOT file (of which a sample is attached: [TEST230417_LOCAL_SMALL.zip](https://github.com/scikit-hep/uproot5/files/11249405/TEST230417_LOCAL_SMALL.zip)) and access a tree ("Z") that is inside it. One of the branches of this tree, "Cluster[6]", should be an array of 6 custom zCluster objects for each event. Each zCluster object includes several members, mostly float and TVector3 entries. I checked with ROOT and all the branch dimensions are as expected, as seen in the TBrowser screenshots below: the tree contains 116 events, whereas all the Cluster[6] numerical subbranches contain 116*6 = 696 entries.

![Z_Event_iEv](https://user-images.githubusercontent.com/62517661/232466982-a2d77f83-1c39-4ffc-a9e6-be7737ed16e1.jpg)
![Z_Event_Cluster_nHits](https://user-images.githubusercontent.com/62517661/232466977-79dfddea-fc19-4205-aaf6-ebdcabb56498.jpg)

If I try to open the same file in uproot via
```
file = uproot.open("TEST230417_LOCAL_SMALL.root")
tree = file["Z"]
```
I get the correct tree, i.e. `tree.num_entries` outputs 116 and `tree.show()` outputs
```
name                 | typename                 | interpretation                
---------------------+--------------------------+-------------------------------
Event                | zEvent                   | AsGroup(<TBranchElement 'Ev...
Event/TObject        | (group of fUniqueID:u... | AsGroup(<TBranchElement 'TO...
Event/TObject/fUn... | uint32_t                 | AsDtype('>u4')
Event/TObject/fBits  | uint32_t                 | AsDtype('>u4')
Event/iEv            | int32_t                  | AsDtype('>i4')
Event/W              | double                   | AsDtype('>f8')
Event/pK             | TLorentzVector           | AsStridedObjects(Model_TLor...
Event/xK             | TVector3                 | AsStridedObjects(Model_TVec...
Event/iPair          | int32_t                  | AsDtype('>i4')
Event/ig[6]          | int32_t[6]               | AsDtype("('>i4', (6,))")
[...]

PR #911, opened 2023-06-30
fix: interpretation for arrays of non-numerical objects issue 880 part2

ioanaif:
This PR addresses the second part of issue #880, where only one of the 6 Clusters were present into the final array. This was due to the interpretation being `AsObjects(Model_zCluster)` instead of `AsObjects(AsArray(False, False, Model_zCluster, (6,)))`.


src/uproot/interpretation/identify.py
```diff
@@ -480,15 +480,23 @@ def interpretation_of(branch, context, simplify=True):
...
                     model_cls = uproot.containers.AsArray(False, False, model_cls, dims)
                 else:
                     if hasattr(model_cls, "header"):
                         model_cls._header = False
                     model_cls = uproot.containers.AsArray(True, False, model_cls, dims)
 
-            out = uproot.interpretation.objects.AsObjects(model_cls, branch)
+            if dims != () and not branch._streamer_isTClonesArray:
+                out = uproot.interpretation.objects.AsObjects(model_cls, branch)
+                out._forth = False
```
jpivarski, line 496: Right, slow functionality is better than no functionality.

jpivarski (approved):
Looks great!

For posterity, @ioanaif and I talked about this on Zoom. Adding a new AwkwardForth implementation for a previously unhandled operation would be a big project. The first step toward fixing #880 is to make it work at all, with a Python (not AwkwardForth) implementation. This is a good solution to the current problem.

jpivarski (approved):
I had two review requests. I didn't know that could happen.

@ioanaif, merge when you're ready!

PR #912, opened 2023-07-03
fix: pandas and double nested vectors issue 885

ioanaif:
Double nested ROOT arrays when opened with pandas their type remained to be an `STLVector`. This happened because `AwkwardForth` was not being invoked for `pandas` and the processing result was a numpy array which was then added to a `series` without any other transformations.


jpivarski (changes requested):
I think this is generally the right solution: there are `array` types that `uproot.interpretation.library.Pandas.finalize` should handle, but doesn't.

However, passing any `array` through `ak.to_numpy(array)` is too all-inclusive. Also, using any exception in that (and also the `pandas.Series` constructor) to trigger awkward-pandas is also too broad. There are details below: the safest way to fix this particular bug is to catch the previously-unhandled type of `array` (is it `ObjectArray`?) and prepare that the same way Awkward-handling code does, then pass it to awkward-pandas.

(I have a suggested change below, but the more I think about it, that suggested change is wrong: it assumes that the previously-unhandled type of `array` is Awkward, but I think it might be `ObjectArray`.)

jpivarski (approved):
Perfect! Thank you!

I think it's done and can be merged, but I'll let you do that in case you have anything to add at the last minute.

Issue #443, opened 2021-09-18
Reading files from S3 with authenticated http headers (s3http, s3https)


wdconinc:
The Electron Ion Collider (EIC) is storing large data set on storage servers that we access via the S3 protocol. Reading from these files is currently not supported by uproot since dedicated [signed authentication headers](https://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html) are required.

ROOT has support for the S3 (and Google Storage) protocol in [TS3WebFile](https://root.cern.ch/doc/master/TS3WebFile_8cxx_source.html) and [TS3HTTPRequest](https://root.cern.ch/doc/master/TS3HTTPRequest_8cxx_source.html). This allows for specifying files with s3http:// and s3https:// protocols (as well as gshttp:// and gshttps://).

I am filing this issue as way to keep track of discussion, suggestions, and other interested parties. I will be opening a WIP PR when we have some code that others may want to comment on. Currently we are anticipating that we will be able to implement this as a modification of `src/uproot/source/http.py` rather than by adding an entirely new physical layer.

wdconinc:
The user-facing interface strategy would be to support `s3http://` and `s3https://` URLs with specification of the S3 credentials in one of the following methods:
- environment variables `S3_ACCESS_KEY` and `S3_SECRET_KEY` (as done by TS3WebFile),
- syntax `s3http://S3_ACCESS_KEY:S3_SECRET_KEY@example.com` to emulate the `AUTH=S3_ACCESS_KEY:S3_SECRET_KEY` optional argument in TS3WebFile (this mirrors the approach for basic authentication for https connections).

jpivarski:
#692 would likely be the best way forward. Hopefully, we can find someone to take this up again (as I understand it, @nsmith- is very busy!). I can help with ensuring that it fits the `Source` protocol (connection to Uproot).

PR #916, opened 2023-07-05
feat: Support reading from S3

veprbl:
This introduces support for for accessing files from S3. The `minio` client is used to interact with S3 service provider to obtain a presigned http URL that is valid for 7 days. The S3 method for getting objects is not used for simplicity, but also there is an assumption that the HTTP multipart queries might have a better performance.

With regards to choice of URI format, the `s3://<bucket>/<object>` schema as supported by the `aws` tool is implemented with some extensions to allow for non-AWS endpoints. Officially, there are also supposed to be another two formats[1]:

 - *virtual-hosted-style* `https://<bucket>.s3.<region>.amazonaws.com/<object>`
 - *path-style* `https://s3.<region>.amazonaws.com/<bucket>/<object>` which is "deprecated"

For these urls is not clear how to differentiate between the standard HTTPS and the S3 in the case when non-AWS endpoint is used.
It is worth noting that, in ROOT, `TS3WebFile` introduces some *path-style* schemas that don't fit any of the known tools.

[1] https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html

### TODO:
 - [x] update documentation
 - [x] add tests (need permanent public instance)

Resolves: #443

cc @wdconinc @sly2j

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,8 @@ The following libraries are also useful in conjunction with Uproot, but are not 
 
 **For accessing remote files:**
 
-   * `xrootd`: if reading files with `root://` URLs.
+   * `minio`: if reading files with `s3://` URIs.
+   * `xrootd`: if reading files with `root://` URIs.
    * HTTP/S access is built in (Python standard library).
 
 **For distributed computing with [Dask](https://www.dask.org/):**
--- a/docs-sphinx/basic.rst
+++ b/docs-sphinx/basic.rst
@@ -18,7 +18,7 @@ The :doc:`uproot.reading.open` function can be (and usually should be) used like
     >>> with uproot.open("path/to/dataset.root") as file:
     ...     do_something...
 
-to automatically close the file after leaving the ``with`` block. The path-name argument can be a local file (as above), a URL ("``http://``" or "``https://``"), or XRootD ("``root://``") if you have the `Python interface to XRootD <https://anaconda.org/conda-forge/xrootd>`__ installed. It can also be a Python file-like object with ``read`` and ``seek`` methods, but such objects can't be read in parallel.
+to automatically close the file after leaving the ``with`` block. The path-name argument can be a local file (as above), a URL ("``http://``" or "``https://``"), S3 ("``s3://``) or XRootD ("``root://``") if you have the `Python interface to XRootD <https://anaconda.org/conda-forge/xrootd>`__ installed. It can also be a Python file-like object with ``read`` and ``seek`` methods, but such objects can't be read in parallel.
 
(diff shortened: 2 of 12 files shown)


jpivarski:
I should explicitly let you know that I've been keeping this PR in my email inbox because I think it's important, but I just realized that I haven't said anything about it yet.

(I'm in a talk about open source practices and I just guiltily remembered that this is still open.)

jpivarski (approved):
Worse still, for my justification, is that this is an easy PR to review. I just read through all of the changes and it looks great. I can see that it's primarily a URI translation: the `s3://` URI scheme can be rewritten as an HTTPS scheme, and then the `HTTPSource` can take over.

I thought there was an issue with getting arguments from `uproot.open` into the `S3Source` constructor, but then I answered my own question: it goes through `**self._options`:

https://github.com/scikit-hep/uproot5/blob/a8644df7a96ee2d788adb1e35f326ba160bedf48/src/uproot/reading.py#L578-L583

Have you tested that? (With both `uproot.open` and another, like `uproot.iterate`?)

It can be hard to write a permanent CI test for the testing suite without having a file that won't ever be deleted or moved—worse still if it requires AWS secrets to authenticate. Would this work?

```
s3://pivarski-princeton/chep-2021-jagged-jagged-jagged/zlib9-rntuple-jagged0.root
```

It's publicly accessible and it won't be going away because it's used in so many big-data tests. You'd have to set `entry_stop=100` if you try to read from a TTree, though I don't see a need to do that (accessing just the metadata object should be enough to show that S3 is working). Also, I own it, and I'm easy to get in touch with if it ever does go away.

I should also let you know that we plan to revive #692, which should eventually run all backends through fsspec, and that would include S3. But it's better to add this feature now and have it be useful before the long-term plan becomes a reality.

wdconinc:
Another approach to setting up a CI test for S3 is to spin up a Minio server as part of the job, write an S3 file to it with their client, then connect to it with uproot. Something like https://github.com/fclairamb/docker-minio-github-actions.

ioanaif:
Hi! 
This looks ready to go, it only misses testing. Once the tests are added (and they pass) we can merge this.

*The S3 files will be persist so the test suite will be reliable

nsmith-:
Can I point out that if we were to take fsspec dependency as in #692 we would have S3 for free? That said, it was not finished.

jpivarski:
Thanks, @nsmith-, for clarifying. It's what I meant by

> I should also let you know that we plan to revive #692, which should eventually run all backends through fsspec, and that would include S3. But it's better to add this feature now and have it be useful before the long-term plan becomes a reality.

At that point in the future, we would roll back this explicit implementation (and probably release a new minor version of Uproot). But this will be useful in the interim.

nsmith-:
Apologies! I failed to read your whole comment. Indeed this is useful in interim.

jpivarski (approved):
This looks good to me, and it has a test (both for an existing and a non-existing file), so I'd merge it.

Are you ready? It looks like you are, since there hasn't been any recent activity. Just respond with an "ok" and I'll merge it!

veprbl:
Yes. I think this is ready.

PR #927, opened 2023-08-07
fix: don't assume Uproot is in global scope in TPython::Eval

jpivarski:


jpivarski:
Let me know, @mramospe and @franciscosili, if this PR fixes the issue in your context. As you can see above, I wasn't able to reproduce it.

agoose77:
I can't reproduce with your test, but this one works for me
```python
# BSD 3-Clause License; see https://github.com/scikit-hep/uproot5/blob/main/LICENSE


import ROOT


def test_import():
    h = ROOT.TProfile3D()
    __import__("uproot").from_pyroot(h)
```

jpivarski:
I verified that TProfile3D raises a segfault in `main` (whereas TH1F does not), but it's fixed in this PR. I wonder what's special about TProfile3D—maybe there's a new class version for it? Or maybe we don't have predefined streamers for it?

https://github.com/scikit-hep/uproot5/blob/a8644df7a96ee2d788adb1e35f326ba160bedf48/src/uproot/behaviors/TH3.py#L4-L5

Well, this does verify that there's a segfault without the fix and no segfault with the fix, so I'll ask @ioanaif to review this PR and merge it if everything's good. (I'm not planning on making any more changes.)

ioanaif (approved):
LGTM!

PR #930, opened 2023-08-10
fix: expressions failing in pandas issue 922

### LINKED ISSUE #922: "where" in expression fails with library "pandas"
## Short version
uproot version: 5.0.10
Using the keyword `where` in an expression when loading with `library="pd"` fails with
```
AttributeError: 'numpy.ndarray' object has no attribute 'to_frame'
```
(full stacktrace below)

## longer version
When loading a root file, using `where` in the expression fails for pandas (it should be supported [AFAI can see in PythonLanguage](https://github.com/scikit-hep/uproot5/blob/a8644df7a96ee2d788adb1e35f326ba160bedf48/src/uproot/language/python.py#L318). It seems to be only with that expression (I've just noticed that other ternary operators like min, max are not (yet?) supported, so maybe this is the reason?)
I've tried to find the core issue, but didn't succeed.

Code to reproduce (**creates file in current folder**):
```python
# create tmp file 
with uproot.recreate("tmp.root") as file:
     file['tree'] = {'b1': [1, 5, 9], 'b2': [3, 6, 11]}

# producing the bug
with uproot.open("tmp.root") as file:
     tree = file['tree']
	# different options
     tree.arrays(['log(b1)'], library='pd')  # arbitrary expressions work, like log
     tree.arrays(['where(b1 < b2, b1, b2)'], library='np')  # works with np
     tree.arrays(['where(b1 < b2, b1, b2)'], library='pd')  # fails with pd
     
```

Full stacktrace:

<details>
<summary>Details</summary>

```
AttributeError                            Traceback (most recent call last)
Cell In[8], line 3
      1 with uproot.open("tmp.root") as file:
      2     tree = file['tree']
----> 3     print(tree.arrays(['where(b1 < b2, b1, b2)'], library='pd'))

File ~/mambaforge/envs/zfit311/lib/python3.11/site-packages/uproot/behaviors/TBranch.py:898, in HasBranches.arrays(self, expressions, cut, filter_name, filter_typename, filter_branch, aliases, language, entry_start, entry_stop, decompression_executor, interpretation_executor, array_cache, library, ak_add_doc, how)
    891 del arrays
    893 expression_context = [
    894     (e, c) for e, c in expression_context if c["is_primary"] and not c["is_cut"]
    895 ]
    897 return _ak_add_doc(
--> 898     library.group(output, expression_context, how), self, ak_add_doc
    899 )

File ~/mambaforge/envs/zfit311/lib/python3.11/site-packages/uproot/interpretation/library.py:867, in Pandas.group(self, arrays, expression_context, how)
    865 elif uproot._util.isstr(how) or how is None:
    866     arrays, names = _pandas_only_series(pandas, arrays, expression_context)
--> 867     return _pandas_memory_efficient(pandas, arrays, names)
    869 else:
    870     raise TypeError(
    871         "for library {}, how must be tuple, list, dict, str (for "
    872         "pandas.merge's 'how' parameter, or None (for one or more"
    873         "DataFrames without merging)".format(self.name)
    874     )

File ~/mambaforge/envs/zfit311/lib/python3.11/site-packages/uproot/interpretation/library.py:793, in _pandas_memory_efficient(pandas, series, names)
    791 for name in names:
    792     if out is None:
--> 793         out = series[name].to_frame(name=name)
    794     else:
    795         out[name] = series[name]

AttributeError: 'numpy.ndarray' object has no attribute 'to_frame'
```

</details>

ioanaif:
This PR fixes #922 

The issue was caused by an assumption that in the code path that series are made of other series which are then to be transformed in dataframes.


jpivarski (approved):
This is a better way of dealing with it than what I had in mind: modifying all of the functions (like `where`) so that `Series` inputs become `Series` outputs. This gets them all in one spot.

PR #934, opened 2023-08-12
chore: skip pandas test if pandas is not installed

natsukium:
The test introduced in #930 depends on pandas.
Like other tests, it should be skipped if pandas cannot be imported.


jpivarski (approved):
Absolutely; thanks for fixing this!

natsukium:
Thanks for your review!

PR #941, opened 2023-08-17
fix: clamp start and stop

agoose77:
Fixes case where the number of entries is less than the entry start in Dask.

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -889,15 +889,11 @@ class _UprootOpenAndRead:
             self.real_options,
         )
         num_entries = ttree.num_entries
-        start, stop = istep_or_start, nsteps_or_stop
-        if not ischunk:
-            events_per_step = math.ceil(num_entries / nsteps_or_stop)
-            start, stop = (istep_or_start * events_per_step), min(
-                (istep_or_start + 1) * events_per_step, num_entries
-            )
-        elif (not 0 <= start < num_entries) or (not 0 <= stop <= num_entries):
-            raise ValueError(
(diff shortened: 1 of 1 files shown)


src/uproot/_dask.py
```diff
@@ -889,15 +889,11 @@ def __call__(self, file_path_object_path_istep_nsteps_ischunk):
...
-                (istep_or_start + 1) * events_per_step, num_entries
-            )
-        elif (not 0 <= start < num_entries) or (not 0 <= stop <= num_entries):
-            raise ValueError(
-                f"""explicit entry start ({start}) or stop ({stop}) from uproot.dask 'files' argument is out of bounds for file
+        if ischunk:
+            start, stop = istep_or_start, nsteps_or_stop
+            if (not 0 <= start < num_entries) or (not 0 <= stop <= num_entries):
+                raise ValueError(
+                    f"""explicit entry start ({start}) or stop ({stop}) from uproot.dask 'files' argument is out of bounds for file
```
agoose77, line 896: This is mainly just a style change

src/uproot/_dask.py
```diff
@@ -906,8 +902,14 @@ def __call__(self, file_path_object_path_istep_nsteps_ischunk):
     {ttree.object_path}
 
 which has {num_entries} entries"""
+                )
+        else:
+            events_per_step = math.ceil(num_entries / nsteps_or_stop)
```
agoose77, line 907: Here I added a `min` to ensure that we can't set `start` to `>num_entries`

jpivarski: `start` can be equal to `num_entries`, and if so, the range will have length zero as you say.

jpivarski (approved):

PR #942, opened 2023-08-17
test: better test for vectorVectorDouble

jpivarski:
It wasn't testing `library="ak"` before. Also, it's a small enough file that we can test the whole thing.


ioanaif (approved):
LGTM!

PR #956, opened 2023-09-13
fix: add minimimal emscripten support via non-HTTP sources

agoose77:
#868 fixed some import errors, but we still used threads for non-HTTP sources. In this PR, we close the gap and make it possible to disable threads for all sources. By default, this is done for emscripten platforms.

I didn't add HTTP support here — it's possible with e.g. `XMLHttpRequest`, but we'll need to spend some more cycles on it. Crucially, CORS means that this can be fiddly, but it would be more performant for emscripten platforms where `mmap` is not supported.

See https://agoose77.github.io/pyhep-23-jupyterlite/lab/index.html for a repo that uses this patch.


ioanaif:
Nice! 
For me the file opening part is still not working in the demo. However, I am not sure if it is a problem caused by my browser (firefox - happened in the past for CoDaS)

agoose77:
@ioanaif I use Firefox too! Does it give a useful traceback e.g. `FileNotFoundError`? If so, it might help to force-refresh the page.

ioanaif:
Yes. Last time this happened for my browser because the file was too large

src/uproot/source/http.py
```diff
@@ -298,25 +299,24 @@ def multifuture(source, ranges, futures, results):
         )
 
         def task(resource):
+            nonlocal connection
```
jpivarski, line 302: That's right: `connection[0]` was a Python2ism. `global` is not the right scope, and `nonlocal` didn't exist.

jpivarski (approved):
The new argument, `no_threads`, overrides `num_*workers` and there should be some note in the documentation to indicate this. (It's always a little more confusing when one argument changes what another does.)

Here are all of the places where `num_*workers` arguments are defined:

```bash
% fgrep -r 'workers (int' src/uproot/ --include='*.py'
src/uproot/reading.py:    * num_workers (int; 1)
src/uproot/reading.py:    * num_fallback_workers (int; 10)
src/uproot/reading.py:    * num_workers (int; 1)
src/uproot/reading.py:    * num_fallback_workers (int; 10)
src/uproot/_dask.py:    * num_workers (int; 1)
src/uproot/_dask.py:    * num_fallback_workers (int; 10)
src/uproot/behaviors/TBranch.py:    * num_workers (int; 1)
src/uproot/behaviors/TBranch.py:    * num_fallback_workers (int; 10)
src/uproot/behaviors/TBranch.py:    * num_workers (int; 1)
src/uproot/behaviors/TBranch.py:    * num_fallback_workers (int; 10)
```

Actually, I don't think `no_threads` is documented at all, including the fact that its default depends on `sys.platform`.

src/uproot/source/http.py
agoose77: Hmm, you're right. At one point I was making this an implicit thing, but there's sense in exposing it.

jpivarski:
> > Actually, I don't think no_threads is documented at all, including the fact that its default depends on sys.platform.
> 
> Hmm, you're right. At one point I was making this an implicit thing, but there's sense in exposing it.

If `no_threads` is in the list of `**options`, then it's already exposed. A user could override that manually.

agoose77:
Let me hold this until I finish some last changes :)

agoose77:
@ioanaif and @jpivarski, what do you think about testing this PR? We could just parameterise the `use_threads` argument, but that will more than double our test time.

jpivarski:
We could double our test time on a subset of tests. Pick one that seems to be focusing on backend features, one that tests HTTP and XRootD with various options, non-existent files, etc., and double those with `use_threads=True/False`.

agoose77:
I've updated the tests. The documentation is probably not perfect; we talk about the executor being a threaded executor, but sometimes it's a trivial executor. I'm happy to not worry about that for now, as I think we should drop our executor abstraction in the near future in favour of the existing solution in Python 3's `concurrent.futures.`

src/uproot/_dask.py
```diff
@@ -127,6 +127,7 @@ def dask(
     * timeout (float for HTTP, int for XRootD; 30)
     * max_num_elements (None or int; None)
     * num_workers (int; 1)
+    * use_threads (bool; False if emscripten, else True)
```
jpivarski, line 130:
```suggestion
    * use_threads (bool; False on the emscripten platform (i.e. in a web browser), else True)
```

src/uproot/behaviors/TBranch.py
```diff
@@ -160,6 +160,7 @@ def iterate(
     * timeout (float for HTTP, int for XRootD; 30)
     * max_num_elements (None or int; None)
     * num_workers (int; 1)
+    * use_threads (bool; False if emscripten, else True)
```
jpivarski, line 163:
```suggestion
    * use_threads (bool; False on the emscripten platform (i.e. in a web browser), else True)
```

src/uproot/behaviors/TBranch.py
```diff
@@ -332,6 +333,7 @@ def concatenate(
...
```
jpivarski, line 336:
```suggestion
    * use_threads (bool; False on the emscripten platform (i.e. in a web browser), else True)
```

src/uproot/reading.py
```diff
@@ -85,6 +85,7 @@ def open(
     * timeout (float for HTTP, int for XRootD; 30)
     * max_num_elements (None or int; None)
     * num_workers (int; 1)
+    * use_threads (bool; False if emscripten, else True)
```
jpivarski, line 88:
```suggestion
    * use_threads (bool; False on the emscripten platform (i.e. in a web browser), else True)
```

src/uproot/reading.py
```diff
@@ -541,6 +542,7 @@ class ReadOnlyFile(CommonFileMethods):
...
```
jpivarski, line 545:
```suggestion
    * use_threads (bool; False on the emscripten platform (i.e. in a web browser), else True)
```

jpivarski (approved):
I'm only making a documentation suggestion (for users who haven't heard of emscripten).

PR #957, opened 2023-09-14
fix: tests were failing because dask_awkward.lib.testutils needs pyarrow

jpivarski:
This is an immediate fix, but when dask-contrib/dask-awkward#365 makes it into a release, we should remove the line

```python
pytest.importorskip("pyarrow")  # dask_awkward.lib.testutils needs pyarrow
```

from all of the following files:

* test_0652_dask-for-awkward.py
* test_0876-uproot-dask-blind-steps.py
* test_0700-dask-empty-arrays.py
* test_0755-dask-awkward-column-projection.py


jpivarski:
@ioanaif, I'm going to fast-track this (self-approve and merge it) because it's needed to get our tests to run.

PR #960, opened 2023-09-18
fix: adapt to new shape touching in awkward/dask-awkward

lgray:
Following the conventions in https://github.com/dask-contrib/dask-awkward/pull/368

@agoose77 @douglasdavis


lgray:
I think this is ~about right but I'm probably missing something

agoose77:
Closed in favour of #966

PR #962, opened 2023-09-20
feat: add support for current RNTuple files

ioanaif:


jpivarski (approved):
We talked about this on Zoom; it fixes the RNTuple metadata change and also adds split decoding.

It all looks good and I'll be merging it after the tests pass.

PR #965, opened 2023-10-03
fix: inverted axes for variances of 2D weighted histograms when transformed to hist

ioanaif:


ioanaif:
The [_values_variances](https://github.com/scikit-hep/uproot5/blob/0096168b7a4f2d7d66ee73e67310364a9f6d5320/src/uproot/behaviors/TH1.py#L318) method can be used for the computation in `to_boost`. However, the computation there seems to me to be wrong as well, consider the following test: 


```
>>> import numpy
>>> import uproot
>>> import ROOT
>>> hroot2 = ROOT.TH2F("hroot2", "", 2, 0, 1, 3, 0, 1)
>>> hroot2.Sumw2()
>>> a = [0.12, 0.32, 0.1, 0.3, 0.4,0.7, 0.12, 0.12, 0.32, 0.1, 0.3, 0.4,0.7, 0.12]
>>> b = [0.22, 0.121, 0.8, 0.3, 0.4,0.7, 0.23, 0.12, 0.32, 0.1, 0.3, 0.4,0.7, 0.12]
>>> c = [23, 19, 17, 11 , 7, 5, 3,23, 19, 17, 11 , 7, 5, 3]
>>> for i in range(14):
...   hroot2.Fill(a[i], b[i], c[i])
... 
>>> huproot2 = uproot.from_pyroot(hroot2)
>>> values = huproot2.values(flow=True)
>>> variances = numpy.zeros(values.shape, dtype=numpy.float64)
>>> sumw2 = huproot2.member("fSumw2", none_if_missing=True)
>>> sumw2 = numpy.asarray(sumw2, dtype=sumw2.dtype.newbyteorder("="))
>>> sumw2 = numpy.reshape(sumw2, values.shape)
>>> positive = sumw2 > 0
>>> variances[positive] = sumw2[positive]
>>> (variances == huproot2.variances(flow=True)).all()
False
>>> variances
array([[   0.,    0.,    0.,    0.,    0.],
       [2329.,    0.,    0.,    0.,   98.],
       [   0.,    0.,    0.,  289.,   50.],
       [   0.,    0.,    0.,    0.,    0.]])
>>> huproot2.variances(flow=True)
array([[   0.,    0.,    0.,    0.,    0.],
       [   0., 2329.,   98.,  289.,    0.],
       [   0.,    0.,    0.,   50.,    0.],
       [   0.,    0.,    0.,    0.,    0.]])
```

ioanaif:
In TH2, `_values_variances` does perform the axis swap, thus all is good in the computation of the variances.

I removed both the previous computation and my fix to the computation from `to_boost` and replaced it with simply calling `.variances(flow=True)` since variances are already computed at that stage. 

This should fix any problem with the variances' computation and remove duplication for easier debug in the future. 

It's ready to go imo. ☺️

jpivarski (approved):
Okay, great!

I saw that you're using `self.variances`, which runs `self._values_variances`, and so the values are being computed twice. (That's the original reason why this function reimplemented it.)

But we can arrange to have `values` computed only once by doing `self._values_variances` in the if-clause and `self.values` in the else-clause. In both cases, `flow=True`.

I think this is ready to go, but since I made an edit, I'll give you the chance to counter-edit before merging. Go ahead and merge when you think it's ready.

PR #967, opened 2023-10-03
feat: adding a very basic FSSpecSource

lobis:
This PR simply adds the changes from https://github.com/scikit-hep/uproot5/pull/692 with the necessary updates to make it compatible with current uproot.

A very basic `Source` is implemented for `fsspec` allowing the user to optionally use `fsspec` by setting the appropiate handler in the read operation.

```
with uproot.open(
    "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root",
    http_handler=uproot.source.fsspec.FSSpecSource,
) as f:
    data = f["Events/MET_pt"].array(library="np")
    assert len(data) == 40
```


jpivarski:
Great to see this starting! (We'll talk on Zoom in an hour.) I'm watching it, but I removed myself as a reviewer for now because you can use the "request review" button as a way of reminding me about it when it's done. I don't think I get emails about a PR going out of draft mode, but I do get emails about requests-to-review, so we use that as a last step in our process.

pyproject.toml
```diff
@@ -57,6 +57,8 @@ dev = [
     "awkward-pandas;python_version >= \"3.8\"",
 ]
 test = [
+    "aiohttp",
+    "fsspec-xrootd",
```
lobis, line 61: Should we keep only `fsspect-xrootd` or should we also add `fsspect` explicitly? `fsspec` is included inside `fsspect-xrootd`. Just wondering what would be best practise.

src/uproot/source/fsspec.py
```diff
@@ -0,0 +1,139 @@
...
+                "xrootd_handler",
+                "http_handler",
+                "s3_handler",
+                "object_handler",
+                "max_num_elements",
+                "num_workers",
+                "num_fallback_workers",
+                "use_threads",
+                "begin_chunk_size",
+                "minimal_ttree_metadata",
```
lobis, line 36: I think this is the best place to remove the uproot-specific options (addressing comment in code above).

The CI was failling because uproot-specific options were being passed into which should have been removed due to the list of options being updated after this was written.

I guess it would be a good idea to avoid this hardcoded list of uproot options. I don't think it's easy to do as of right now. Looks like the list of valid options is maintained on `reading.py`, if it doesn't exist already perhaps we could add a validation of options (enforce user didn't use an invalid keyword) from a list of valid options and import this list to remove uproot-specific options?

pyproject.toml
agoose77: The usual practice is - if you import anything from package X, declare it in your dependencies, even if package Y requires X. Otherwise, package Y could change its dependency in future, and your package would break! :)

src/uproot/source/fsspec.py
lobis: Update: load options dictionary from `uproot.reading.open.defaults`

lobis:
@jpivarski @nsmith- I think this is ready to go.

Currently the `fsspec` source works in place of the http, file, s3 and xrootd sources. At this point fsspec is only present as a test dependency and this can be merged without causing any problems as it's fully optional.

One of my next steps (for a dedicated PR) would be to refactor the `reading.open` options for handlers (`http_handler`, `file_handler`, etc.) in order to use a single handler option (`handler`) that can be set to the currenty existing sources or the new fsspec source with the goal of setting the fsspec source as the default once the integration is complete.

I think this would introduce some limitations such as the ability to open some files with one protocol and some with another in the same function call, but as we discussed this is an acceptable compromise. The ability to use the previous sources (`http`, `s3`, etc.) will remain and the corresponding source will be assigned from parsing the uri string.

tests/test_0692_fsspec.py
```diff
@@ -0,0 +1,75 @@
...
+
+import pytest
+
+import uproot
+import uproot.source.fsspec
+
+
+def validate_coffea_file_0(file):
+    data = file["Events/MET_pt"].array(library="np")
+    assert len(data) == 40
```
jpivarski, line 11: In principle, performing repetitive actions in a helper function would be good practice, but it does make errors in the pytest output harder to read.

You've probably noticed that the other tests do a lot of copy-paste. We have different standards for the testing code: it doesn't need to be as nice and maintainable as the package code—most of it is never changed after first being written (which is why we name test files after PR numbers, as I see you've noticed).

src/uproot/source/fsspec.py
```diff
@@ -0,0 +1,126 @@
...
+        data = self._fs.cat_ranges(
+            [self._file_path] * len(ranges),
+            [start for start, _ in ranges],
+            [stop for _, stop in ranges],
+        )
+        chunks = []
+        for item, (start, stop) in zip(data, ranges):
+            future = uproot.source.futures.TrivialFuture(item)
+            chunk = uproot.source.chunk.Chunk(self, start, stop, future)
+            uproot.source.chunk.notifier(chunk, notifications)()
```
jpivarski, line 109: It's a blocking call in this first implementatation, but will be made asynchronous later.

jpivarski (approved):
This looks great! It adds a new class (which doesn't disturb any existing code yet) and it's following all of our local conventions. I think it's ready to merge.

tests/test_0692_fsspec.py
lobis: Makes sense!

tests/test_0692_fsspec.py
```diff
@@ -0,0 +1,73 @@
...
+    ) as f:
+        data = f["Events/MET_pt"].array(library="np")
+        assert len(data) == 40
+
+
+@pytest.mark.network
+def test_open_fsspec_local(tmp_path):
+    url = "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root"
+
+    # download file to local
```
nsmith-, line 34: Could this one be done with scikit-hep/testdata package instead?

pyproject.toml
nsmith-: This was discussed a bit in the last PR https://github.com/scikit-hep/uproot5/pull/692#issuecomment-1227709118 basically to avoid having an unnecessarily long chain of `install this` errors when using for the first time.

tests/test_0692_fsspec.py
lobis: https://github.com/scikit-hep/scikit-hep-testdata/pull/130

PR #971, opened 2023-10-04
feat: Use a single `handler` argument on `uproot.reading.open`

lobis:
Currently the `uproot.reading.open` method takes optional arguments such as `http_handler`, `file_handler` etc. instead of a single argument `handler`. This allows to declare multiple handlers for different file schemas.

This PR falls under the umbrella of the `fsspec` integration (issue incoming). The goal of this integration would be that there is a single source `fsspec` that can handle all schemas. Optionally the user is able to define a custom handler. Existing handlers should work as before (may be deprecated in the future once fsspec integration is validated).

A good first step would be to refactor the current code to only allow a single `handler` argument and if not defined automatically infer this from the file path string.


jpivarski:
This is something that would disrupt current use of Uproot (or provide two ways of doing something? Would `handler` take precedence over `*_handler`?). Therefore, we might need to schedule when it gets merged to avoid the upcoming 5.1.0 release.

lobis:
> This is something that would disrupt current use of Uproot (or provide two ways of doing something? Would `handler` take precedence over `*_handler`?). Therefore, we might need to schedule when it gets merged to avoid the upcoming 5.1.0 release.

Yes this would be somewhat disruptive so it needs to be merged carefully (update docs, etc.). I was planning on removing `*_handler` option entirely and replace it with `handler` (which in practise would never be overriden by the user as the `fsspec` handler should handle everything). Maybe this PR can keep the old `*_handler* options available but the new `handler` takes precendence in a way that it mantains backwards compatibility, I'll think about it.

lobis:
Not sure if this has been spotted yet: I saw the CI fail when accessing the `s3` bucket (https://github.com/scikit-hep/uproot5/actions/runs/6426306777/job/17450310571?pr=971#step:7:5667) returning a 503 code (Service Unavailable). Shouldn't S3 be able to handle virtually any load?

lobis:
This PR is ready for review. The deprecation warnings make the `file_path_to_source_class`method complicated to read but it basically does:

- If specified use the handler in the corresponding `*_handler`option as before. These options are now `None` by default.
- Use the new `handler` otherwise.

In https://github.com/scikit-hep/uproot5/pull/984 I remove these for good and declare the fsspec source as default which should handle all cases (discussions related to this can be done in the PR).

In principle it should be ready to merge but I'm not sure if I should wait for the ci to pass (failing due to unrelated issues).

jpivarski:
That's a good update. This will be ready to squash-and-merge once the tests pass. I'm going to look at one of those to see what's happening.

lobis:
> That's a good update. This will be ready to squash-and-merge once the tests pass. I'm going to look at one of those to see what's happening.

I actually forgot to add that `*_handler` options take preference over `handler`. I'll update and merge. Tests are passing now (server issues I guess).

jpivarski:
The dalitzdata.root is being opened; the `TimeoutErrors` are happening much later, during the read. We can even see some successfully read TBaskets before the socket breaks.

It's only used in two tests,

* `tests/test_0088-read-with-http.py::test_issue176`
* `tests/test_0088-read-with-http.py::test_issue176_again`

...

jpivarski:
On my computer (not CI), I can run these tests in `main`, `main-v510`, and `open-handlers-unify`. It seems like the HTTP server just decided that it doesn't like GitHub Actions anymore.

What happens if you skip these tests (`reason="starterkit.web.cern.ch is breaking HTTP connections in GitHub Actions"`)? Does the rest of the suite run?

jpivarski:
That last commit did not address the starterkit.web.cern.ch server connection failure, but it started working. Maybe the transient issue with the server is over? (If it is, we'll just ignore it for now. Angus's idea of hosting our own in-CI mini-servers could solve it more generally.)

jpivarski:
@nsmith-, I'm in favor of merging this. If you want to give it a look-over before that happens, let me know. Note that this is going into `main-v510`.

PR #973, opened 2023-10-04
test: use file in skhep-testdata for issue #121

lobis:


tests/test_0088-read-with-http.py
```diff
@@ -26,8 +28,6 @@ def test_issue176_again():
 
 @pytest.mark.network
 def test_issue121():
-    with uproot.open(
-        "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root"
-    ) as f:
+    with uproot.open(skhep_testdata.data_path("uproot-issue121.root")) as f:
```
jpivarski, line 31: This one wouldn't count as `@pytest.mark.network` because `scikit-hep_testdata` downloads and caches the file, so Uproot is using the local file system. The reason for the `@pytest.mark.network` is to skip some tests when running on high-latency networks, such as a home computer, because Uproot needs to make several round trips to find seek points within a file. On a home computer, the file that was fetched by `scikit-hep_testdata` is probably already local (we've probably already run this test).

tests/test_0692_fsspec.py
```diff
@@ -5,11 +5,13 @@
...
 
+import skhep_testdata
+
 
 @pytest.mark.network
 def test_open_fsspec_http():
     with uproot.open(
-        "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root",
+        skhep_testdata.data_path("uproot-issue121.root"),
         http_handler=uproot.source.fsspec.FSSpecSource,
```
jpivarski, line 15: Same for this one, and it's not really using the `http_handler`, since it's a local file (as far as Uproot is concerned).

tests/test_0692_fsspec.py
```diff
@@ -29,7 +31,7 @@ def test_open_fsspec_github():
 
 @pytest.mark.network
 def test_open_fsspec_local(tmp_path):
-    url = "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root"
+    url = skhep_testdata.data_path("uproot-issue121.root")
 
     # download file to local
```
jpivarski, line 36: This test may now be redundant because `scikit-hep-testdata` has already downloaded the file.

tests/test_0692_fsspec.py
jpivarski: Maybe what you want instead is

```
https://github.com/scikit-hep/scikit-hep-testdata/raw/v0.4.33/src/skhep_testdata/data/uproot-issue121.root
```

and then it would be HTTP access for Uproot.

tests/test_0692_fsspec.py
```diff
@@ -20,7 +22,7 @@ def test_open_fsspec_http():
 def test_open_fsspec_github():
     pytest.skip("not working yet")
     with uproot.open(
-        "github://CoffeaTeam:coffea@master/tests/samples/nano_dy.root",
+        "github://scikit-hep:scikit-hep-testdata@main/src/skhep_testdata/data/uproot-issue121.root",
```
jpivarski, line 25: Interesting: I didn't know about `github://`. But perhaps you should link to a tag, `v0.4.33`, instead of `main`, which could (in principle) remove this file in the future.

jpivarski (changes requested):
See below!

tests/test_0692_fsspec.py
lobis: I was under the wrong impression that `skhep_testdata.data_path` would just return the full url for the file instead of a local path after downloading the file. Everything should be alright now.

jpivarski (approved):
Is the `github://` test using `api.github.com` with its rate limit? (This is continued from a conversation on Slack.)

If it is, let's just drop the `github://` test. We can assume that fsspec does its job correctly (managing exotic backends like the GitHub one) and just verify that our connection to fsspec is doing what it's supposed to. We should continue testing local files, HTTP, XRootD, and maybe S3 because Uproot users are known to use them, but I'd say that the GitHub one is just "nice to have."

Other than that, I think this PR is ready to merge.

nsmith- (approved):
If the github URL has a rate limit policy that may cause tests to fail then yeah it makes sense to drop it. Otherwise, looks good!

lobis:
> Is the `github://` test using `api.github.com` with its rate limit? (This is continued from a conversation on Slack.)
> 
> If it is, let's just drop the `github://` test. We can assume that fsspec does its job correctly (managing exotic backends like the GitHub one) and just verify that our connection to fsspec is doing what it's supposed to. We should continue testing local files, HTTP, XRootD, and maybe S3 because Uproot users are known to use them, but I'd say that the GitHub one is just "nice to have."
> 
> Other than that, I think this PR is ready to merge.

I think I will keep the pytest skip instead of just removing the test, I think it's nice to have an example of some of the exotic fsspec sources. I think there is some value in testing it explicitly though, for example the github schema url has two `:` in it which made the old "split path and object" helper method fail. I guess we could explicitly test the helper method instead through.

jpivarski:
Okay. If we're ever working on something unrelated and see failures in this test, we'll unceremoniously drop it. You'll notice that some of the `@pytest.mark.skip` reasons are "such-and-such a server is flaky."

Issue #974, opened 2023-10-05
Can't read keys in root files with colon in name


nikoladze:
When i have a key in a root file with a colon in the name, e.g.

```python
>>> import uproot
>>> uproot.__version__
'5.0.12'

f = uproot.recreate("test.root")
f["bla:bla"] = "blub"
```

Then reading this seems to run in some kind of infinite (?) recursion

<details>

```python
In [2]: f["bla:bla"]
---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 f["bla:bla"]

File ~/.local/lib/python3.11/site-packages/uproot/writing/writable.py:968, in WritableDirectory.__getitem__(self, where)
    966 if self._file.sink.closed:
    967     raise ValueError("cannot get data from a closed file")
--> 968 return self._get_del_search(where, True)

File ~/.local/lib/python3.11/site-packages/uproot/writing/writable.py:927, in WritableDirectory._get_del_search(self, where, isget)
    925     else:
    926         last = step
--> 927         step = step[item]
    929 elif isinstance(step, WritableTree):
    930     rest = items[i:]

File ~/.local/lib/python3.11/site-packages/uproot/writing/writable.py:968, in WritableDirectory.__getitem__(self, where)
    966 if self._file.sink.closed:
    967     raise ValueError("cannot get data from a closed file")
--> 968 return self._get_del_search(where, True)

File ~/.local/lib/python3.11/site-packages/uproot/writing/writable.py:927, in WritableDirectory._get_del_search(self, where, isget)
    925     else:
    926         last = step
--> 927         step = step[item]
    929 elif isinstance(step, WritableTree):
    930     rest = items[i:]

    [... skipping similar frames: WritableDirectory.__getitem__ at line 968 (1489 times), WritableDirectory._get_del_search at line 927 (1489 times)]

File ~/.local/lib/python3.11/site-packages/uproot/writing/writable.py:968, in WritableDirectory.__getitem__(self, where)
    966 if self._file.sink.closed:
    967     raise ValueError("cannot get data from a closed file")
--> 968 return self._get_del_search(where, True)

File ~/.local/lib/python3.11/site-packages/uproot/writing/writable.py:927, in WritableDirectory._get_del_search(self, where, isget)
    925     else:
    926         last = step
--> 927         step = step[item]
    929 elif isinstance(step, WritableTree):
    930     rest = items[i:]

[...]

Issue #972, opened 2023-10-04
Integration of `fsspec`


lobis:
This issue is to track the progress of the `fsspec` integration as it's going to be consisting of multiple PRs.

This work is part of an IRIS-HEP Fellowship. Here are some of the PRs performed during this period:

- https://github.com/scikit-hep/uproot5/pull/692 (extensive discussion)
- https://github.com/scikit-hep/uproot5/pull/967
- https://github.com/scikit-hep/uproot5/pull/971
- https://github.com/scikit-hep/uproot5/pull/976
- https://github.com/scikit-hep/uproot5/pull/979
- https://github.com/scikit-hep/uproot5/pull/983
- https://github.com/scikit-hep/uproot5/pull/984 (TODO: remember to merge this in the future when the deprecation warning expires).
- https://github.com/scikit-hep/uproot5/pull/992
- https://github.com/scikit-hep/uproot5/pull/996
- https://github.com/CoffeaTeam/fsspec-xrootd/pull/26
- https://github.com/scikit-hep/uproot5/pull/999
- https://github.com/scikit-hep/uproot5/pull/1001
- https://github.com/scikit-hep/uproot5/pull/1002
- https://github.com/scikit-hep/uproot5/pull/1007
- https://github.com/scikit-hep/uproot5/pull/1009
- https://github.com/scikit-hep/uproot5/pull/1010
- https://github.com/scikit-hep/uproot5/pull/1012
- https://github.com/scikit-hep/uproot5/pull/1014
- https://github.com/scikit-hep/uproot5/pull/1015
- https://github.com/scikit-hep/uproot5/pull/1016
- https://github.com/scikit-hep/uproot5/pull/1018
- https://github.com/scikit-hep/uproot5/pull/1020
- https://github.com/fsspec/filesystem_spec/pull/1426
- https://github.com/scikit-hep/uproot5/pull/1021
- https://github.com/scikit-hep/uproot5/pull/1022
- https://github.com/scikit-hep/uproot5/pull/1023
- https://github.com/scikit-hep/uproot5/pull/1024
- https://github.com/scikit-hep/uproot5/pull/1026
- https://github.com/scikit-hep/uproot5/pull/1027

Issue #920, opened 2023-07-12
Can't open files with long file names


BenGalewsky:
Uproot Version: 5.0.10

I have a very long filename:
```
filename = "/servicey/output/7242feb5-db50-46b3-a3f7-9588c2a204ac/scratch/root:::atlasxrootd-kit.gridka.de:1094::pnfs:gridka.de:atlas:disk-only:atlasdatadisk:rucio:mc16_13TeV:05:ce:DAOD_PHYS.30899209._000097.pool.root.1"
```

When I attempt to open the file with
```python
uproot.open(filenme)
```

It can't open the file and it looks like someone is truncating the name:
```
FileNotFoundError: [Errno 2] No such file or directory: '/servicey/output/7242feb5-db50-46b3-a3f7-9588c2a204ac/scratch/root:::atlasxrootd-kit.gridka.de:1094::pnfs:gridka.de:atlas:disk-only:atlasdatadisk:rucio:mc16_13TeV:05:ce'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/site-packages/uproot/reading.py", line 141, in open
    file = ReadOnlyFile(
  File "/usr/local/lib/python3.10/site-packages/uproot/reading.py", line 581, in __init__
    self._source = Source(
  File "/usr/local/lib/python3.10/site-packages/uproot/source/file.py", line 108, in __init__
    self._open()
  File "/usr/local/lib/python3.10/site-packages/uproot/source/file.py", line 118, in _open
    self._fallback = uproot.source.file.MultithreadedFileSource(
  File "/usr/local/lib/python3.10/site-packages/uproot/source/file.py", line 250, in __init__
    self._open()
  File "/usr/local/lib/python3.10/site-packages/uproot/source/file.py", line 254, in _open
    [FileResource(self._file_path) for x in range(self._num_workers)]
  File "/usr/local/lib/python3.10/site-packages/uproot/source/file.py", line 254, in <listcomp>
    [FileResource(self._file_path) for x in range(self._num_workers)]
  File "/usr/local/lib/python3.10/site-packages/uproot/source/file.py", line 38, in __init__
    raise uproot._util._file_not_found(file_path) from err
FileNotFoundError: file not found
```

Note that the error message is showing the filename truncated to 168 characters.

For sanity checking purposes, I tried using python's open function:
```python
open(filename)
[...]

PR #976, opened 2023-10-05
feat: refactor url - object split (motivated by `fsspec` integration)

lobis:
Currently the `uproot._util.py` file contains some helper methods for things such as url parsing, split root object from file path etc.

Besides the splitting of the root object from a URI which is non-standard, fsspec should be able to handle everything else and we could remove or simplify some helper methods.

Since this uses fsspec but fsspec is not currently a dependency my solution was to do the new url split using fsspec only if its installed, otherwise do the old one, which may cause some problems.

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -280,6 +280,41 @@ def file_object_path_split(path):
     """
     Split a path with a colon into a file path and an object-in-file path.
     """
+
+    # TODO: rename path for URI or URL (confusing, here the input "path" can also include schema, etc.)
+    try:
+        import fsspec.utils
+
+        # remove whitespace
+        path = path.rstrip()
+        path = path.lstrip()
+
+        # split url into parts
+        parsed_url = fsspec.utils.urlsplit(path)
+
+        parts = parsed_url.path.split(":")
+        if len(parts) == 1:
+            obj = None
+        elif len(parts) == 2:
+            obj = parts[1]
+            path = path.rstrip(":" + obj)
+            obj.lstrip()
+        else:
+            raise ValueError(
+                f"too many colons in file path: {path} for url {parsed_url}"
+            )
+
+        return path, obj
+        # path does not include the scheme, etc
+
+    except ImportError:
+        return _file_object_path_split_use_when_fsspec_not_installed(path)
+
+
+def _file_object_path_split_use_when_fsspec_not_installed(path):
+    """
+    Split a path with a colon into a file path and an object-in-file path.
+    """
     path = regularize_path(path)
 
     try:
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -276,7 +276,7 @@ _remote_schemes = ["ROOT", "S3", "HTTP", "HTTPS"]
 _schemes = ["FILE", *_remote_schemes]
 
 
-def file_object_path_split(path: str) -> tuple[str, str] or tuple[str, None]:
+def file_object_path_split(path):
     """
     Split a path with a colon into a file path and an object-in-file path.
 
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -319,10 +319,37 @@ def file_object_path_split(path):
 def file_path_to_source_class(file_path, options):
     """
     Use a file path to get the :doc:`uproot.source.chunk.Source` class that would read it.
+
+    Returns a tuple of (class, file_path) where the class is a subclass of :doc:`uproot.source.chunk.Source`.
+
+    The "handler" option is the preferred way to specify a custom source class.
+    The "*_handler" options are for backwards compatibility and will override the "handler" option if set.
     """
     import uproot.source.chunk
 
     file_path = regularize_path(file_path)
+
+    out = options["handler"]
+    if out is not None:
+        if not (isinstance(out, type) and issubclass(out, uproot.source.chunk.Source)):
+            raise TypeError(
+                "'handler' is not a class object inheriting from Source: " + repr(out)
+            )
+        # check if "object_handler" is set
+        if (
+            options["object_handler"] is not None
+            or options["file_handler"] is not None
+            or options["xrootd_handler"] is not None
+            or options["s3_handler"] is not None
+            or options["http_handler"] is not None
+        ):
+            # These options will override the "handler" option for backwards compatibility
+            warnings.warn(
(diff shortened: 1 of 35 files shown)


lobis:
Perhaps it would make sense to add `fsspec` as a dependency for this PR after the next major release? This would avoid having potentially different behaviours for users having `fsspec` installed or not, and I guess it would also make the code easier to write and read we wouldn't have to handle both cases.

src/uproot/_util.py
```diff
@@ -280,6 +280,44 @@ def file_object_path_split(path):
...
+
+    # TODO: once fsspec is added as a dependency, this try/except can be removed and the old function can be deleted
+
+    # TODO: rename path for URI or URL (confusing, here the input "path" can also include schema, etc.)
+    try:
+        import fsspec.utils
+
+        # remove whitespace
+        path = path.rstrip()
+        path = path.lstrip()
```
jpivarski, line 292:
```suggestion
        path = path.strip()
```

jpivarski (commented):
When this is done, fsspec will become a strict dependency.

> This would avoid having potentially different behaviours for users having fsspec installed or not

If that's possible, then you would be introducing a behavior change now for those users who happen to have fsspec installed for other reasons. We don't want the path-handling to change at all. Could you get more certainty about that?

Meanwhile, this is not a change that has to happen now (or ever, technically). The original splitting was using a Python standard library function. Even when we start relying on fsspec for all file backends, there's no reason we couldn't still use the standard `urlparse` for URL parsing. Does `fsspec.utils.urlsplit` do something special?

I'm on the fence about this one.

lobis:
> When this is done, fsspec will become a strict dependency.
> 
> > This would avoid having potentially different behaviours for users having fsspec installed or not
> 
> If that's possible, then you would be introducing a behavior change now for those users who happen to have fsspec installed for other reasons. We don't want the path-handling to change at all. Could you get more certainty about that?
> 
> Meanwhile, this is not a change that has to happen now (or ever, technically). The original splitting was using a Python standard library function. Even when we start relying on fsspec for all file backends, there's no reason we couldn't still use the standard `urlparse` for URL parsing. Does `fsspec.utils.urlsplit` do something special?
> 
> I'm on the fence about this one.

Yes I think you are right we should not use the fsspec url parsing when we can just use urllib. At the end I ended up refactoring the helper method to use `urllib` and added an explicit test for it. The old method got confused with things such as double `:` in the url, etc.

I feel that this method is more concise but I am worried it does not fully reproduce the old one (everything looks good so far, all the tests pass).

nsmith-:
Might be worth reviewing all the supported syntaxes of `files` argument in https://uproot.readthedocs.io/en/latest/uproot._dask.dask.html as it is a bit more than `uproot.open` and thinking about what would make the most sense to normalize paths to something uniform.

jpivarski:
The colon-splitting between filename and object (not related to #974—that's a colon in an object path) has been a lot of trouble. There's a list of problems it's caused on https://github.com/scikit-hep/uproot5/issues/920#issuecomment-1633050022, including platform-dependent issues with `C:\` on Windows.

On page 19 of [this talk](https://indico.jlab.org/event/459/contributions/11547/) I analyzed a large dataset of user code to see how many people are using it, to see if we could ever get rid of it. The result was 10% of `uproot.open` calls explicitly use the colon and 64% are unknown because they pass in a variable as the filename.

So the path interpretation is unfortunately complex and we should leave it untouched if possible. It seems to me that it should be possible, since replacing the backend doesn't change the path-or-URL that we send to the backend.

nsmith-:
Personally I'd be in favor of dropping `url:filepath` in favor of `{"url": "filepath"}` but I believe there was some discussion in the past on this.

jpivarski:
> Might be worth reviewing all the supported syntaxes of `files` argument in https://uproot.readthedocs.io/en/latest/uproot._dask.dask.html as it is a bit more than `uproot.open` and thinking about what would make the most sense to normalize paths to something uniform.

I think we have consistency across the file-opening functions under control. File syntax is never interpreted differently by different file-opening functions, but some functions have more options than others.

  * `uproot.open` can only take one file, so it has
    * `str`/`bytes` (filepath-colon-objectpath)
    * `pathlib.Path` (filepath only)
    * object with `read` and `seek` methods (file-like object)
    * length-1 dict of filepath, objectpath (control where the split happens)
  * `uproot.iterate` can take multiple files, so it has
    * `str`/`bytes` (filepath-colon-objectpath)
    * `pathlib.Path` (filepath only)
    * glob syntax in `str`/`bytes` or `pathlib.Path`, including bash extensions (multiple filepaths, but only one objectpath)
    * any-length dict of filepath, objectpath (control where the split happens and get multiple filepaths, multiple objectpaths)
    * already-open TTree objects (to chain them)
    * iterables of the above (to chain them)
  * `uproot.concatenate` can also take multiple files, so it has
    * all the same options as `uproot.iterate`
  * `uproot.dask` can take multiple files and needs to partition them somehow, so it has
    * all the same options as `uproot.iterate`
    * any-length dict of filepath, dict of `{"object_path": OBJECTPATH, "steps": STEPS}` where `OBJECTPATH` is the objectpath and `STEPS` is either a list of offsets or a list of start-stop pairs (entry numbers for the partitions).

(`uproot.dask` has three ways to partition, but each one is mutually exclusive of the other two. If you try to use more than one, it will raise an error.)

The above is complex, but I believe that it is under control. Each one of these methods was motivated by a request (a long history...).

lobis:
To clarify (after an in-person discussion with @jpivarski I think there was some confusion):

- This was motivated by the `fsspec` integration (https://github.com/scikit-hep/uproot5/issues/972).
- I tried using the newly added `fsspec` source with some exotic urls such as `github://scikit-hep:scikit-hep-testdata@v0.4.33/src/skhep_testdata/data/uproot-issue121.root` but it didn't work due to the helper function not working correctly, this is the reason for this PR: make a more robust url / object split helper function.
- I thought it made sense to refactor this function to use `urllib` as it's being used elsewhere in the code.
- The new implementation is more concise and it looks like it handles all previous cases (atleast all the tests pass, if this is not the case we should add them).
- I added a new test just to explicitly test this (as the github api url test had to be skipped due to api rate limits).
- This is not just a refactoring, it should make this method correclty handle more cases and hopefully 100% cover previous ones.

jpivarski (approved):
We've been staring at this for a while, it _should_ be okay. Let's try it out in production and find out!

(We'll know where to look if anyone runs into any new problems with colon-parsing.)

PR #979, opened 2023-10-06
feat: `fsspec` source non-blocking chunks

lobis:
This PR adds a non-blocking implementation of the `chunks` interface.

It uses the `AbstractFileSystem.cat_file` api to get a byte range of the file. This api should only request the selected range (as opposed to the whole file, then slicing). In order to make sure this is the case I opened a discussion [here](https://github.com/fsspec/filesystem_spec/discussions/1388).

It also adds some type annotations to the `uproot.source` files.


src/uproot/source/fsspec.py
```diff
@@ -97,17 +97,20 @@ def chunks(self, ranges, notifications):
...
-            [stop for _, stop in ranges],
-        )
+
         chunks = []
-        for item, (start, stop) in zip(data, ranges):
-            future = uproot.source.futures.TrivialFuture(item)
+        for start, stop in ranges:
+            future = uproot.source.futures.Future(
+                self._fs.cat_file, (self._file_path, start, stop)
+            )
```
jpivarski, line 105: This correctly makes a Future and returns immediately. That's good.

src/uproot/source/fsspec.py
```diff
@@ -97,17 +97,20 @@ def chunks(self, ranges, notifications):
...
             chunk = uproot.source.chunk.Chunk(self, start, stop, future)
             uproot.source.chunk.notifier(chunk, notifications)()
             chunks.append(chunk)
+
+        # TODO: Am I using the futures API correctly? Should I use some executor / resource future instead?
+        for chunk in chunks:
+            chunk.future._run()
```
jpivarski, line 112: This blocks until all `chunks` are full, which defeats the purpose of making them Futures in the above loop. The idea is that this function should return a list of `chunks` that might or might not be full (more likely not, if the server is slow).

src/uproot/source/fsspec.py
```diff
@@ -97,17 +97,20 @@ def chunks(self, ranges, notifications):
...
+
         chunks = []
-        for item, (start, stop) in zip(data, ranges):
-            future = uproot.source.futures.TrivialFuture(item)
+        for start, stop in ranges:
+            future = uproot.source.futures.Future(
+                self._fs.cat_file, (self._file_path, start, stop)
+            )
             chunk = uproot.source.chunk.Chunk(self, start, stop, future)
             uproot.source.chunk.notifier(chunk, notifications)()
```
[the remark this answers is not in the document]
lobis: thanks for the detailed explanation!

tests/test_0692_fsspec.py
```diff
@@ -66,3 +67,32 @@ def test_open_fsspec_xrootd():
...
+        start_time = time.time()
+        chunks = source.chunks(
+            [(0, 100), (50, 55), (200, 400)], notifications=notifications
+        )
+        time_to_get_chunks = time.time() - start_time
+        # TODO: this might not be the best way to test it is non-blocking
+        # This is non-blocking, so it should be fast (0.0005390644073486328 seconds last time I ran it)
+        assert (
+            time_to_get_chunks < 0.01
+        ), f"This should be fast (non-blocking). Elapsed: {time_to_get_chunks} seconds"
```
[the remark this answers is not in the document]
jpivarski: No—do not test the blockingness. In CI, it will be running on some machine that might be inherently slower or something else might be running at the same time, and users with the sdist are invited to run the tests on their own machine—any of these can make the test fail for false reasons.

Basically, we can't test invisible aspects of computing. The performance characteristics are inside the black box, and tests can only test black-box outputs. But we can just be sure to code it in a way that has the right characteristics.

lobis:
The PR is again ready to be reviewed.

Now it should work as intended, the `chunks` method should be non-blocking.

I have some questions regarding the management of the executor. Currently I am creating a local executor in the `chunks` method but perhaps I should do proper lifecycle management in a similar fashion to what is being done in the http source.

src/uproot/source/fsspec.py
```diff
@@ -97,17 +97,21 @@ def chunks(self, ranges, notifications):
         self._num_requests += 1
         self._num_requested_chunks += len(ranges)
         self._num_requested_bytes += sum(stop - start for start, stop in ranges)
-        data = self._fs.cat_ranges(
-            [self._file_path] * len(ranges),
-            [start for start, _ in ranges],
-            [stop for _, stop in ranges],
-        )
+
+        executor = uproot.source.futures.ResourceThreadPoolExecutor(ranges)
```
jpivarski, line 101: This executor goes out of scope, possibly/probably before the work is finished. I think that would shut it down prematurely (the work never finishes, which would cause the event loop in src/uproot/behaviors/TBranch.py to hang).

The HTTPSource and XRootDSource do this correctly: they have a `ResourceThreadPoolExecutor` attached to the Source (`self`). The "resource" is a file handle, something that absolutely must be closed when the remote file leaves its `with` block. Most kinds of file handles are limited in number (too many open file handles and the OS will complain).

`ranges` is not a limited resource. If the fsspec interface is truly stateless (no handle), then we can get away with a `ThreadPoolExecutor`, rather than a `ResourceThreadPoolExecutor`.

jpivarski (commented):
This comes down to a fundamental question: what is fsspec's scope? How much does it do? There's this `cat_file` interface, which does the whole process from URL to file bytes, but it's a blocking interface. Is there an async interface? If so, does it have a way to attach a function to call when the async function is done? 

If so, then FSSpecSource doesn't need to have an executor. Suppose there's a function

```python
FS.async_cat_file(path: str, start: int, stop: int) -> concurrent.futures.Future
```

then you could use it like this:

```python
for item, (start, stop) in zip(data, ranges):
    future = self._fs.async_cat_file(self._file_path, start, stop)
    chunk = uproot.source.chunk.Chunk(self, start, stop, future)
    future.add_done_callback(uproot.source.chunk.notifier(chunk, notifications))
    chunks.append(chunk)
```

(although I think that `add_done_callback` expects the function to take 1 argument and our notifier takes 0 arguments, but these things are easy to work around). If the fsspec interface uses `asyncio.Future` instead of `concurrent.futures.Future` (which we designed around), then it would be a little different, which could be done by wrapping it with the interface we expect (i.e. `future.result()` instead of `await future`).

So... what can fsspec do for us?

src/uproot/source/fsspec.py
lobis: My initial thought was also that the executor going out of scope prematurely would cause it to shutdown and not finish the work but it passes the test so it's somehow working.

jpivarski: By accident, perhaps? I don't know. But we definitely can't rely on it.

lobis:
> If so, then FSSpecSource doesn't need to have an executor

I don't understand this, don't you always need some kind of executor to run futures?

jpivarski:
The executor might be inside of the fsspec library. I don't know that—I'm just thinking it might be, and if it is, we should take advantage of it.

Also, it might use `async` instead of an executor, but the effect would be the same: we might be able to get it to do the concurrent waiting-for-I/O for us.

lobis:
> This comes down to a fundamental question: what is fsspec's scope? How much does it do? There's this `cat_file` interface, which does the whole process from URL to file bytes, but it's a blocking interface. Is there an async interface? If so, does it have a way to attach a function to call when the async function is done?

From my understanding `fsspec` supports an async interface only for the http file system. This is enabled by passing the option `asynchronous=True` on creation. If this option is set to true, some methods of the filesystem will become asynchronous (such as `cat_file`). Running in asynchronous mode would need an event loop.

I imagine it would be possible to work with the sync filesystem but use an async one inside the `chunks` method so all the `asyncio` logic will be contained, otherwise it can get complicated. I will look into this.

Regarding this PR: I think it's done (unless there is some mistake) as currently the chunks interface is non-blocking. Depending on the complexity of leveraging the fsspec asyncio interface I will include it into this PR, a separate PR, or do nothing.

src/uproot/source/fsspec.py
```diff
@@ -13,16 +15,28 @@ class FSSpecSource(uproot.source.chunk.Source):
...
 
         # Remove uproot-specific options (should be done earlier)
-        exclude_keys = set(uproot.reading.open.defaults.keys())
-        opts = {k: v for k, v in kwargs.items() if k not in exclude_keys}
+        exclude_keys = set(default_options.keys())
+        opts = {k: v for k, v in options.items() if k not in exclude_keys}
 
         self._fs, self._file_path = fsspec.core.url_to_fs(file_path, **opts)
+
+        if self._use_threads and self._fs.__class__.__name__ == "HTTPFileSystem":
```
jpivarski, line 33:
```suggestion
        if self._use_threads and type(self._fs).__name__ == "HTTPFileSystem":
```

I see that we can't do an `isinstance` check against `fsspec.implementations.http.HTTPFileSystem` because we don't know if `fsspec.implementations.http` is installed. It's an optional dependency of fsspec.

src/uproot/source/fsspec.py
```diff
@@ -37,12 +51,22 @@ def __repr__(self):
...
 
+    def __getstate__(self):
+        state = dict(self.__dict__)
+        state.pop("_executor")
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+        self._open()
+
```
jpivarski, line 62:
```suggestion
```

I don't see any reason why a `Source` should be pickleable.

jpivarski (approved):
This is looking good. Does it depend on #983?

src/uproot/source/fsspec.py
lobis: I just implemented it because it was implemented in other sources. I guess I can safely remove those in this PR since it's not being use anywhere.

lobis:
> This is looking good. Does it depend on #983?

It does not.

src/uproot/source/fsspec.py
jpivarski: We'll keep pickleability.

jpivarski:
`test_0692_fsspec.py::test_fsspec_chunks` needs

```python
pytest.importorskip("aiohttp")
```

in the failing function. That's a non-strict dependency, so any of our tests that use it have to be skipped when the package isn't available.

This one failed because it's Python 3.12 and presumably the package isn't available _yet_, so I'm glad we got this chance to check.

If there are any tests that need fsspec or another non-strict dependency to run, those tests need to have `pytest.importorskip`, either in the individual function or at the top of the Python file for module-level.

jpivarski:
I think file.py needs `from __future__ import annotations`.

PR #943, opened 2023-08-17
feat: refactoring the AwkwardForth code-discovery process

jpivarski:
tests/test_0637-setup-tests-for-AwkwardForth.py:

- [x] test_00
- [x] test_01
- [x] test_02
- [x] test_03
- [x] test_04
- [x] test_05
- [x] test_06
- [x] test_07
- [x] test_08
- [x] test_09
- [x] test_10
- [x] test_11
- [x] test_12
- [x] test_13
- [x] test_14
- [x] test_15
- [x] test_16
- [x] test_17
- [x] test_18
- [x] test_19
- [x] test_20
- [x] test_21
- [x] test_22
- [x] test_23
- [x] test_24
- [x] test_25
- [x] test_27
- [x] test_28
- [x] test_30
- [x] test_31
- [x] test_32
- [x] test_33
- [x] test_34
- [x] test_35
- [x] test_36
- [x] test_37
- [x] test_38
- [x] test_39
- [x] test_40
- [x] test_41
- [x] test_42
- [x] test_43
- [x] test_44
- [x] test_47
- [x] test_51
- [x] test_52
- [x] test_53
- [x] test_54
- [x] test_55
- [x] test_56
- [x] test_57
- [x] test_58
- [x] test_59
- [x] test_60
- [x] test_61
- [x] test_62
- [x] test_63
- [x] test_64
- [x] test_65
- [x] test_66
- [x] test_67
- [x] test_68
- [x] test_69
- [x] test_70
- [x] test_71
- [x] test_72
- [x] test_73
- [x] test_74
- [x] test_75
- [x] test_76
- [x] test_77
- [x] test_78
- [x] test_79
- [x] test_80
- [x] test_81

tests/test_0798_DAOD_PHYSLITE.py

- [x] test_AnalysisJetsAuxDyn_GhostTrack
- [x] test_TruthBosonAuxDyn_childLinks
- [x] test_TruthPhotonsAuxDyn_parentLinks
- [x] test_TruthTopAuxDyn_parentLinks

PR #980, opened 2023-10-06
chore: bump Python version

agoose77:
This PR introduces `pyupgrade`, bumps the oldest supported Python interpreter from 3.7 to 3.8, and adds support for 3.12 in the metadata.

It also applies a fix for the removal of `NumpyArray.__array__`.


jpivarski (approved):
Yes, 5.1.0 is a good time for Uproot to drop Python 3.7, though this should merge into the `agoose77/wip-shape-tracking` branch.

Or better yet, a new `main-510` branch should be introduced that `agoose77/wip-shape-tracking` and `agoose77/feat!-bump-python-version` both merge into. This `main-510` will accumulate everything that's going into Uproot 5.1.0 and not anything that will go in later. When 5.1.0 is finally released, `main-510` merges into `main` and disappears. During this transition time, we shouldn't be merging anything big into `main`, either.

I'm in principle in favor of pyupgrade, but if it's going to make a lot of small changes across the codebase, _we should not introduce it_ until after
  * merging of `main-510` into `main`
  * merging of #943 into `main` (that's one big PR that touches a lot of code across the codebase)

lobis:
I guess we should also remove python 3.7 from the GitHub Actions workflow file right?

jpivarski:
> I guess we should also remove python 3.7 from the GitHub Actions workflow file right?

Good point! That would go with this PR. (Go ahead and make the modification directly; we won't merge it until everyone's satisfied with it.)

agoose77:
@lobis that's weird — I thought I had done that here ... I think I'm working on too many PRs at once :laughing: Thanks!

lobis:
Should we also add python 3.12 to the CI? (I tried doing so but found some problems, it's a bit hard to debug with the current status of the pipeline, but I think it should be an easy fix).

agoose77:
> Should we also add python 3.12 to the CI? 

Definitely!

agoose77:
@jpivarski this is ready to merge. It updates our Python min, max versions (>=3.8, <=3.12), and fixes a use of `np.asarray(content)` that is broken by the latest awkward.

PR #984, opened 2023-10-11
feat: remove `*_handler` option, fsspec source as default handler

lobis:
After https://github.com/scikit-hep/uproot5/pull/971 the `*_handler` option is deprecated but the options remained for backwards compatibility.

This PR removes the backwards compatible options and sets the `fsspec` source as the default handler, which should be able to handle anything. The `handler` option is kept for to allow the user to specify a custom source or one in the `uproot.source` module. These sources are kept but I think there shouldn't be a reason to use them.


lobis:
Closing but saved for later if we remember.

PR #985, opened 2023-10-12
fix: pull out `.data` from `NumpyArray`

agoose77:


jpivarski:
starterkit.web.cern.ch just started working again in #971, so I'm going to just try re-running the failed tests to see if they work here, too.

jpivarski:
In the results of `fgrep -rl '.asarray(' src` (on `main`):

* src/uproot/_util.py:
  * used in `ensure_numpy`
* src/uproot/containers.py: nothing that could be a `Content`
* src/uproot/writing/_cascadetree.py:
  * used on the output of `ak.num` and `ak.flatten`, but those are all implicitly `highlevel=True`
  * used on `branch_array.layout.offsets` and `layout.offsets` (`Index`)
  * used on `branch_array` and `content`, which might be `Content`
* src/uproot/behaviors/TProfile2D.py: nothing that could be a `Content`
* src/uproot/behaviors/TH3.py: nothing that could be a `Content`
* src/uproot/behaviors/TGraphAsymmErrors.py: nothing that could be a `Content`
* src/uproot/behaviors/TProfile3D.py: nothing that could be a `Content`
* src/uproot/behaviors/TH2.py: nothing that could be a `Content`
* src/uproot/behaviors/TGraph.py: nothing that could be a `Content`
* src/uproot/behaviors/TAxis.py: nothing that could be a `Content`
* src/uproot/behaviors/TProfile.py: nothing that could be a `Content`
* src/uproot/behaviors/TH1.py: nothing that could be a `Content`
* src/uproot/interpretation/library.py:
  * used on `array.offsets` (`Index`)
  * used on `array.data` where `array` is a `NumpyArray` (good)

In the results of `fgrep -rl 'ensure_numpy' src`:

* src/uproot/writing/identify.py:
  * used in cases that might be `Content`

`np.array` is rarely used to cast data (almost all cases are `np.asarray`) and those that are casting data all look safe to me: I don't think any of them could be `Content`.

agoose77:
I *think* this PR catches all cases, now.

jpivarski (approved):
I agree. I just went through the list and I think everything is correct. This will be Uproot 5.0.13.

PR #986, opened 2023-10-12
fix: support `Content` objects in writing

agoose77:


jpivarski:
This is a forward-port of #985, right? It's missing `content` → `content.data` in

https://github.com/scikit-hep/uproot5/blob/4a977f65d112310f3607ac7b679d43df46cdc8fa/src/uproot/writing/_cascadetree.py#L785

agoose77:
Yes, but that change is already in this branch.

jpivarski:
Indeed it is!

jpivarski (approved):

PR #990, opened 2023-10-13
fix: typo in `uproot.source.xrootd.XRootDSource` name

jpivarski:
I scanned all of @lobis's PRs for "XRootDSource" and "source.root" and it only appeared in #971.

We need a test for it, though.

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -432,7 +432,7 @@ after the first `import uproot` or use `@pytest.mark.filterwarnings("error:::upr
     elif parsed_url.scheme.upper() == "ROOT":
         out = options["xrootd_handler"]
         if out is None:
-            out = uproot.source.root.XRootDSource
+            out = uproot.source.xrootd.XRootDSource
         else:
             warnings.warn(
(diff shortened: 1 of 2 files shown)


lobis (approved):

lobis:
@jpivarski I added a test for it. `uproot.open` was not being tested against `xrootd` files (the source was being used directly) that is why this typo had been passing the tests. (Perhaps this test should be in another file though).

jpivarski (commented):
Most of the `Sources` tests were very early, around `test_0001_*.py` through `test_0005_*.py`. They were probably written before `uproot.open` existed, and that's how this slipped through. I know there are `uproot.open` uses with HTTP in other parts of the test suite (testing things other than merely being able to open the file with HTTP).

But this test is a positive improvement: it's walking through the control flow that leads from a URL with `root://` to loading the `XRootDSource`, and would have caught the bug. So for now, at least, I'll merge this and make Uproot 5.1.1.

I'm not allowed to approve this PR, probably because I started it. Oh well, I'll merge without approval.

PR #992, opened 2023-10-15
feat: `asyncio` LoopExecutor and async fsspec source

lobis:
This PR adds a new type of executor exclusive to the fsspec source that allows to run tasks asynchronously in a single thread using the fsspec event loop (for async-capable backends).

The list of commits for this PR is very long because the design changed a few times:

First we added a generic `LoopExecutor` capable of submitting tasks to a loop running in a different thread. This class was responsible for managing the lifetime of the loop and thread. There were some issues when submitting fsspec coroutines (I still don't really understand why).

Finally I realised that fsspec provides it's own loop, so we avoid having to manage all this and just use the fsspec's implementation of a "loop in a can" (@nsmith- ). This executor is defined in the `fsspec.py` source file as currently it only makes sense to use it from the fsspec source since you have to provide a loop. The executor does nothing, it's just a wrapper for compatibility.

When implementing this thin executor I thought that implementing an ABC executor class would be a good idea (IMHO).

### DIFF
--- a/src/uproot/source/fsspec.py
+++ b/src/uproot/source/fsspec.py
@@ -87,7 +87,7 @@ class FSSpecSource(uproot.source.chunk.Source):
     def __exit__(self, exception_type, exception_value, traceback):
         self._fh = None
         self._file.__exit__(exception_type, exception_value, traceback)
-        self._executor.shutdown()
+        # self._executor.shutdown()
 
     def chunk(self, start, stop) -> uproot.source.chunk.Chunk:
         """
--- a/src/uproot/source/fsspec.py
+++ b/src/uproot/source/fsspec.py
@@ -38,9 +38,16 @@ class FSSpecSource(uproot.source.chunk.Source):
         self._fs, self._file_path = fsspec.core.url_to_fs(file_path, **opts)
 
         if self._use_threads:
-            self._executor = concurrent.futures.ThreadPoolExecutor(
-                max_workers=self._num_workers
-            )
+            if self._fs.async_impl:
+                self._executor = uproot.source.futures.LoopExecutor()
+                # Bind the loop to the filesystem
+                self._fs = fsspec.filesystem(
+                    protocol=self._fs.protocol, loop=self._executor.loop
+                )
+            else:
+                self._executor = concurrent.futures.ThreadPoolExecutor(
+                    max_workers=self._num_workers
+                )
         else:
             self._executor = uproot.source.futures.TrivialExecutor()
 
@@ -133,9 +140,15 @@ class FSSpecSource(uproot.source.chunk.Source):
 
         chunks = []
         for start, stop in ranges:
-            future = self._executor.submit(
-                self._fs.cat_file, self._file_path, start, stop
-            )
+            if self._fs.async_impl:
+                # submit a coroutine
+                future = self._executor.submit(
+                    self._fs._cat_file(self._file_path, start, stop)
+                )
+            else:
+                future = self._executor.submit(
+                    self._fs.cat_file, self._file_path, start, stop
+                )
             chunk = uproot.source.chunk.Chunk(self, start, stop, future)
             future.add_done_callback(uproot.source.chunk.notifier(chunk, notifications))
             chunks.append(chunk)
--- a/src/uproot/source/fsspec.py
+++ b/src/uproot/source/fsspec.py
@@ -30,9 +30,9 @@ class FSSpecSource(uproot.source.chunk.Source):
         default_options = uproot.reading.open.defaults
         self._use_threads = options.get("use_threads", default_options["use_threads"])
         self._num_workers = options.get("num_workers", default_options["num_workers"])
-        self._use_async = (
-            False  # automatically determined from the protocol and other options
-        )
+        # Add the possibility to set use_async directly as a hidden option.
+        # It is not encouraged to do so but may be useful for testing purposes.
+        self._use_async = options.get("use_async", None) if self._use_threads else False
 
         # TODO: is timeout always valid?
 
@@ -42,12 +42,18 @@ class FSSpecSource(uproot.source.chunk.Source):
 
         protocol = fsspec.core.split_protocol(file_path)[0]
         fs_has_async_impl = fsspec.get_filesystem_class(protocol=protocol).async_impl
+        # If not explicitly set (default), use async if possible
+        self._use_async = (
+            fs_has_async_impl if self._use_async is None else self._use_async
+        )
+        if self._use_async and not fs_has_async_impl:
+            # This should never be triggered unless the user explicitly set the `use_async` flag for a non-async backend
+            raise ValueError(f"Filesystem {protocol} does not support async")
 
         if not self._use_threads:
             self._executor = uproot.source.futures.TrivialExecutor()
-        elif fs_has_async_impl:
+        elif self._use_async:
             self._executor = FSSpecLoopExecutor(fsspec.asyn.get_loop())
-            self._use_async = True
         else:
             self._executor = concurrent.futures.ThreadPoolExecutor(
                 max_workers=self._num_workers
@@ -170,6 +176,16 @@ class FSSpecSource(uproot.source.chunk.Source):
         The number of bytes in the file.
         """
         return self._fs.size(self._file_path)
+
+    # We need this because user may use other executors not defined in `uproot.source` (such as `concurrent.futures`)
+    # that do not have this interface. If not defined it defaults to calling this property on the source's executor
+    @property
+    def closed(self) -> bool:
+        """
+        True if the associated file/connection/thread pool is closed; False
+        otherwise.
+        """
+        return False
 
 
 class FSSpecLoopExecutor(uproot.source.futures.Executor):
--- a/src/uproot/source/fsspec.py
+++ b/src/uproot/source/fsspec.py
@@ -139,6 +139,7 @@ class FSSpecSource(uproot.source.chunk.Source):
         self._num_requested_bytes += sum(stop - start for start, stop in ranges)
 
         chunks = []
+        # _cat_file is async while cat_file is not
         cat_file = self._fs._cat_file if self._fs.async_impl else self._fs.cat_file
         for start, stop in ranges:
             future = self._executor.submit(cat_file, self._file_path, start, stop)
--- a/src/uproot/source/fsspec.py
+++ b/src/uproot/source/fsspec.py
@@ -22,7 +22,7 @@ class FSSpecSource(uproot.source.chunk.Source):
     to get many chunks in one request.
     """
 
-    def __init__(self, file_path, **options):
+    def __init__(self, file_path: str, **options):
         import fsspec.core
 
         default_options = uproot.reading.open.defaults
@@ -89,7 +89,7 @@ class FSSpecSource(uproot.source.chunk.Source):
         self._file.__exit__(exception_type, exception_value, traceback)
         # self._executor.shutdown()
 
-    def chunk(self, start, stop) -> uproot.source.chunk.Chunk:
+    def chunk(self, start: int, stop: int) -> uproot.source.chunk.Chunk:
         """
         Args:
             start (int): Seek position of the first byte to include.
--- a/src/uproot/source/fsspec.py
+++ b/src/uproot/source/fsspec.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import concurrent.futures
 import queue
 
 import uproot
@@ -27,7 +26,6 @@ class FSSpecSource(uproot.source.chunk.Source):
 
         default_options = uproot.reading.open.defaults
         self._use_threads = options.get("use_threads", default_options["use_threads"])
-        self._num_workers = options.get("num_workers", default_options["num_workers"])
 
         # TODO: is timeout always valid?
 
@@ -38,22 +36,17 @@ class FSSpecSource(uproot.source.chunk.Source):
         self._fs, self._file_path = fsspec.core.url_to_fs(file_path, **opts)
 
         if self._use_threads:
-            if self._fs.async_impl:
-                self._executor = uproot.source.futures.LoopExecutor()
+            self._executor = uproot.source.futures.LoopExecutor()
 
-                # Bind the loop to the filesystem
-                async def make_fs():
-                    return fsspec.filesystem(
-                        protocol=self._fs.protocol, loop=self._executor.loop
-                    )
-
-                self._fs = self._executor.submit(make_fs).result()
-                assert self._fs.loop is self._executor.loop, "loop not bound"
-                assert self._fs.loop.is_running(), "loop not running"
-            else:
-                self._executor = concurrent.futures.ThreadPoolExecutor(
-                    max_workers=self._num_workers
+            # Bind the loop to the filesystem
+            async def make_fs():
+                return fsspec.filesystem(
+                    protocol=self._fs.protocol, loop=self._executor.loop
                 )
+
+            self._fs = self._executor.submit(make_fs).result()
+            assert self._executor.loop is self._executor.loop, "loop not bound"
+            assert self._executor.loop.is_running(), "loop not running"
         else:
             self._executor = uproot.source.futures.TrivialExecutor()
 
@@ -145,11 +138,18 @@ class FSSpecSource(uproot.source.chunk.Source):
         self._num_requested_bytes += sum(stop - start for start, stop in ranges)
 
         chunks = []
-        # _cat_file is async while cat_file is not
-        use_async = self._fs.async_impl and isinstance(
-            self._executor, uproot.source.futures.LoopExecutor
-        )
-        cat_file = self._fs._cat_file if use_async else self._fs.cat_file
+
+        async def cat_file_wrapped(*args):
+            # wrapper for the filesystems without async_impl
+            return await self._executor.loop.run_in_executor(
+                None, self._fs.cat_file, *args
+            )
+
+        # The fsspec source will always use the LoopExecutor if `use_threads` is True,
+        # this means that the `cat_file` method must be async, so we wrap it in case the filesystem is not async-capable
+        cat_file_async = self._fs._cat_file if self._fs.async_impl else cat_file_wrapped
+        cat_file = cat_file_async if self._use_threads else self._fs.cat_file
+
         for start, stop in ranges:
             future = self._executor.submit(cat_file, self._file_path, start, stop)
             chunk = uproot.source.chunk.Chunk(self, start, stop, future)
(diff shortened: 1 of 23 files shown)


src/uproot/source/fsspec.py
```diff
@@ -38,9 +38,20 @@ def __init__(self, file_path, **options):
...
+                        protocol=self._fs.protocol, loop=self._executor.loop
+                    )
+
+                self._fs = self._executor.submit(make_fs).result()
+            else:
+                self._executor = concurrent.futures.ThreadPoolExecutor(
+                    max_workers=self._num_workers
+                )
         else:
             self._executor = uproot.source.futures.TrivialExecutor()
```
jpivarski, line 56: The decision to use an event loop, a thread pool, or nothing (synchronous/blocking).

I don't see any reason why we should keep the `ThreadPoolExecutor`-based solution around. Maybe there will be some advantage to letting the event loop use multiple threads or each thread has its own event loop, but the main thing that the old `ThreadPoolExecutor`-based solution gets wrong is that it does not send all of the requests immediately.

The switch that falls back on `ThreadPoolExecutor` is `self._fs.async_impl`. If an fsspec backend doesn't implement async, shouldn't we do an event loop anyway? That is, if the function that makes a request makes a request when it is called doesn't return an output until the response is in, couldn't we wrap that up in a future manually?

None of the above should delay this PR, though. It's a next step.

src/uproot/source/fsspec.py
```diff
@@ -132,10 +143,13 @@ def chunks(
...
+        )
+        cat_file = self._fs._cat_file if use_async else self._fs.cat_file
         for start, stop in ranges:
-            future = self._executor.submit(
-                self._fs.cat_file, self._file_path, start, stop
-            )
+            future = self._executor.submit(cat_file, self._file_path, start, stop)
             chunk = uproot.source.chunk.Chunk(self, start, stop, future)
             future.add_done_callback(uproot.source.chunk.notifier(chunk, notifications))
             chunks.append(chunk)
```
jpivarski, line 155: Submitting all of the requests on what may be the event loop, may be the `ThreadPoolExecutor`, and may be a `TrivialExecutor`, wrapped up in something that has the interface of an executor.

The extra indirection might not be necessary. If we're going to switch between three (or two?) possible methods, why not do it right here?

Again, that's just something to think about.

src/uproot/source/futures.py
```diff
@@ -469,3 +472,44 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.shutdown()
+
+    @property
+    def loop(self) -> asyncio.AbstractEventLoop:
+        return self._loop
+
+    def submit(self, coroutine, *args) -> asyncio.Future:
+        coroutine_object = coroutine(*args)
+        return asyncio.run_coroutine_threadsafe(coroutine_object, self._loop)
```
jpivarski, line 515: Implementation of the `LoopExecutor`. This is the part that had the possible race condition. I don't know asyncio well enough to be sure that it's resolved, but CI was sensitive to it, so when everything passes, we'll take that as a sign that it's likely okay.

jpivarski (approved):
I've highlighted and commented upon the parts that are relevant for this PR (distinguishing them from the typing improvements).

I approve the PR; it should get merged once the test issues are resolved.

src/uproot/source/fsspec.py
```diff
@@ -38,9 +38,20 @@ def __init__(self, file_path, **options):
...
+            if self._fs.async_impl:
+                self._executor = uproot.source.futures.LoopExecutor()
+
+                # Bind the loop to the filesystem
+                async def make_fs():
+                    return fsspec.filesystem(
+                        protocol=self._fs.protocol, loop=self._executor.loop
+                    )
+
+                self._fs = self._executor.submit(make_fs).result()
```
lobis, line 50: This is some of the problematic code. This fix was added by @nsmith- and fixed one of the race conditions. I am not sure if this is still the problem though.

lobis:
This is the log when running:

`pytest tests/test_0692_fsspec.py::test_fsspec_chunks`

This only fails on ubuntu.

```
========================================================================= test session starts =========================================================================
platform linux -- Python 3.10.12, pytest-7.4.2, pluggy-1.3.0
rootdir: /uproot
configfile: pyproject.toml
plugins: timeout-2.2.0, rerunfailures-12.0
timeout: 600.0s
timeout method: signal
timeout func_only: False
collected 1 item                                                                                                                                                      

tests/test_0692_fsspec.py .                                                                                                                                     [100%]

========================================================================== 1 passed in 2.71s ==========================================================================
Fatal error on SSL transport
protocol: <asyncio.sslproto.SSLProtocol object at 0xffff79695f30>
transport: <_SelectorSocketTransport closing fd=18>
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 924, in write
    n = self._sock.send(data)
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/asyncio/sslproto.py", line 690, in _process_write_backlog
    self._transport.write(chunk)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 930, in write
    self._fatal_error(exc, 'Fatal write error on socket transport')
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 725, in _fatal_error
    self._force_close(exc)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 737, in _force_close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 753, in call_soon
    self._check_closed()
  File "/usr/lib/python3.10/asyncio/base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Fatal error on SSL transport
protocol: <asyncio.sslproto.SSLProtocol object at 0xffff796955d0>
transport: <_SelectorSocketTransport closing fd=19>
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 924, in write
    n = self._sock.send(data)
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/asyncio/sslproto.py", line 690, in _process_write_backlog
    self._transport.write(chunk)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 930, in write
    self._fatal_error(exc, 'Fatal write error on socket transport')
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 725, in _fatal_error
    self._force_close(exc)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 737, in _force_close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 753, in call_soon
    self._check_closed()
  File "/usr/lib/python3.10/asyncio/base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Fatal error on SSL transport
protocol: <asyncio.sslproto.SSLProtocol object at 0xffff79696cb0>
transport: <_SelectorSocketTransport closing fd=20>
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 924, in write
    n = self._sock.send(data)
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/asyncio/sslproto.py", line 690, in _process_write_backlog
    self._transport.write(chunk)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 930, in write
    self._fatal_error(exc, 'Fatal write error on socket transport')
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 725, in _fatal_error
    self._force_close(exc)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 737, in _force_close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 753, in call_soon
    self._check_closed()
  File "/usr/lib/python3.10/asyncio/base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Fatal error on SSL transport
protocol: <asyncio.sslproto.SSLProtocol object at 0xffff796961d0>
transport: <_SelectorSocketTransport closing fd=21>
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 924, in write
    n = self._sock.send(data)
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/asyncio/sslproto.py", line 690, in _process_write_backlog
    self._transport.write(chunk)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 930, in write
    self._fatal_error(exc, 'Fatal write error on socket transport')
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 725, in _fatal_error
    self._force_close(exc)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 737, in _force_close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 753, in call_soon
    self._check_closed()
  File "/usr/lib/python3.10/asyncio/base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Fatal error on SSL transport
protocol: <asyncio.sslproto.SSLProtocol object at 0xffff79697a00>
transport: <_SelectorSocketTransport closing fd=22>
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 924, in write
    n = self._sock.send(data)
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/asyncio/sslproto.py", line 690, in _process_write_backlog
    self._transport.write(chunk)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 930, in write
    self._fatal_error(exc, 'Fatal write error on socket transport')
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 725, in _fatal_error
    self._force_close(exc)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 737, in _force_close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 753, in call_soon
    self._check_closed()
  File "/usr/lib/python3.10/asyncio/base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Fatal error on SSL transport
protocol: <asyncio.sslproto.SSLProtocol object at 0xffff796953c0>
transport: <_SelectorSocketTransport closing fd=17>
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 924, in write
    n = self._sock.send(data)
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/asyncio/sslproto.py", line 690, in _process_write_backlog
    self._transport.write(chunk)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 930, in write
    self._fatal_error(exc, 'Fatal write error on socket transport')
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 725, in _fatal_error
    self._force_close(exc)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 737, in _force_close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 753, in call_soon
    self._check_closed()
  File "/usr/lib/python3.10/asyncio/base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed

```

src/uproot/source/fsspec.py
nsmith-: To be clear, the issue that was fixed was not a race condition but a initialization routine accessing some data that is only (apparently) valid when accessed from the thread running the event loop. So we schedule the initialization in that loop and wait for it here.

src/uproot/source/fsspec.py
nsmith-: If the backend doesn't implement async, then the synchronous interface implies that _some thread_ will block waiting for a request. So if we want any concurrency, we need multiple threads. Event loops will not work any better than a for loop with code that is blocking on IO.

src/uproot/source/fsspec.py
```diff
@@ -86,7 +86,10 @@ def __enter__(self):
 
     def __exit__(self, exception_type, exception_value, traceback):
         self._fh = None
-        self._executor.shutdown()
+        # TODO: proper cleanup. Shutting down the loop executor causes problems
+        # self._executor.shutdown()
+        if not isinstance(self._executor, uproot.source.futures.LoopExecutor):
+            self._executor.shutdown()
         self._file.__exit__(exception_type, exception_value, traceback)
```
nsmith-, line 93: I think the file exit should be before the executor shuts down

lobis: I tried it but it doesn't make a difference.

lobis:
[c615eca](https://github.com/scikit-hep/uproot5/pull/992/commits/c615eca1c30c7b9bd1a01ff16a3626df6a3c66d3) fixes the issue with the `fsspec` chunks test (but now `xrootd` `fsspec` tests are failing...).

I think the problem has something to do with the shutting down of the loop executor. It's lifetime should in principle be bound to the source, but there is something wrong when doing the shutdown.

src/uproot/source/futures.py
```diff
@@ -469,3 +472,45 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+
+
+class LoopExecutor:
+    def __repr__(self):
+        return f"<LoopExecutor at 0x{id(self):012x}>"
+
+    def __init__(self):
+        self._loop = asyncio.new_event_loop()
+        # TODO: remove daemon=True (or not?)
+        self._thread = threading.Thread(target=self._run, daemon=True)
```
agoose77, line 484: I don't think we want a `daemon` here — we want fsspec to be able to gracefully close  its resources if possible.

src/uproot/source/futures.py
```diff
@@ -469,3 +472,45 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+        self._loop.call_soon_threadsafe(self._loop.stop)
+        self._thread.join()
+
+    def _run(self):
+        asyncio.set_event_loop(self._loop)
+        try:
+            self._loop.run_forever()
+        finally:
+            self._loop.run_until_complete(self._loop.shutdown_asyncgens())
+            self._loop.close()
```
agoose77, line 501: We probably need to run cleanups here too, *and* shield these from cancellation

```python
    # Find all running tasks:
    pending = asyncio.all_tasks()

    # Run loop until tasks done:
    loop.run_until_complete(
        asyncio.gather(*[
            asyncio.shield(t) for t in pending
            ]
        )
    )
```

src/uproot/source/futures.py
nsmith-: Yeah I agree, it should be possible to do this cleanup correctly

src/uproot/source/futures.py
nsmith-: This is mostly my fault. I don't understand why there isn't a canned "loop-in-a-thread" in the stdlib or elsewhere.

agoose77: @nsmith I wouldn't ascribe fault — asyncio is a bit of an ugly beast (I use `trio` in my personal projects these days), but it's the one we're stuck with.

The intention with `asyncio` is that you only really have one loop. What we're doing is a bit unconventional, but clearly necessary to handle a blocking high level UX.

agoose77: We effectively want this: https://github.com/python/cpython/blob/3.8/Lib/asyncio/runners.py#L8-L73

nsmith-: Ah, yes! The previous error was an unhandled exception, and I would not be surprised if we have another one.

lobis:
Looks like something "out of our control" (fsspec?) is submitting tasks to the loop after it has finished processing the requested tasks (get the chunks). This fails if the loop is not running (we shut it down because we don't expect to do any more work).

Why this happens (and why it only happens in some OS...) is a mistery to me.

Maybe the pattern of spawning a thread bound to a source has some fundamental flaw? Maybe we should use a single loop for everything async-related, so making the `LoopExecutor` a singleton? IF (big if) we want to transition uproot into an async code with sync wrappers this pattern would be more suited for it, as it's one step closer to having a single loop blocking the main thread.

src/uproot/source/fsspec.py
lobis: I wish I had read this comment before ba7a6bf702e6332fbeb64c6822a29737055601aa 🤦🏻‍♂️ 😄

lobis:
For anyone following up: I think I know what the problem is: something is trying to shutdown the executor while the intended tasks are running, this should not happen. Now I only need to find it.

agoose77:
> For anyone following up: I think I know what the problem is: something is trying to shutdown the executor while the intended tasks are running, this should not happen. Now I only need to find it.

This makes sense - I was curious as to whether this was happening in https://github.com/scikit-hep/uproot5/pull/992#issuecomment-1766819203

However, we should be blocking on the future that corresponds with the loop result, so I am surprised. I can take another look at that today

lobis:
I realised that fsspec provides and manages it's own loop - exactly what we want, so I trashed the previous implementation of the `LoopExecutor` and just used this one instead. I updated the PR description (https://github.com/scikit-hep/uproot5/pull/992#issue-1943634061).

src/uproot/source/fsspec.py
```diff
@@ -155,3 +163,16 @@ def closed(self) -> bool:
         otherwise.
         """
         return False
+
+
+class FSSpecLoopExecutor(uproot.source.futures.Executor):
+    def __init__(self, loop: asyncio.AbstractEventLoop):
+        self.loop = loop
+
+    def submit(self, coroutine, *args) -> asyncio.Future:
```
nsmith-, line 195: This is actually returning a `concurrent.futures.Future` which is subtly different than an asyncio Future. I guess mypy would flag this if we were actively testing with it.

nsmith- (approved):
FSSpec provides a loop for all of its IO in a daemon thread that lasts the duration of the program:
https://github.com/fsspec/filesystem_spec/blob/3fbb6fee33b46cccb015607630843dea049d3243/fsspec/asyn.py#L145-L149
It is constructed on first use in a threadsafe manner. I think this is a reasonable solution, though I suppose it should implement shutting down async generators and waiting for pending tasks as is done in the [standard library](https://github.com/python/cpython/blob/3.8/Lib/asyncio/runners.py#L8-L73) `asyncio.run`. Perhaps @martindurant knows if this is purposely not implemented.

PR #994, opened 2023-10-18
fix: remove unused hist import in test_0965

GaetanLepage:
Test 0965 fails:
```
==================================== ERRORS ====================================_____ ERROR collecting tests/test_0965-inverted-axes-variances-hist-888.py _____ImportError while importing test module '/build/source/tests/test_0965-inverted-axes-variances-hist-888.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/nix/store/pzf6dnxg8gf04xazzjdwarm7s03cbrgz-python3-3.10.12/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_0965-inverted-axes-variances-hist-888.py:5: in <module>
    import hist
E   ModuleNotFoundError: No module named 'hist'
```

It seems that the other tests relying on `hist` import it using `hist = pytest.importorskip("hist")`.
I did the same in this problematic test.

### DIFF
--- a/tests/test_0965-inverted-axes-variances-hist-888.py
+++ b/tests/test_0965-inverted-axes-variances-hist-888.py
@@ -2,10 +2,10 @@
 
 import pytest
 import numpy
-import hist
 import uproot
 import skhep_testdata
 
+pytest.importorskip("hist")
 ROOT = pytest.importorskip("ROOT")
 
 


tests/test_0965-inverted-axes-variances-hist-888.py
```diff
@@ -2,10 +2,10 @@
 
 import pytest
 import numpy
-import hist
 import uproot
 import skhep_testdata
 
+hist = pytest.importorskip("hist")
```
ioanaif, line 8: `hist` is not used in these tests, thus it can be simply removed as an import

jpivarski:
```suggestion
```

jpivarski (approved):
Thanks for catching this!

I'll merge it when the tests pass.

jpivarski:
Okay, now it's in a good state. The lines where there's a hidden dependency on hist (Uproot would raise `ModuleNotFoundError` suggesting that you install it) are

https://github.com/scikit-hep/uproot5/blob/e4b4b80cea06537108cf49555b4bb631b64bad15/tests/test_0965-inverted-axes-variances-hist-888.py#L25

and

https://github.com/scikit-hep/uproot5/blob/e4b4b80cea06537108cf49555b4bb631b64bad15/tests/test_0965-inverted-axes-variances-hist-888.py#L46

Now that the file has

https://github.com/scikit-hep/uproot5/blob/e4b4b80cea06537108cf49555b4bb631b64bad15/tests/test_0965-inverted-axes-variances-hist-888.py#L8

I'll enable "auto-merge (squash)."

jpivarski:
@all-contributors please add @GaetanLepage for test

Thanks!

PR #996, opened 2023-10-18
chore: add types to most of the `uproot.source` module

lobis:
Add types to most of the sources module.

Only added types where I was sure of the type.

I left things such as urls, file paths, etc. untyped instead of typing them as `str` since user may pass a `pathlib` path or similar.

### DIFF
--- a/src/uproot/source/chunk.py
+++ b/src/uproot/source/chunk.py
@@ -60,7 +60,9 @@ class Source:
         :doc:`uproot.source.chunk.Chunk`.
         """
 
-    def chunks(self, ranges, notifications: queue.Queue) -> list[Chunk]:
+    def chunks(
+        self, ranges: list[(int, int)], notifications: queue.Queue
+    ) -> list[Chunk]:
         """
         Args:
             ranges (list of (int, int) 2-tuples): Intervals to fetch
@@ -164,7 +166,9 @@ class MultithreadedSource(Source):
         self._executor.submit(future)
         return chunk
 
-    def chunks(self, ranges, notifications: queue.Queue) -> list[Chunk]:
+    def chunks(
+        self, ranges: list[(int, int)], notifications: queue.Queue
+    ) -> list[Chunk]:
         self._num_requests += 1
         self._num_requested_chunks += len(ranges)
         self._num_requested_bytes += sum(stop - start for start, stop in ranges)
--- a/src/uproot/source/chunk.py
+++ b/src/uproot/source/chunk.py
@@ -31,7 +31,7 @@ class Resource:
     """
 
     @property
-    def file_path(self):
+    def file_path(self) -> str:
         """
         A path to the file (or URL).
         """
@@ -92,7 +92,7 @@ class Source:
         """
 
     @property
-    def file_path(self):
+    def file_path(self) -> str:
         """
         A path to the file (or URL).
         """
--- a/src/uproot/source/chunk.py
+++ b/src/uproot/source/chunk.py
@@ -464,7 +464,7 @@ outside expected range {self._start}:{self._stop} for this Chunk""",
             )
 
 
-def notifier(chunk: Chunk, notifications: queue.Queue) -> callable:
+def notifier(chunk: Chunk, notifications: queue.Queue):
     """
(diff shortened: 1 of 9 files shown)


src/uproot/source/chunk.py
```diff
@@ -240,7 +244,7 @@ class Chunk:
     _dtype = numpy.dtype(numpy.uint8)
 
     @classmethod
-    def wrap(cls, source, data, start=0):
+    def wrap(cls, source: Source, data: numpy.ndarray, start: int = 0):
```
jpivarski, line 247: I think there might be cases in which `wrap` is used to wrap `bytes`. These `memoryview` objects are used equivalently.

```suggestion
    def wrap(cls, source: Source, data: numpy.ndarray | bytes, start: int = 0):
```

src/uproot/source/file.py
```diff
@@ -69,7 +69,9 @@ def get(self, start, stop):
         return self._file.read(stop - start)
 
     @staticmethod
-    def future(source, start, stop):
+    def future(
+        source: uproot.source.file.MultithreadedFileSource, start: int, stop: int
+    ):
```
jpivarski, line 74: Is it always a `MultithreadedFileSource`? It might always be some kind of file Source.

```suggestion
    def future(
        source: uproot.source.chunk.Source, start: int, stop: int
    ):
```

src/uproot/source/http.py
```diff
@@ -351,7 +353,7 @@ def task(resource):
     )
     _content_range = re.compile(b"Content-Range: bytes ([0-9]+-[0-9]+)", re.I)
 
-    def is_multipart_supported(self, ranges, response):
+    def is_multipart_supported(self, ranges: list[(int, int)], response) -> bool:
```
jpivarski, line 356: It's an [HTTPResponse](https://docs.python.org/3/library/http.client.html#httpresponse-objects).

```suggestion
    def is_multipart_supported(self, ranges: list[(int, int)], response: http.client.HTTPResponse) -> bool:
```

src/uproot/source/http.py
```diff
@@ -383,7 +391,14 @@ def handle_no_multipart(self, source, ranges, futures, results):
...
 
-    def handle_multipart(self, source, futures, results, response, ranges):
+    def handle_multipart(
+        self,
+        source: uproot.source.chunk.Source,
+        futures,
+        results,
+        response,
+        ranges: list[(int, int)],
+    ):
```
jpivarski, line 401: The types of `futures` and `results` can be determined from the way this method is called (only in one place). The `results` is (if I remember right) a length-1 list that gets updated in-place because Pythons 2 and 3 differed in their uses of `global` and `nonlocal`, so we got the same effect—a mutable reference—by wrapping the object to change in a list.

```suggestion
    def handle_multipart(
        self,
        source: uproot.source.chunk.Source,
        futures,
        results,
        response: http.client.HTTPResponse,
        ranges: list[(int, int)],
    ):
```

src/uproot/source/http.py
```diff
@@ -368,7 +370,13 @@ def is_multipart_supported(self, ranges, response):
...
             return True
 
-    def handle_no_multipart(self, source, ranges, futures, results):
+    def handle_no_multipart(
+        self,
+        source: uproot.source.chunk.Source,
+        ranges: list[(int, int)],
+        futures,
+        results,
+    ):
```
jpivarski, line 379: See `handle_multipart` for comments on `futures` and `results`.

jpivarski (approved):
Great, thanks! You can merge this as-is. My suggested commits below are just taking the typing further, which is optional.

All of the `int` argument types are a little dubious: these arguments sometimes take `np.int64` and similar, depending on where the integer came from, and that's technically not an `int`. For runtime type-checking, I use

```python
isinstance(x, numbers.Integral)
```

which takes NumPy integers as well as Python `int`, but I think that filling up these files with `: numbers.Integral` and especially `: list[(numbers.Integral, numbers.Integral)]` would obscure it.

agoose77:
For posterity, I don't think mypy understands the integral type. Also, there's the os.PathLike interface for non-str/bytes types, so we could define that in a local types module.

PR #997, opened 2023-10-18
fix: cache rendered Form and ImplementedFormMappingInfo

lgray:
Restrict more stateful parts of code to client side of dask.

This fixes stickiness of configuration data for NanoEvents since that interface is class driven.
Without caching the rendered form setting cross references or branch behavior would not make it to remote dask workers!

@agoose77 If there's a better way to do this please alter the code!

Once this is reviewed and merged a prompt new release would be appreciated!

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -889,6 +889,7 @@ class UprootReadMixin:
     base_form: Form
     common_keys: frozenset[str]
     interp_options: dict[str, Any]
+    _high_level_form_and_info: tuple[Form, ImplementsFormMappingInfo]
 
     def read_tree(self, tree: HasBranches, start: int, stop: int) -> AwkArray:
         assert start <= stop
@@ -898,7 +899,7 @@ class UprootReadMixin:
         awkward = uproot.extras.awkward()
         nplike = Numpy.instance()
 
-        form, form_info = self.form_mapping(self.base_form)
+        form, form_info = self._high_level_form_and_info
 
         # The remap implementation should correctly populate the generated
         # buffer mapping in __call__, such that the high-level form can be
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
My read of this is that the remapped Form is no longer recomputed on each `read_tree` and `mock`, but once and for all when the `UprootReadMixin` is constructed.

I verified that there are exactly two implementations of `UprootReadMixin`, and both of them properly set this new attribute.

Without your comment, I would have assumed that the motivation is performance (is computing the remapped Form costly?), but you say that it's to solve a cross-referencing issue. I can believe that, but the issue would have to involve more code than I see here. I'm going to assume that @agoose77 knows about this and can review it with that in mind.

If it's about something sent to Dask workers, the new attribute and its data type (Forms) are perfectly pickleable, so this can have the effect of determining the remapped Form on a Dask client or scheduler (not sure which) and sending it to the Dask worker, so that the Dask worker does not try to recompute it. Maybe that was the crux of the issue.

This should get a review from @agoose77 before merging. If you need the fix right away, I can make a new release as soon as it's merged.

lgray:
This is mostly stemming from how nanoevents is implemented. The schemas that eventually create the rendered forms are provided to the machinery as classes, not instances of classes, so any client-side configuration does not propagate to the worker side (without significant re-engineering of nanoevents itself). 

There may be a minor performance gain from not re-executing the same code to build the form thousands of times when on a realistic dataset. The main benefit I see is that this sets a clear firebreak in the interface so that it is clear to people developing schemas and dataformats that anything stateful is client-side *only* and needs to be provided to dask workers via the form and behaviors *alone*.

agoose77:
So, there are indeed a couple of related but distinct considerations to be made here:
1. Performance; form remapping is performed _for each chunk_ for each file.
2. Correctness; schemas are not pickle-aware, so anything performed on the client w.r.t schemas doesn't behave properly

I am guessing that we could customise the schema pickler using a metaclass, but that's probably not the right solution here given (1).

I wasn't worried about the performance of remapping, and expect it to be idempotent, but perhaps that first assumption was too strong.

I will flesh out a PR to fix this. Although we're in "this is very niche" land, I'll try not to break existing code.

lgray:
FWIW this appears to break no existing code at least for what it's concerned (already tested it out on coffea), I'm sure there are some other odds and ends to deal with.

PR #998, opened 2023-10-18
fix: cache form remapping to avoid per-chunk workload

agoose77:
Replaces #997 with a PR that exclusively caches the remapped form.

In the original dask-awkward refactoring PR, I treated form remapping as a cheap operation that could be performed multiple times. However, it appears that this was too strong an assumption, and incidentally brings other consequences when performed at worker time, not just at scheduler time.

This PR caches the form remapping, changing our private interface to store the remapping information rather than its factory.

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -885,8 +885,9 @@ T = TypeVar("T")
 
 
 class UprootReadMixin:
-    form_mapping: ImplementsFormMapping
     base_form: Form
+    expected_form: Form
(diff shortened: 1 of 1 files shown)


agoose77:
@lgray can you check coffea against this PR to confirm no regressions?

lgray:
@agoose77 yep looks good on my side.

jpivarski:
Last touched 12 hours ago, I guess it's ready. I'll update to main to re-run the tests and enable auto-merge.

PR #1001, opened 2023-10-19
feat: update the executor submit interface to take keyword arguments and be compatible with `concurrent.futures.ThreadPoolExecutor`

lobis:
We were passing arguments to the tasks as positional, this makes some of the fsspec sources behave differently because they may have different number / order of arguments.

I also modified the definitions to keep the same as `concurrent.futures.ThreadPoolExecutor`.

I noticied after debugging https://github.com/scikit-hep/uproot5/pull/999 failure.

### DIFF
--- a/src/uproot/source/futures.py
+++ b/src/uproot/source/futures.py
@@ -43,7 +43,7 @@ class Executor(ABC):
         return f"<{self.__class__.__name__} at 0x{id(self):012x}>"
 
     @abstractmethod
-    def submit(self, task, *args):
+    def submit(self, task, /, *args, **kwargs):
         """
         Submit a task to be run in the background and return a Future object
         representing that task.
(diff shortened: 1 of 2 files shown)


src/uproot/source/futures.py
```diff
@@ -249,14 +249,14 @@ def workers(self) -> list[Worker]:
...
-    def submit(self, task, *args):
+    def submit(self, task, /, *args, **kwargs):
         """
         Pass the ``task`` and ``args`` onto the workers'
         :ref:`uproot.source.futures.Worker.work_queue` as a
         :doc:`uproot.source.futures.Future` so that it will be executed when
         one is available.
         """
-        future = Future(task, args)
+        future = Future(task, *args, **kwargs)
```
lobis, line 259: I think this may cause some problems.

lobis: It did, now it's fixed. Perhaps there is a better way to interface this.

jpivarski (approved):
This is good—the Python 3 Executor protocol has a `/` in the argument list, though this wasn't possible in Python 2.

This is a net improvement and can't break anything (except possible users of `uproot.source.futures.*Executor` outside of Uproot, though they should be using the Python standard library instead).

I'll merge it.

Issue #1006, opened 2023-10-24
File-object separator colon parsing PR #976 must be fixed before the next release


jpivarski:
In #976, the parsing that determines where a colon is between a file name and an object path was changed incorrectly: a basic (and important) use-case fails:

```python
>>> import uproot
>>> tree = uproot.open("00376186-543E-E311-8D30-002618943857.root:Events")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jpivarski/irishep/uproot5/src/uproot/reading.py", line 143, in open
    file = ReadOnlyFile(
  File "/home/jpivarski/irishep/uproot5/src/uproot/reading.py", line 584, in __init__
    Source, file_path = uproot._util.file_path_to_source_class(
  File "/home/jpivarski/irishep/uproot5/src/uproot/_util.py", line 508, in file_path_to_source_class
    raise ValueError(f"URI scheme not recognized: {file_path}")
ValueError: URI scheme not recognized: 00376186-543E-E311-8D30-002618943857.root:Events
```

but

```python
>>> tree = uproot.open("/tmp/00376186-543E-E311-8D30-002618943857.root:Events")
```

passes. The failing case is a relative local path; the successful case is an absolute local path.

I verified that 6b1952ea572471c7b64da3650566d6ce7ff6a4bc (just before #976) passes but fdcbb8e94aea734390ffe6babc9df380e97bf117 (just after #976) fails.

PR #1007, opened 2023-10-24
fix: url and object splitting for local files

lobis:
Fixes https://github.com/scikit-hep/uproot5/issues/1006.

Add additional tests to cover more cases.

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -5,11 +5,14 @@ This module defines utilities for internal use. This is not a public interface
 and may be changed without notice.
 """
 
+from __future__ import annotations
+
 import datetime
 import glob
 import itertools
 import numbers
 import os
+import pathlib
 import platform
 import re
 import warnings
@@ -282,7 +285,7 @@ _remote_schemes = ["ROOT", "S3", "HTTP", "HTTPS"]
 _schemes = ["FILE", *_remote_schemes]
 
 
-def file_object_path_split(path):
+def file_object_path_split(path: str) -> tuple[str, str | None]:
     """
     Split a path with a colon into a file path and an object-in-file path.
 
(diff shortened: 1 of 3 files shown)


src/uproot/_util.py
```diff
@@ -296,13 +299,19 @@ def file_object_path_split(path):
     """
 
     path: str = regularize_path(path)
-    # remove whitespace
     path = path.strip()
 
-    # split url into parts
-    parsed_url = urlparse(path)
+    if "://" in path:
```
jpivarski, line 304: This can be tighter ([URI scheme syntax](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax)):

```python
>>> fsm = re.compile("^[a-zA-Z][a-zA-Z0-9\+\.\-]*://")

>>> fsm.match("http://stuff")
<re.Match object; span=(0, 7), match='http://'>

>>> fsm.match("/tmp/filename.root://stuff")
None

>>> fsm.match("/tmp/filename.root:/subdirectory/stuff")
None
```

Maybe instead of matching schemes as `^[a-zA-Z][a-zA-Z0-9\+\.\-]*`, we should have a short list of supported schemes? Is there a way of asking fsspec what it supports?

Because we can't do anything about this:

```python
>>> fsm.match("filename.root://stuff")
<re.Match object; span=(0, 16), match='filename.root://'>
```

unless we say that a scheme can be `root` but not `filename.root`.

jpivarski: On second thought, there's no good reason for users to put superfluous slashes in the object path. `//one/two` is exactly the same as `one/two` except that it's deliberately asking for trouble. Anyway, if someone runs into it, we can tell them to just remove the unnecessary `//` at the beginning of their object path.

But at least do

```suggestion
    if _uri_scheme.match(path):
```

with

```python
_uri_scheme = re.compile("^[a-zA-Z][a-zA-Z0-9\+\.\-]*://")
```

just above this function.

It's legal to have more colons and slashes inside the object path, and requiring the scheme to satisfy the URI specification would cut down on false positives.

jpivarski (approved):
Sorry about this being so difficult, with so many corner cases. If I could do Uproot 3 over again (or whichever version introduced those !*$&@# colons), I wouldn't have added this feature. But once added, people use it a lot.

Another time I could have killed it was in the Uproot 3 → 4 transition, but users accustomed to the colon thought that Uproot 4 couldn't read the files because the colon wasn't implemented at first: #79.

Just make the one change described below with a test that is sensitive to it and then you can merge this PR.

src/uproot/_util.py
lobis: I agree this should be more robust. I will add the uri regex and merge this as is to fix the bug but I will also make another PR to improve this.

I like the idea of using `fsspec` to get the list of available protocols to match (it is possible to retrieve this). I think the logic should be as follows:

- Try to match agains the available protocols.
- If no match, try to find `://` inside the string. If found throw an exception saying it's not a valid procotol and point the user towards the available protocols (print the list). This would throw on cases such as `file.root://object` treating `file.root` as an invalid schema.
- if no schema found in the string, assume its a local file and prepend the `file://` schema.
- process the fully qualified uri.

There is a subtle thing to keep in mind. I believe uproot currently supports uppercase (or any case) schemas, such as "hTtP" because the comparison is caseless. I believe `fsspec` is case sensitive to the schema (and I have not seen any schema with uppercase). `fsspec.filesystem("http")` is valid while `fsspec.filesystem("httP")` is not. This would technically mean breaking changes to the API but I think it makes sense.

PR #1009, opened 2023-10-25
feat: improve uri scheme parsing with list of available schemes from `fsspec`

lobis:
Use `fsspec` to get the list of supported uri schemes. Direct the user to this list in case of an invalid scheme.

**Changes**: paths such as `file.root://object-with-too-many-slashes` will now throw a `ValueError` saying that the `file.root://` scheme is not in the list of valid schemes.

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -280,9 +280,19 @@ def regularize_path(path):
 _windows_drive_letter_ending = re.compile(r".*\b[A-Za-z]$")
 _windows_absolute_path_pattern = re.compile(r"^[A-Za-z]:[\\/]")
 _windows_absolute_path_pattern_slash = re.compile(r"^[\\/][A-Za-z]:[\\/]")
+
 _remote_schemes = ["root", "s3", "http", "https"]
 _schemes = ["file", *_remote_schemes]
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
I suppose there's no legitimate reason to use `://` in a file path or an object path (both of which use `/+` as a delimiter, and the initial `/` in an object path is optional). Therefore, if someone gets the "Invalid URI scheme" error message, they'll likely recognize what it is and they'll have a recourse to work around it.

Is it possible to get the `_remote_schemes` from fsspec? Just wondering.

Anyway, this PR is good and may be merged.

lobis:
> I suppose there's no legitimate reason to use `://` in a file path or an object path (both of which use `/+` as a delimiter, and the initial `/` in an object path is optional). Therefore, if someone gets the "Invalid URI scheme" error message, they'll likely recognize what it is and they'll have a recourse to work around it.
> 
> Is it possible to get the `_remote_schemes` from fsspec? Just wondering.
> 
> Anyway, this PR is good and may be merged.

Yes, we can get all schemes from `fsspec`. I left the previous code there because `fsspec` is not yet a dependency. Once it is this will be further simplified.

jpivarski:
> I left the previous code there because `fsspec` is not yet a dependency.

Of course! That makes a lot of sense.

PR #1010, opened 2023-10-25
test: local http server for tests

lobis:
Add a local http server pytest fixture. The http server is the python built-in but the request handler comes from https://github.com/danvk/RangeHTTPServer since the python one does not handle range requests. This handler also supports multipart range header requests!

The current implementation serves files from the `skhep_testdata` cache directory and downloads them on demand (only once) via calls to `skhep_testdata.data_path`.

There is an open issue to add this to python: https://github.com/python/cpython/issues/86809.

I replaced most of the previous instances of remote http access by this local server but not all of them.

This PR also enables back some tests that were marked as skip due to network issues.


agoose77:
This particular package isn't super well-maintained (it has a single maintainer who's expressed an interest in handing off to someone else). However, it's also a very simple, self-contained part of the test system. We could probably just vendor it in future, if we need to. So, on the balance of cost-tradeoffs, I think this is a sensible inclusion.

agoose77 (approved):

PR #1012, opened 2023-10-31
test: s3fs pytest unraisable exception

lobis:
This PR doesn't fix the underlying cause of this error in pytest but it makes it go away from the CI.

- Add back previosly skipped `S3Source` tests. There is an issue with the test that checks for exception when the file does not exist: many times this throws a timeout that doesn't go away with retries, I will just add the timeout exceptions to the list of valid exceptions.
- Add the infamous `s3fs` test but only on python 3.11 and above. In older python version we have this mysterious error.
- (named some older test functions that were matching the substring "s3" by mistake and was slightly annoying).

I'm 90% confident that this mysterious error won't cause a problem when using uproot with s3 and fsspec but it's the reason I didn't set it as default in https://github.com/scikit-hep/uproot5/pull/1023. Now that I think about it, I don't think it was a good idea to not set the s3 default to s3fs. It probably works fine and doing so will probably let us know if this is the case, so I will do it (in another PR).


jpivarski (approved):
The diff is quite reasonable, but the list of commits is very long! Were you debugging in CI? (Not that I haven't done it myself...)

Thanks for fixing this. It looks done now, so I'll merge it.

lobis:
> The diff is quite reasonable, but the list of commits is very long! Were you debugging in CI? (Not that I haven't done it myself...)

Yes, before I gave up on fixing the underlying issue I tried many different things.

I'm curious, is there a problem I'm missing with the long list of commits? I'll get squashed, it'll just be a very long commit message, am I right?

jpivarski:
That's right. It will be a long description; the title of the commit will be the title of the PR plus " (#1012)". It's only a problem for mouseover in the GitHub web interface, which shows the long description of the commit you're hovering over.

tests/test_0692_fsspec_reading.py
```diff
@@ -53,13 +54,17 @@ def test_open_fsspec_local():
...
         uproot.source.s3.S3Source,
         None,
     ],
 )
 def test_open_fsspec_s3(handler):
     pytest.importorskip("s3fs")
+    if sys.version_info < (3, 11):
+        pytest.skip(
+            "https://github.com/scikit-hep/uproot5/pull/1012",
+        )
```
nsmith-, line 67: @lobis why specifically versions less than 3.11?

lobis: No particular reason, it just failed on older versions, I never could find out why.

PR #1013, opened 2023-10-31
test: testing sshfs with local ssh server

lobis:
This PR adds an ssh server only to the ubuntu build of the CI to test accessing files over ssh via fsspec.

After testing it I realised that `sshfs` does not implement some interface required by uproot for this to work (`cat_file`) so it raises a `NotImplementedError`. I'm not sure if this is just not implemented and can be implemented or if there is a limitation why this has not been implemented (you cannot request a chunk of a file, you need to request it all?).

We could handle each `fsspec` case separately having a fallback in case some interface is not implemented, but I think it would complicate things too much.

After writing this PR I think it makes sense to raise a more detailed exception if fsspec does not implemented the required interface, so the user can post an issue in the corresponding fsspec repo. `fsspec` probably has a way to list the available methods (or it should!)

### DIFF
--- a/tests/test_0692_fsspec.py
+++ b/tests/test_0692_fsspec.py
@@ -73,7 +73,7 @@ def test_open_fsspec_ssh():
     try:
         user = subprocess.check_output(["whoami"]).strip().decode("ascii")
         host = "localhost"
-        ssh_command = f"ssh {user}@{host} 'echo hello'"
+        ssh_command = ["ssh", f"{user}@{host}", "'echo hello'"]
         result = subprocess.run(
             ssh_command,
             shell=True,
--- a/tests/test_0692_fsspec.py
+++ b/tests/test_0692_fsspec.py
@@ -66,30 +66,36 @@ def test_open_fsspec_s3(handler):
         assert len(data) == 8004
 
 
-@pytest.mark.parametrize("handler", [uproot.source.fsspec.FSSpecSource, None])
-def test_open_fsspec_ssh(handler):
+def test_open_fsspec_ssh():
     pytest.importorskip("sshfs")
 
     # check localhost has ssh access to itself
-    user = subprocess.check_output(["whoami"]).strip().decode("ascii")
-    host = "localhost"
-    ssh_command = f"ssh {user}@{host} 'echo hello'"
-    result = subprocess.run(
-        ssh_command,
-        shell=True,
-        text=True,
-        capture_output=True,
-    )
-    if result.returncode != 0:
-        pytest.skip(f"Cannot run ssh command {ssh_command}: {result.stderr}")
+    try:
+        user = subprocess.check_output(["whoami"]).strip().decode("ascii")
+        host = "localhost"
+        ssh_command = f"ssh {user}@{host} 'echo hello'"
+        result = subprocess.run(
+            ssh_command,
+            shell=True,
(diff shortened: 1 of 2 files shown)


tests/test_0692_fsspec.py
```diff
@@ -65,16 +66,36 @@ def test_open_fsspec_s3(handler):
...
+    try:
+        user = subprocess.check_output(["whoami"]).strip().decode("ascii")
+        host = "localhost"
+        ssh_command = f"ssh {user}@{host} 'echo hello'"
+        result = subprocess.run(
+            ssh_command,
+            shell=True,
+            text=True,
+            capture_output=True,
+        )
```
jpivarski, line 82: Generally, the list-form of subprocess commands are preferred because they're less susceptible to malicious code injections. I think if someone somehow commandeered the `whoami` command to return "`; rm -rf /*`; ", then the list-form would not allow it to run because individual items in the list can't contain unquoted spaces.

```suggestion
        user = subprocess.check_output(["whoami"]).strip().decode("ascii")
        host = "localhost"
        ssh_command = ["ssh", f"{user}@{host}", "'echo hello'"]
        result = subprocess.run(
            ssh_command,
            shell=True,
            text=True,
            capture_output=True,
        )
```

jpivarski (approved):
I hadn't known that fsspec provided an ssh backend, but it could be really useful. (I wonder about the usefulness of the S3 backend, given that S3 is a wrapper around HTTP.)

Does this test run in our CI? It's written in a way that it could run if sshd is installed, but that doesn't mean that it will run.

Okay, it doesn't actually run ([Windows with Python 3.10](https://github.com/scikit-hep/uproot5/actions/runs/6708273686/job/18229380987?pr=1013)):

```
SKIPPED [1] tests\test_0692_fsspec.py:87: ssh access to localhost failed with ssh access to localhost failed with ssh: connect to host localhost port 22: Connection refused
```

or in [Ubuntu with Python 3.11](https://github.com/scikit-hep/uproot5/actions/runs/6708273686/job/18229384573?pr=1013):

```
SKIPPED [1] tests/test_0692_fsspec.py:98: sshfs needs to implement some interface
```

At least it is _possible_ for this test to fail, if the attempt raised any exception other than NotImplementedError. (It would be strange for a test to not be able to fail.)

If you've tested this on your computer, you have sshd, and it passes, that's good enough for me. Even if you haven't, this is fine to merge as-is. If there's a problem with it that we discover later, then we'll deal with it then.

lobis:
> If you've tested this on your computer, you have sshd, and it passes, that's good enough for me. Even if you haven't, this is fine to merge as-is. If there's a problem with it that we discover later, then we'll deal with it then.

No, I actually never tested this. After running the test I saw that `sshfs` does not implement `cat_file` interface (required to request a chunk of the file), so it's not possible to use uproot to read ssh files with fsspec at this time.

jpivarski:
So it will always fail with NotImplementedError until they (or we) add that feature. That's okay, it's a placeholder, and it will alert us to when they do add it (if it fails).

tests/test_0692_fsspec.py
lobis: Wow I didn't know that, I always wondered why the recommended way was using a list... now I have an answer (and unlocked a new fear!)

PR #1014, opened 2023-11-02
test: use `paramiko` for ssh instead of `sshfs`

lobis:
After https://github.com/scikit-hep/uproot5/pull/1013 I opened an issue in fsspec https://github.com/fsspec/filesystem_spec/issues/1411 to discuss the issue of not having `cat_files` as part of the interface for the ssh backend.

I realised that I was using the wrong package: `sshfs` which is a third party implementation of an ssh backend (to my defense `s3fs` is the name for the `fsspec` `s3` backend...). `fsspec` already has an ssh backend that can be used if the `paramiko` package is installed. And it implements the `cat_file` interface so it works!

This PR updates the tests to reflect this (now the tests actually access the file over ssh).

**Minor feature**:

I noticed that if the fsspec handler was not being explicitly specified the ssh would faild because it would default to the old sources. I modified the behaviour so that instead of failing when an scheme is not recognized by the old sources, it will now use fsspec instead (if possible). When fsspec becomes default this can be removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ test = [
   "fsspec",
   "fsspec-xrootd",
   "s3fs; python_version<\"3.12\"",  # asyncio not available
-  "sshfs; python_version<\"3.12\"",  # asyncio not available
+  "paramiko",
   "pytest>=6",
   "pytest-timeout",
   "pytest-rerunfailures",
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
This is great! Also, I noticed that these tests did run (and passed):

```
tests/test_0692_fsspec.py::test_open_fsspec_ssh[FSSpecSource] PASSED     [ 85%]
tests/test_0692_fsspec.py::test_open_fsspec_ssh[None] PASSED             [ 85%]
```

This is ready to merge and it adds a user-visible feature, not pure refactoring.

PR #1015, opened 2023-11-02
test: cover more fsspec backends

lobis:
Add test for more fsspec backends: memory, zip, tar.

I also modifed the github uri test to run but skip if the server returns 403 error (too many requests) this way it's tested some of the time.

Probably nobody will ever use uproot to open `.root.zip` files but in any case it's a good test to check the parsing of complex uris.

From adding this tests I noticed that the `uproot._util.file_object_path_split` fails when chaining fsspec protocols on windows paths. I modified the function so it would work (and added additional tests).


jpivarski (commented):
> I also modifed the github uri test to run but skip if the server returns 403 error (too many requests) this way it's tested some of the time.

:heavy_check_mark:

> Probably nobody will ever use uproot to open `.root.zip` files but in any case it's a good test to check the parsing of complex uris.

Actually, people often zip their ROOT files to upload them to GitHub Issues, which rejects file extensions that it doesn't recognize. So this saves one step in testing!

> From adding this tests I noticed that the `uproot._util.file_object_path_split` fails when chaining fsspec protocols on windows paths. I modified the function so it would work (and added additional tests).

:heavy_check_mark:

The `memory://` scheme could be used with memory-mapped files, replacing `MemmapSource`. Then again, I can't think of any reason why that's better than just using `MemmapSource`...

But okay!

This PR is ready to be merged, as soon as you're ready.

PR #1016, opened 2023-11-06
feat: basic fsspec writing

lobis:
This PR adds writing of root files via fsspec which allows writing files to some interesting backends such as ssh, memory, s3, etc. Some are not yet supported such as http(s). In principle all the fsspec backends that support "wb" mode should work fine.

If the backend requires some special option, it can be passed in the `uproot.recreate` call such as `uproot.recreate("ssh:///tmp/file.root", host="localhost", user="user")`.

This is a basic implementation. File is writen on close.

The implementation is as follows:
- If the path to write is a file-like object, behaviour is as before.
- If it is a string, it will check if it has some scheme. If it doesn't behavour is as before.
- If it does have some scheme, it will use fsspec to create a file-like object and produce a sink from it, in a similar way to passing a file-like object to the `uproot.recreate` method. The sink will hold a reference to the object and is responsible for closing it.

Minor features:
- When writing to a local path, the parent directories of the file will be created if they don't exist instead of throwing an exception.

As mentioned this is a basic writer using fsspec. When PR becomes a requirement I will refactor the sink class and the surrounding code to make this cleaner.


src/uproot/sink/file.py
```diff
@@ -9,11 +9,26 @@
...
+    return (
+        callable(getattr(obj, "read", None))
+        and callable(getattr(obj, "write", None))
+        and callable(getattr(obj, "seek", None))
+        and callable(getattr(obj, "tell", None))
+        and callable(getattr(obj, "flush", None))
+        # TODO: what was the motivation behind the checks below? (need to remove them for some fsspec backends)
+        # and (not hasattr(obj, "readable") or obj.readable())
+        # and (not hasattr(obj, "writable") or obj.writable())
+        # and (not hasattr(obj, "seekable") or obj.seekable())
```
lobis, line 28: @jpivarski do you recall why these lines were added? Is it okay to remove them? Some of the fsspec file-like objects (such as the ssh one) implement these interfaces so this wouldn't recognize it as a file-like object.

src/uproot/writing/writable.py
```diff
@@ -36,14 +39,11 @@
 from uproot._util import no_filter, no_rename
 
 
-def create(file_path, **options):
+def create(file_path: str | IO, **options):
```
lobis, line 42: I'm not sure if this annotations make sense. I think its confusing that the variable is called `file_path` but it also supports file-like objects (it took me some time to realise that I could just pass file-like objects).

I am in favour of renaming this variable to something like `file_path_or_object` or atleast anotate it to make it clear, but I'm not sure if this is the correct way to annotate it.

lobis:
TODO: in order to correctly close the handler in all circumstances, instead of passing the fsspec file handler I should pass the "open file" object as if it was a path and open and close it inside the sink.

tests/test_0692_fsspec_writing.py
```diff
@@ -0,0 +1,107 @@
...
+
+        with uproot.open(uri) as f:
+            assert f["tree"]["x"].array().tolist() == [1, 2, 3]
+
+
+@pytest.mark.parametrize(
+    "scheme",
+    [
+        "",
+        "file://",
```
nsmith-, line 51: Can we add here also `simplecache::file://` per https://filesystem-spec.readthedocs.io/en/latest/features.html#remote-write-caching ? I expect this is the most common usage (e.g. `simplecache::s3://mybucket/myfile` or `simplecache::root://storage/path/to/file.root`)

nsmith- (commented):
Since we know uproot will rewrite many parts of the file as it progresses, I wonder if it makes sense to universally append a `simplecache::` to any protocol that isn't already a local one?

tests/test_0692_fsspec_writing.py
jpivarski: We talked about that in person yesterday. I don't think it's a good idea for Uproot to be smart and do the simplecache automatically. I can't think of a reason why someone would want to _not_ do simplecache, except perhaps to show how bad it is (as a line item in a performance table). But it would be a bad surprise if `root://` actually meant `samplecache::root://` (somebody gets surprised because they don't have 2 GB of disk space when they thought they were writing remotely), and if this automatically happens, there would be no way to get plain `root://` (for whatever reason they think they need it).

I think the best way to do this is for Uproot to naively do what is asked of it, but include the `simplecache::` prefix in all of the documentation. Most people will copy-paste.

lobis: > Can we add here also simplecache::file://

I did not add it on the test because it doesn't currently work due to how path parsing is done. I will rewrite this to use fsspec for path parsing once fsspec is a hard requirement. (otherwise it works).

nsmith-: Ok, so users are coached to use it, makes sense.
As for the test, this will need to wait until a future PR then?

lobis: I will do some modifications so that it works with this PR because I think it's important that it works.

After trying to do the parsing with fsspec, I think it's still to hard to cover all the cases (due to the path:object syntax), I'm not sure it's even possible because you an have things such as: `simplecache::zip://insidezip://file///localfile.zip` (and add some other procotols with ports such as http to the mix...).

jpivarski: Let's just not allow colons inside of the `object_path`. So it becomes a search for at most one colon, _not_ in a context that could be interpreted as a URI scheme, `::` namespace, or Windows drive letter.

lobis: I don't think we are currently supporting colons inside the object path.

I just made some changes to update the split so it works with url chains but It's not going to support the full capabilities of fsspec. It's fine for this PR but I'll think of some solution.

Right now I like this solution (not 100% sure it's going to work though):

- Uproot can only be used to open files with `.root` extension.
- Object is specified after the root file in the urlpath (not necessarily at the end). For example: `            "simplecache::zip://uproot-issue121.root:Events/MET_pt::file:///tmp/pytest-of-runner/pytest-0/test_fsspec_zip0/uproot-issue121.root.zip"`. In this case the object is `Events/MET_pt`.
- Find the object by searching for a filename ending in `.root` followed by single `:`. (Will this be robust enough?).

jpivarski: We can't assume that all ROOT files end with a `.root` extension, but we _could_ limit the applicability of the colon-parsing to just those files that do. (The colon-parsing is supposed to be a convenience. If you have a weird situation—a ROOT file not ending in `.root`—you don't deserve a convenience!)

The rule for what we do has to be simple enough to communicate and hopefully guess or intuit. Although the rules are complicated now, they're made that way to correspond to intuition. But how about the following?

1. If the string contains `.root:`, the colon of the _last_ `.root:` is used to split between URI-path and object-path.
2. Otherwise, the whole string is interpreted as a URI-path, every time.

This breaking change would have to go in when the minor version changed. Ideally, we would need to have warned users about this ahead of time, but I don't see a way to do that.

jpivarski: The above would not remove the existing rules that `pathlib.Path` is entirely interpreted as a URI-path, every time, and the `{"uri-path": "object-path"}` syntax would still be recognized.

lobis: I agree with those rules, they are simple enough and should be robust (we'll see if they can handle all cases, thankfully you cannot name a windows drive ".root"...).

- Find `.root` followed by single `:` (but not by `::`!)
- There shouldn't be any restriction to object paths except they cannot contain `:` (in that case use the dict notation).
- I think it makes more sense to apply the rule to the first `.root:` instead of the last given how fsspec chains urls, but I cannot think of any reasonable case where it would trigger more than once (a zip file having a .root extension instead of .zip...). In this case for instance matching the first appearance would make it work.

In any case I will make a separate PR for this into `main-fsspec` (which will be live next minor version).

Otherwise do you @nsmith- @jpivarski have any other comments regarding this PR? I'm not planning on further changes.

jpivarski: I moved this to #1025.

src/uproot/writing/writable.py
jpivarski: Yes, this argument was historical—the ability to use objects instead of paths was added later. A name like `file_path_or_object` is rather long; I'd be tempted to use a word like `source` except that it's the name of a class (so I'd then expect the `source` to be a `Source`). Perhaps `src` would be a good choice.

_Except_ that this argument is passed around a lot of functions. If it has one name (`file_path`) in some functions and another (`src`) in others, that would increase the confusion, rather than decrease it. Finding out what is the correct set of function arguments to update would be a project in itself.

Adding a type hint makes short work of that, just as the documentation did (though the type hint is more visible than the documentation).

tests/test_0692_fsspec_writing.py
```diff
@@ -0,0 +1,125 @@
...
+    # when https://github.com/fsspec/filesystem_spec/pull/1426 is merged this will work, we must remove the catch
+    with pytest.raises(ValueError):
+        with uproot.recreate(uri) as f:
+            f["tree"] = {"x": np.array([1, 2, 3])}
+
+        with uproot.open(uri) as f:
+            assert f["tree"]["x"].array().tolist() == [1, 2, 3]
+
+        if scheme == "simplecache::memory://":
+            raise ValueError("wait for next fsspec release and remove this")
```
jpivarski, line 125: Doesn't this raise an error?

jpivarski (approved):
It all looks good! I see that it's tested for local files, in-memory files, and ssh, which are three cases that make sense (depending on how remote the ssh server is).

I'll need to know more about what a "chain of filesystems" would do. Is it something we'd even be able to use, given our API?

I think it's ready to merge. I don't know why the one case below isn't raising an error, and it's an odd way to make a note-to-self about what to do next.

tests/test_0692_fsspec_writing.py
lobis: The error it's going to be catched on the outer scope. The idea is that the `memory://` will currently raise a `ValueError` because of a bug in fsspec that has already been corrected but not pushed to pypi yet. Once this is pushed to pypi this will no longer raise a `ValueError` and the test will fail, prompting to just remove the catches. I later added the `simplecache::memory//` scheme which for some reason doesn't fail so I added this manual raise which should also be removed.

lobis:
> I'll need to know more about what a "chain of filesystems" would do. Is it something we'd even be able to use, given our API?

I guess it has some niche uses, for example, opening a root file inside a zip file available via http. It should also work for writing. But in general I think only the local caching for writing will be used "frequently".

matthewfeickert:
:wave: I just wanted to raise that this PR breaks `pyhf` nightly HEAD of dependency tests. For example, installing `uproot` from 28b0f7bdbeb18b54ae2af4740607d982eea51b35 causes `pyhf`'s `test_export_root_histogram` test in `tests/test_export.py` to fail. Here's the [relevant bit](https://github.com/scikit-hep/pyhf/blob/43a84a477a93df640faea5c0526dc3abeead9759/tests/test_export.py#L411-L421)

```python
def test_export_root_histogram(mocker, tmp_path):
    """
    Test that pyhf.writexml._export_root_histogram writes out a histogram
    in the manner that uproot is expecting and verifies this by reading
    the serialized file
    """
    mocker.patch("pyhf.writexml._ROOT_DATA_FILE", {})
    pyhf.writexml._export_root_histogram("hist", [0, 1, 2, 3, 4, 5, 6, 7, 8])

    with uproot.recreate(tmp_path.joinpath("test_export_root_histogram.root")) as file:
        file["hist"] = pyhf.writexml._ROOT_DATA_FILE["hist"]
```

which then fails with a `TypeError` and the following traceback

<details>
<summary>Collapsed for space:</summary>

```pytb
========================================================================================== FAILURES ==========================================================================================
_________________________________________________________________________________ test_export_root_histogram _________________________________________________________________________________

mocker = <pytest_mock.plugin.MockerFixture object at 0x7ff50e8d4cd0>, tmp_path = PosixPath('/tmp/pytest-of-feickert/pytest-11/test_export_root_histogram0')

    def test_export_root_histogram(mocker, tmp_path):
        """
        Test that pyhf.writexml._export_root_histogram writes out a histogram
        in the manner that uproot is expecting and verifies this by reading
        the serialized file
        """
        mocker.patch("pyhf.writexml._ROOT_DATA_FILE", {})
        pyhf.writexml._export_root_histogram("hist", [0, 1, 2, 3, 4, 5, 6, 7, 8])
    
>       with uproot.recreate(tmp_path.joinpath("test_export_root_histogram.root")) as file:

mocker     = <pytest_mock.plugin.MockerFixture object at 0x7ff50e8d4cd0>
tmp_path   = PosixPath('/tmp/pytest-of-feickert/pytest-11/test_export_root_histogram0')

tests/test_export.py:420: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.pyenv/versions/pyhf-dev-cpu/lib/python3.11/site-packages/uproot/writing/writable.py:148: in recreate
    sink = _sink_from_path(file_path, **storage_options)
        file_path  = PosixPath('/tmp/pytest-of-feickert/pytest-11/test_export_root_histogram0/test_export_root_histogram.root')
        options    = {}
        storage_options = {}
../../../.pyenv/versions/pyhf-dev-cpu/lib/python3.11/site-packages/uproot/writing/writable.py:80: in _sink_from_path
    return uproot.sink.file.FileSink.from_object(file_path_or_object)
        file_path_or_object = PosixPath('/tmp/pytest-of-feickert/pytest-11/test_export_root_histogram0/test_export_root_histogram.root')
        storage_options = {}
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'uproot.sink.file.FileSink'>, obj = PosixPath('/tmp/pytest-of-feickert/pytest-11/test_export_root_histogram0/test_export_root_histogram.root')

    @classmethod
    def from_object(cls, obj) -> FileSink:
        """
        Args:
            obj (file-like object): An object with ``read``, ``write``, ``seek``,
                ``tell``, and ``flush`` methods.
    
        Creates a :doc:`uproot.sink.file.FileSink` from a file-like object, such
        as ``io.BytesIO``. The object must be readable, writable, and seekable
        with ``"r+b"`` mode semantics.
        """
        if (
            callable(getattr(obj, "read", None))
            and callable(getattr(obj, "write", None))
            and callable(getattr(obj, "seek", None))
            and callable(getattr(obj, "tell", None))
            and callable(getattr(obj, "flush", None))
            and (not hasattr(obj, "readable") or obj.readable())
            and (not hasattr(obj, "writable") or obj.writable())
            and (not hasattr(obj, "seekable") or obj.seekable())
        ):
            self = cls(None)
            self._file = obj
        else:
>           raise TypeError(
                """writable file can only be created from a file path or an object
    
    * that has 'read', 'write', 'seek', and 'tell' methods
    * is 'readable() and writable() and seekable()'"""
            )
E           TypeError: writable file can only be created from a file path or an object
E           
E               * that has 'read', 'write', 'seek', and 'tell' methods
E               * is 'readable() and writable() and seekable()'

cls        = <class 'uproot.sink.file.FileSink'>
obj        = PosixPath('/tmp/pytest-of-feickert/pytest-11/test_export_root_histogram0/test_export_root_histogram.root')

../../../.pyenv/versions/pyhf-dev-cpu/lib/python3.11/site-packages/uproot/sink/file.py:55: TypeError
================================================================================== short test summary info ===================================================================================
FAILED tests/test_export.py::test_export_root_histogram - TypeError: writable file can only be created from a file path or an object
```

</details>

I can raise an Issue if that's helpful, but I thought I'd start here in case this is relevant to ongoing discussion.

alexander-held:
Hi, I ran into the same thing in `cabinetry`, here is a minimal reproducer:
```python
import pathlib
import uproot

uproot.recreate(pathlib.Path("f.root"))
```
I will open an issue for reference.

jpivarski:
Looking at the stack trace, it must be because pyhf was relying on passing a file-like object `obj` that satisfies

```python
    callable(getattr(obj, "read", None))
and callable(getattr(obj, "write", None))
and callable(getattr(obj, "seek", None))
and callable(getattr(obj, "tell", None))
and callable(getattr(obj, "flush", None))
and (not hasattr(obj, "readable") or obj.readable())
and (not hasattr(obj, "writable") or obj.writable())
and (not hasattr(obj, "seekable") or obj.seekable())
```

and this has somehow changed. (It wasn't supposed to.) @lobis?

----------------

I just tried a test and I don't see what's wrong with it.

```python
import os
import numpy as np
import uproot

class FileLikeObject:
    def __init__(self):
        self.data = np.zeros(0, dtype=np.uint8)
        self.pos = 0
    def tobytes(self) -> bytes:
        return self.data.tobytes()
    def ensure(self, size: int) -> None:
        if size > len(self.data):
            data = np.zeros(size, dtype=np.uint8)
            data[:len(self.data)] = self.data
            self.data = data
    def read(self, num_bytes: int) -> bytes:
        self.pos += num_bytes
        self.ensure(self.pos)
        return self.data[self.pos - num_bytes : self.pos].tobytes()
    def write(self, data: bytes) -> None:
        self.pos += len(data)
        self.ensure(self.pos)
        self.data[self.pos - len(data) : self.pos] = np.frombuffer(data, np.uint8)
    def seek(self, offset: int, whence: int = os.SEEK_SET) -> None:
        assert whence in (os.SEEK_SET, os.SEEK_CUR, os.SEEK_END)
        if whence == os.SEEK_SET:
            self.pos = offset
        elif whence == os.SEEK_CUR:
            self.pos += offset
        elif whence == os.SEEK_END:
            self.pos = len(self.data) + offset
        self.ensure(self.pos)
    def tell(self) -> int:
        return self.pos
    def flush(self) -> None:
        pass
    # def readable(self) -> bool:
    #     return True
    # def writable(self) -> bool:
    #     return True
    # def seekable(self) -> bool:
    #     return True

file = FileLikeObject()
writable_file = uproot.recreate(file)
writable_file["hist"] = np.histogram(np.random.normal(0, 1, 1000), bins=100, range=(-5, 5))

file.tobytes()   # lots of bytes

readable_file = uproot.open(file)
readable_file["hist"].to_hist()
```

image

It doesn't matter whether the `readable`, `writable`, `seekable` methods exist or not.

alexander-held:
> Looking at the stack trace, it must be because pyhf was relying on passing a file-like object obj that satisfies

I think this happened via the `tmp_path` fixture in `pytest` creating a `pathlib.Path` (-like?) object.

jpivarski:
I need to read all of my email before responding to anything...

lobis:
There is a check to check if the argument is a path or an object. It was doing the check poorly (checking if it's a string) and failed when a path object was being passed. Now it tests for object (with a new helper function), otherwise assumes path-like object (and transforms to string).

https://github.com/scikit-hep/uproot5/pull/1031 is ready to be merged with the fix.

PR #1022, opened 2023-11-14
feat: fsspec as required dependency

lobis:
`main-fsspec` will exist alongside `main` and release candidate distributions will be produced from this branch until it's deemed ready to merge into `main`. It will be merged without squashing. (changes from `main` should be rabased instead of merged in order to avoid merge commits in `main` since we're not squashing).


jpivarski (approved):
I guess this is it—it's finally time to merge `main-fsspec` into `main`.

Be sure to not squash-and-merge this one, since we want to preserve the PR-level granularity in our commit history.

image

"Create a merge commit" is what you want, if it will let you do that with our "Require linear history" on.

image

If it complains about linear history, I can temporarily turn that off, merge, and then turn it back on.

If it requires you to rebase, be careful.

Issue #173, opened 2020-11-02
multiprocessing and uproot4 


ramankhurana:
Hello 

I am trying to run uproot4 parallel on multiple files but it seems to be stuck at reading the TTree from root files. It used to work for uproot 3 but not in uproot4. 

I found there are some possibilities to do parallel processing using uproot4.reading.open but i could not understand from documentation what exactly I should be doing to use it for multiple files in parallel. 

Thanks you.

jpivarski:
There aren't supposed to be any constraints. Uproot (both versions) spawn threads, but it's possible for objects in a multiprocessing environment to spawn their own threads. You don't want to be passing objects derived from one open file from one process into another (something you should avoid in general because processes don't share memory or file handles), but it should be perfectly okay for processes in `multiprocessing` to each open their own files and work with them independently.

ramankhurana:
Thanks for quick response as always. Yes indeed, I am not trying to pass any object from one thread to another, just to separate independent threads. 

I quote an example here but rootfile are in public area to do the test, if not possible I can share it somewhre on thw weblink, 
```
import uproot4
import awkward1 as ak
import multiprocessing as mp
from functools import partial

file_=["/afs/cern.ch/user/k/khurana/public/test.root", "/afs/cern.ch/user/k/khurana/public/test.root"]

def readfile(inputfile):
    file_=uproot4.open(inputfile)
    print ("file opened")
    tree_ = file_["outTree"].arrays()
    print ("number of events: ", len(tree_))
```

Now if I call this using 
```
readfile(file_[0])
``` 

it works fine, 

however when tried to use the multiprocessing: 

iterable = file_                                                                                                                                                                                            
pool = mp.Pool(2)                                                                                                                                                                                           
func = partial(readfile)                                                                                                                                                                                    
pool.map(func, iterable)                                                                                                                                                                                    
pool.close()                                                                                                                                                                                                
pool.join()                                                                                                                                                                                                 


it is stuck forever, 

Do you have any clue what can be a reason for this, or you see something missing in the snippet. I wonder this was working fine for another version of code in uproot3. 

Thanks.

jpivarski:
"Stuck forever" is a symptom of an error in Uproot 4 that has no equivalent in Uproot 3, so that's a hint for me.

[...]

PR #1023, opened 2023-11-14
feat: set fsspec as default source

lobis:
- Removes support for the `*_handler` arguments (`s3_handler`, `file_handler`, etc.) which were already on a deprecation notice.

- Sets the fsspec source as default for all cases except:
    - file-like objects 
    - s3 due to https://github.com/scikit-hep/uproot5/pull/1012 (if set as default pytest will fail)


tests/test_0017_multi_basket_multi_branch_fetch.py
```diff
@@ -310,6 +310,7 @@ def test_cache():
         skhep_testdata.data_path("uproot-sample-6.20.04-uncompressed.root"),
         object_cache=100,
         array_cache="100 MB",
+        handler=uproot.source.file.MemmapSource,
```
lobis, line 313: Not so sure I am not breaking anything here, should the `fsspec` source behave exactly the same as the previous default source?

tests/test_0173_empty_and_multiprocessing_bugs.py
```diff
@@ -17,8 +16,8 @@ def test_empty():
         assert t["z"].array(library="np").tolist() == []
 
 
-def readone(filename):
-    with uproot.open(filename) as f:
+def read_one(filename):
+    with uproot.open(filename, handler=uproot.source.file.MemmapSource) as f:
```
lobis, line 20: https://github.com/scikit-hep/uproot5/issues/173

This won't work with fsspec at the moment (meaning it won't work without manually specifing a different source), I'm not sure what the consecuences might be

tests/test_0017_multi_basket_multi_branch_fetch.py
jpivarski: If the default file-reader becomes non-memory mapping, that's okay. `MemmapSource` is still opt-in, and memory-mapping wasn't as important of a performance feature as it was originally thought to be. (The data have to be decompressed, byte-swapped, and concatenated, so the data in the file can't be used directly, anyway.)

tests/test_0173_empty_and_multiprocessing_bugs.py
jpivarski: From the looks of it, this test was going after a bug that's particular to memory-mapping. Explicitly setting the handler to recreate that situation is fine.

jpivarski (approved):
Good—this is going into `main-fsspec`, to be part of the 5.2.0rc1 pre-release.

It's ready to merge into that (non-default) branch.

nsmith- (commented):
I suppose we should ask @eduardo-rodrigues based on his input to #692 
For me it seems to deserve a minor version bump, otherwise OK.

nsmith-:
Well, I guess this is into `main-fsspec`, so what is the path to merge `main-fsspec` into `main`?

lobis:
> Well, I guess this is into `main-fsspec`, so what is the path to merge `main-fsspec` into `main`?

- https://github.com/scikit-hep/uproot5/pull/1022

It'll be merged when 5.2.0 is released (decemeber?) but in the meantime we will publish rc from this branch

eduardo-rodrigues:
> I suppose we should ask @eduardo-rodrigues based on his input to #692 For me it seems to deserve a minor version bump, otherwise OK.

Hello. Sorry to come late to the party. Sounds good, @nsmith- what you propose anyway.

PR #1024, opened 2023-11-15
fix: s3 source options and repr

lobis:
When reviewing tests I noticed the `uproot.source.S3.S3Source` has some issues not catched in the current test.

- Test was checking for a generic exception instead of a `FileNotFoundError`
- class `repr` failed due to not having correctly set `self._file_path`
- default options were not correctly loaded

I also added a skip to the test for failure (file not found) which apparently is causing the CI to fail (https://github.com/scikit-hep/uproot5/pull/1012). Not sure if the error comes from the `s3.S3Source` or `s3fs` source but it's definitely s3 related.


lobis:
I think these tests are what were causing the weird error in https://github.com/scikit-hep/uproot5/pull/1012!

src/uproot/source/http.py
```diff
@@ -607,7 +609,7 @@ def __repr__(self):
         if len(self._file_path) > 10:
             path = repr("..." + self._file_path[-10:])
         fallback = ""
-        if self._fallback is not None:
+        if hasattr(self, "_fallback") and self._fallback is not None:
```
jpivarski, line 612: Just a little simpler:

```suggestion
        if getattr(self, "_fallback", None) is not None:
```

jpivarski: It must have been getting here because it was raising an exception before `self._fallback` was even defined/added to the instance (somewhere in `__init__`, before line 576). Another solution would be to move the

```python
        self._fallback = None
```

earlier, but this is fine.

jpivarski (approved):
Thanks for fixing these. I see that there's still ongoing work on #1012 to fix `test_s3_fail`.

This is mergeable.

PR #1026, opened 2023-11-15
test: Emscripten build and test on GitHub Actions

lobis:
Based on https://pyodide.org/en/stable/development/building-and-testing-packages.html and https://github.com/numpy/numpy/blob/main/.github/workflows/emscripten.yml.

(Currently doesn't work since awkward-cpp is not available in the right version but it'll be fixed soon).

- https://github.com/pyodide/pyodide/pull/4298

This action will build the uproot wheel and run the tests using nodejs.

I will label and skip those tests that don't make sense to run on the browser (local files, etc.).


agoose77:
@lobis you can use the WASM artefact for awkward-cpp from our release workflow to test the CI: https://github.com/scikit-hep/awkward/actions/runs/6894740990

This won't last for ever (I suspect 30 days is the limit on our artefact caching atm).

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      # clone awkward and build a wheel, copy the wheel
+      - name: Awkward wheel
+        run: |
+          git clone https://github.com/scikit-hep/awkward.git
+          cd awkward
+          git checkout main
+          git submodule update --init --recursive
+          pyodide build
+          mv dist/*.whl ../dist
```
agoose77, line 53:
```suggestion
          git clone https://github.com/scikit-hep/awkward.git
          cd awkward
          git checkout main
          git submodule update --init --recursive
          pipx run nox -s prepare -- --headers --signatures --tests
          pyodide build
          mv dist/*.whl ../dist
```

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          node-version: 20
+
+      # clone awkward and build a wheel, copy the wheel
+      - name: Awkward wheel
+        run: |
+          git clone https://github.com/scikit-hep/awkward.git
+          cd awkward
+          git checkout main
+          git submodule update --init --recursive
```
agoose77, line 51:
```suggestion
          git submodule update --init --recursive
          pipx run nox -s prepare -- --headers --signatures --tests
```

agoose77: n.b. you'll still need awkward's C++ components too

lobis: My initial goal was to test the build (only uproot) and run the tests. However the uproot build will only work if the dependencies (awkward) are available, and there is a significant delay between the publication in pypi and it's availability from pyodide, this means that this CI will fail for a while whenever there is a release that modifies dependencies.

To test the CI job I though of building the awkward (and awkward-cpp) wheel in this CI job (in particular the version of awkward listed in pyproject.toml).

I am having some troubles understanding the necessary steps though.

jpivarski:
The new **docs/readthedocs.org:uproot** test requirement can be satisfied by merging #1084/[jpivarski/fix-readthedocs-documentation](https://github.com/scikit-hep/uproot5/tree/jpivarski/fix-readthedocs-documentation) into this PR (or by merging in `main`, which has it).

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,69 @@
...
+          pipx run nox -s prepare -- --headers --signatures --tests
+          pyodide build
```
agoose77, line 53: `pyodide build` only builds the pure-Python component, not the C++ build.

ariostas:
I think that #1272 is starting to be in pretty good shape, so I'll close this PR.

PR #1027, opened 2023-11-15
test: review skipped tests (networking timeouts)

lobis:
In this PR I review the skipped tests due to "flakiness". They are all xrootd tests.

After removing the skip some of them failed for various reasons, I made changes so tests pass. It's not the cleanest fix but this xrootd source is going to be replaced by fsspec (atleast for the default behaviour).

Ocasionaly some test will fail due to server issues, we should retry those. I think raising a specific error when this happens and configuring pytest to retry on this error is the way to go (this is what we done at the moment, but it doesn't work since we don't see `requests` errors). I updated the `http` source to throw a `http.client.HTTPException` and xrootd to throw a `TimeoutError`. This currently works for the ocasional xrootd and s3 fails. 

Probably these errors can be improved (not sure if it's correct to throw a `TimeoutError`). Before this changes we were throwing an `OSError` which imho is too generic to do retries on.


.github/workflows/build-test.yml
```diff
@@ -93,4 +93,4 @@ jobs:
 
       - name: Run pytest
         run: |
-          python -m pytest -vv tests --reruns 3 --reruns-delay 30 --only-rerun requests.exceptions.HTTPError
```
lobis, line 96: `requests` isn't used.

tests/test_0001_source_class.py
```diff
@@ -322,19 +319,14 @@ def test_xrootd(use_threads):
...
 
 
-@pytest.mark.skip(
-    reason="RECHECK: Run2012B_DoubleMuParked.root is super-flaky right now"
-)
 @pytest.mark.network
 @pytest.mark.xrootd
-@pytest.mark.parametrize("use_threads", [True, False], indirect=True)
-def test_xrootd_deadlock(use_threads):
+def test_xrootd_deadlock():
```
lobis, line 324: Not really sure of what I did here, don't really understand this test. Please review @nsmith- in case I broke something important.

jpivarski (approved):
That's awesome! I've been meaning to get back to tests that were excluded for "flakiness" (also in Awkward Array; some tests were turned off and labeled with "FIXMEs").

I've checked all of the fixes, and it all looks good to me. I'm going to leave this to @nsmith- because you have a question for him inline.

tests/test_0001_source_class.py
nsmith-: I think this test is obsolete now that XRootD upstream has truly fixed the issues discussed in https://github.com/scikit-hep/uproot5/pull/59 and our CI is not testing older xrootd that needs the patch of https://github.com/scikit-hep/uproot5/blob/6b6fa9458b4ae46894e053dee25c56f678567f9d/src/uproot/extras.py#L105-L109
In principle we could in https://github.com/scikit-hep/uproot5/blob/6b6fa9458b4ae46894e053dee25c56f678567f9d/.github/workflows/build-test.yml#L62C11-L62C11 specify older xrootd to test but it seems a bit moot.

nsmith- (approved):

tests/test_0001_source_class.py
nsmith-: Perhaps @chrisburr has input

PR #1032, opened 2023-11-16
feat: set `fsspec` (`s3fs`) as default handler for s3 paths

lobis:
This wasn't added in https://github.com/scikit-hep/uproot5/pull/1023 due to concerns regarding the pytest exception due to `s3fs` (https://github.com/scikit-hep/uproot5/pull/1012).

The issue is not understood yet but from my experience with this issue (it's readlly hard to reproduce) I don't think it's going to result in any problem. I think the issue comes from `s3fs` (but not 100% sure) as we are using this filesystem as any other but the issue only appears here. It'll probably be fixed on some newer release. Besides, this only appears on older version of python (in 3.11 or newer this doesn't appear) so I think it's safe to set the new s3 default to fsspec.

- Increased version to `5.2.0rc1`
- Fix xrootd pickle tests (were not being run before due to the parallel branch not being in sync with main, it should've tripped in another PR).


lobis:
https://github.com/scikit-hep/uproot5/actions/runs/6883516363/job/18724245334#step:8:782

Is this okay @nsmith- ? The test passes with the old xrootd handler but fails with the new one. I'm not quite sure what is going on. Is it okay to skip it, or is this a fundamental issue that needs to be fixed (probably not here though)?

nsmith-:
The pickle-ability of the sources is important. It looks like this is a deficiency of `fsspec-xrootd`. It appears that, although filesystem objects work (https://github.com/CoffeaTeam/fsspec-xrootd/blob/main/tests/test_basicio.py#L99-L108) the `File` objects do not.

nsmith-:
Do I recall correctly that holding an actual file handle is somewhat optional now? Perhaps the fsspec source should not bother with that?

jpivarski (approved):
I checked it over and I see that this is going into `main-fsspec` for 5.2.0rc1. The version number has also been changed in this PR. :heavy_check_mark:

Looks good! Feel free to merge!

PR #1033, opened 2023-11-16
feat: correct fsspec source serialization

lobis:
As discovered in https://github.com/scikit-hep/uproot5/pull/1032 the serialization of the fsspec source wasn't being done correctly.


lobis:
I'll merge this without waiting for approval given it's been implicitly approved in https://github.com/scikit-hep/uproot5/pull/1032#pullrequestreview-1735708927

PR #1042, opened 2023-11-22
chore: use ruff to import annotations

lobis:


lobis:
@agoose77 do you know any rule to enforce a single line jump before imports, to avoid https://github.com/scikit-hep/uproot5/pull/1042/files#diff-7bf9f23c1473db1a176a044ec01de0b429c4f8c6c3998aa7496f0d834a8a1b9cR3-R4?

agoose77:
@lobis I actually prefer that style. If you don't have strong feelings, I'd just leave it.

I'm not aware of any rule / configuration option to enforce this, but it doesn't mean that one doesn't exist. I had another look, and couldn't see anything that stands out to me.

lobis:
> @lobis I actually prefer that style. If you don't have strong feelings, I'd just leave it.
> 
> I'm not aware of any rule / configuration option to enforce this, but it doesn't mean that one doesn't exist. I had another look, and couldn't see anything that stands out to me.

Sure, we can leave it as it is. I couldn't find anything either so it's probably something we shouldn't care about anyway.

PR #1043, opened 2023-11-23
fix: `const std::string` identification

HaarigerHarald:
Hi,

I noticed that serializing an object with root containing a `const std::string` member like:

```cpp
struct Foo {
const std::string bar;
};
```

produces the following error:

```
ValueError: invalid C++ type name syntax at char 6

    const string
----------^
in file foobar.root
```

The "fix" adds another explicit check for `const std::string`, not sure though whether this is the best solution.


jpivarski (approved):
That fixes it, and it's at the right level of formality for this parser. (We have an explicit tokenizer, but not an explicit shift-reduce parser, just these `if` statements.)

This fix doesn't extend to the other types, and presumably one could have `const std::vector<int>` as well, though it might be less common than strings. To add the possibility of `const` to all of these types without doubling the number of `if` statements, perhaps we need an explicit shift-reduce parser...

For now, this is good. Do you have a small (< 1 MB) example file that needs this fix? I'd like to add a test, and for that, we need a file.

HaarigerHarald:
Hi,

here a small file created from within root via:

```cpp
struct Foo { const std::string bar; };
Foo f;
TFile* outfile = TFile::Open("foobar.root", "recreate");
TTree* tree = new TTree("FooBar", "FooBar");
tree->Branch("Foo", &f);
tree->Fill();
outfile->Write();
outfile->Close();
```

[foobar.zip](https://github.com/scikit-hep/uproot5/files/13473045/foobar.zip)

jpivarski:
@all-contributors please add @HaarigerHarald for code

jpivarski (approved):
All I did was add a test, and I verified that the test fails with the bug and passes with the bug fixed, so this is all set.

I'll merge the PR when the tests pass.

Issue #1054, opened 2023-12-05
Local files with colons not correclty handled by newest version of fsspec


lobis:
The current implementation of the `fsspec` source relies on being able to handle files with `:` in it's name (this is a way to specify which objects to read).

Up to `fsspec` version `v2023.10.0` this worked. `fsspec` is able to read a file such as `local-file-with:colons.root` or `local-file-with-colons.root:object` and treat them as local files (`local` protocol). Now this results in an error.

Depending on wether this is a intended or unintended change in fsspec behaviour we'll modify how the fsspec uproot source reads local files. If this was unintentional a test could be added to `fsspec` to prevent this from happening in the future.

PR #1055, opened 2023-12-06
test: add test for issue 1054 (newer fsspec failing to parse files with colons in name)

lobis:
I add an explicit test for this issue. This is currently also catched by other tests in the `main-fsspec` branch but I wouldn't be caught if we didn't make the switch to use fsspec for default local files (which is not clear yet).

Closes https://github.com/scikit-hep/uproot5/issues/1054

I will use this PR to fix this issue in case it needs some code change. I am hoping that https://github.com/fsspec/filesystem_spec/issues/1447 will address this (I assume it's a bug on fsspec part).

Currently this PR is failing until either fsspec publishes a new release with a fix, we set an upper limit for supported fsspec versions, or we modify how we handle paths with colons in our code. This should be handled before the next release.

I also added an explicit test for windows since colons cannot be part of a filename in windows.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -65,9 +65,6 @@ jobs:
       - name: Pip install the package
         run: python -m pip install .[test,dev]
 
-      - name: DELETE ME DEBUG ONLY
-        run: python -m pip install git+https://github.com/lobis/filesystem_spec@master
-
       - name: Run pytest
         run: |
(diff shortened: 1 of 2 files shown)


lobis:
Fixed in https://github.com/fsspec/filesystem_spec/pull/1452

lgray:
Should be fixed now in https://pypi.org/project/fsspec/2023.12.2/ @lobis

PR #1056, opened 2023-12-07
test: zstandard should be a test dependency, and xxhash goes with lz4.

jpivarski:


ioanaif (approved):
LGTM! (discussed this over Zoom)

PR #1062, opened 2023-12-12
build: change build to autogen version info

lgray:
This may help in streamlining releases in the future, but it's up to opinion on operations.
It also makes it easier for users to see if they are using a build installed from source.

@jpivarski @lobis


lgray:
@jpivarski some of these failures seem ~persistent but are obviously not related. 

Anyway - this one isn't super high priority. Decided to propose the option to y'all!

jpivarski (approved):
I had been avoiding this on principle, but I can no longer remember the reason why. Since I can't remember it, I guess it's time to give in and let hatch-vcs manage the version.py.

We should merge this _after_ the 5.2.0 release.

Thanks!

lgray:
@jpivarski reminder of this one, but again low priority.

jpivarski:
Yeah, I know. Okay, I'll try it out now.

jpivarski:
It worked, and I don't see any reason to not use hatch-vcs. Awkward Array uses hatchling as well, though its release process is more complex because there's an awkward-cpp part that absolutely must be versioned by hand. (Its version number is essentially an ABI between the two parts of the project, so we want to control that.)

PR #1064, opened 2023-12-13
test: do not skip aiohttp tests for Python 3.12 - update pytest rerun settings

lobis:
`aiohttp` appears to be available for python 3.12 so we can remove the skip.

I also increased the number of reruns in the pytest command to avoid failure due to flaky xrootd.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -67,7 +67,7 @@ jobs:
 
       - name: Run pytest
         run: |
-          python -m pytest -vv tests --reruns 3 --reruns-delay 30 --only-rerun "(?i)http|timeout|connection|socket"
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Good; these things are related to each other and make sense in the same PR.

As a refinement, we should label tests that might be flaky (or maybe the `network` label is already good enough for that) and do two pytest runs, one with only flaky tests and a high `--reruns` and another with no flaky tests and no `--reruns`.

When we introduce a bug and the test framework legitimately finds it, we don't want it running the same failure so many times. It would also inflate our test logs, since this rerun feature is not at the level of GitHub Actions; it's at the level of pytest.

jpivarski:
Oh, I missed the `--only-rerun` part of that. You're already doing the refinement. It's good, then!

PR #1075, opened 2023-12-20
test: fsspec cache

lobis:
test `simplecache`. Made it work with xrootd in https://github.com/CoffeaTeam/fsspec-xrootd/pull/42.


jpivarski (approved):
Thanks! I see how this tests URL chaining, and it correctly does a `pytest.importorskip` for all of the dependencies that it uses. It's also a straightforward addition, so I'll merge it. Merry Christmas!

PR #1076, opened 2023-12-20
test: xrootd server fixture

lobis:
Add a pytest fixture for xrootd server which can be used to serve files locally over xrootd. We might want to switch to use this for all xrootd tests to avoid networking / server issues.

I updated one xrootd test to use this new fixture.

I also renamed the old "server" fixture to "http_server" since now we have more than one server fixture.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -67,7 +67,7 @@ def serve():
 
 
 @pytest.fixture(scope="module")
-def server():
+def http_server():
     with serve() as server_url:
         yield server_url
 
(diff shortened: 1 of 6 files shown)


jpivarski (approved):
Cool! I looked through the implementation of the local XRootD server. I'm not sure why it's an iterator with exactly one yield: I don't think that allows it to run asynchronously.

lobis:
> Cool! I looked through the implementation of the local XRootD server. I'm not sure why it's an iterator with exactly one yield: I don't think that allows it to run asynchronously.

I've writen a couple of fixtures and they all follow the same pattern (this one is actually based on what's used in fsspec-xrootd). My understanding is that fixtures have this structure so that setup / cleanup can be defined in a single function. First the fixture method is run until the yield (setup), then the test code is ran, then the code after the yield (cleanup) is ran.

PR #1077, opened 2024-01-02
feat: supply a pre-calculated base form to avoid file opening

lgray:
Add `known_base_form` option (`awkward.forms.Form | None`) to `uproot.dask` so that opening root files at the client can be avoided for mature analyses where the base forms of all files are well known.

Right now it is mutual exclusive with `open_files=True`, since it brings no advantage in that case.

An initial implementation is included.

I tested the speedup with the following code:
```python3
import uproot
import json
import gzip
import awkward

import pyinstrument

if __name__ == "__main__":

    prof = pyinstrument.Profiler()
    prof.start()

    with gzip.open("nano_form.json.gz", "r") as in_form:
	known_form = awkward.forms.from_json(in_form.read())

    for _ in range(20):
	events = uproot.dask({"tests/samples/nano_dy.root": {"object_path": "Events", "steps": [[0,40]]}}, open_files=False, known_base_form=known_form)

    prof.stop()
    print(prof.output_text(unicode=True, color=True, show_all=True))

    prof = pyinstrument.Profiler()
    prof.start()

    for _ in range(20):
        events = uproot.dask({"tests/samples/nano_dy.root": {"object_path": "Events", "steps": [[0,40]]}}, open_files=False)

    prof.stop()
    print(prof.output_text(unicode=True, color=True, show_all=True))
```

Which on a arm64 macbook pro with SSD gives about a 6x speedup. The gain will be enormous when using xrootd.

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -42,6 +42,7 @@ def dask(
     open_files=True,
     form_mapping=None,
     allow_read_errors_with_report=False,
+    known_base_form=None,
     **options,
 ):
     """
@@ -99,6 +100,8 @@ def dask(
             empty array for these nodes in the task graph. The return of this function then
             becomes a two element tuple, where the first return is the dask-awkward collection
             of interest and the second return is a report dask-awkward collection.
+        known_base_form (awkward.forms.Form | None): If not none use this form instead of opening
+            one file to determine the dataset's form. Only available with open_files=False.
         options: See below.
 
     Returns dask equivalents of the backends supported by uproot. If ``library='np'``,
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
I agree. At the last minute, I dithered on this, thinking that we could solve the underlying problem by optionally incorporating Numba into the TBranch and TLeaf-deserialization process, since these are the two classes that are especially numerous for TTree metadata. In the end, though, I saw that it would have to go through [read_object_any](https://github.com/scikit-hep/uproot5/blob/d5bc01dba96b8ef3c5eff2ddc544318daed35508/src/uproot/deserialization.py#L189), which builds a generic object graph by reference. It would be a bigger project than the timescale on which you need this.

I'll merge this now and release 5.2.1.

PR #1084, opened 2024-01-16
docs: fix ReadTheDocs documentation

jpivarski:


jpivarski:
I got an email and there's the docs/readthedocs.org:uproot test failure above.

I made it a required test, so I probably don't need the emails. PRs will need to keep the docs working.

jpivarski:
Despite all of these build failures, I only got an email notification from the first one in this PR. That's also not particularly useful... I'm turning off the email notifications now.

PR #1085, opened 2024-01-18
feat: add dask_to_root

zbilodea:


lgray:
Just a stylistic comment but I think the function would be better situated within uproot as `uproot.dask.to_root`, thoughts?

@jpivarski

jpivarski:
You mean the function would be an attribute of another function, `uproot.dask`?

I've gone back and forth on that—there used to be an `ak.to_parquet.dataset` for converting already-written Parquet files into a dataset (the directory with metadata convention). But by not being in the same level of hierarchy as the rest of the `ak.*` functions, it was hard to discover.

I understand the logic of writing-from-Dask being associated with reading-from-Dask. Uproot doesn't already have the flat bundle of functions that Awkward Array has, so there's less to maintain. But still, locating functions as attributes on other functions is unusual enough in Python as to be an antipattern—if someone is looking for it, they're unlikely to think of looking there.

Perhaps it should just be a naming convention: `uproot.dask_write`? From a typing point of view, that only turns a dot into an underscore, but from an organizational point of view, there's no nesting of functions. (I don't think it needs "to ROOT" because this is Uproot—it's only ever going to read from or write to ROOT.)

tests/test_dask_to_root.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+    assert len(file_2["tree"]["Jet_Px"].arrays()) == int(len(arr) / partitions)
+
+
+def HZZ_test_3_partitions():
+    """
+    Write data from HZZ with 3 partitions
+    """
+    partitions = 3
+    arr = uproot.open(skhep_testdata.data_path("uproot-HZZ.root"))["events"].arrays()
```
lgray, line 117: At least one of your tests should use a distributed.Client like:
```python3
from distributed import Client

def test_something():
    with Client() as _:
        #do dask things
```

This will ensure everything you've written is serializable and shippable to worker nodes on a cluster!

pyproject.toml
```diff
@@ -39,7 +39,9 @@ dependencies = [
   "numpy",
   "fsspec",
   "packaging",
-  "typing_extensions>=4.1.0; python_version < \"3.11\""
+  "typing_extensions>=4.1.0; python_version < \"3.11\"",
+  "dask >=2023.04.0",
+  "dask-awkward>=2023.12.1"
```
lgray, line 44: dask shouldn't be a top level dependency of the uproot package? @jpivarski 

I suppose below we should import dask within `dask_write` such that it complains when dask is not available, but doesn't impede more standard functioning of uproot.

lgray:
@jpivarski ok that naming scheme makes sense, and I am fine with it!

jpivarski:
In Uproot and Awkward, name the tests by PR number:

```
tests/test_1085_dask_to_root.py
```

src/uproot/version.py
```diff
@@ -0,0 +1,19 @@
+# file generated by setuptools_scm
+# don't change, don't track in version control
```
jpivarski, line 2: Don't check this file into version control; 3816ead34084b0f21bbed8aae9344b4559c47f69 adds a check against that.

pyproject.toml
jpivarski: Correct: `dask` and `dask-awkward` should not be strict dependencies for Uproot. These two lines should be removed.

tests/test_dask_to_root.py
```diff
@@ -0,0 +1,159 @@
+import pytest
+import awkward as ak
+import uproot
+
+skhep_testdata = pytest.importorskip("skhep_testdata")
+dask = pytest.importorskip("dask")
+da = pytest.importorskip("dask.array")
+dask_awkward = pytest.importorskip("dask_awkward")
```
jpivarski, line 8: This allows the tests to run or be skipped, depending on whether Dask is installed, so you can remove them from the pyproject.toml.

jpivarski (changes requested):
This is good; just be sure to

* remove version.py from this PR, now that it's auto-generated
* remove `dask` and `dask-awkward` from the list of dependencies

After that, request a review from me and this will be mergeable.

tests/test_1085_dask_write.py
```diff
@@ -0,0 +1,87 @@
...
+    file_1 = uproot.open(os.path.join(tmp_path, "distribute-part0.root"))
+    assert len(file_1["tree"]["Jet_Px"].arrays()) == math.ceil(len(arr) / partitions)
+    assert ak.all(file_1["tree"]["Jet_Px"].arrays()["Jet_Px"][0] == arr["Jet_Px"])
+    file_2 = uproot.open(os.path.join(tmp_path, "distribute-part1.root"))
+    assert len(file_2["tree"]["Jet_Px"].arrays()) == int(len(arr) / partitions)
+    assert ak.all(file_2["tree"]["Jet_Px"].arrays()["Jet_Px"][0] == arr["Jet_Px"])
+
+
+# if __name__ == "__main__":
+#     test_compute("\my-output")
```
jpivarski, line 87: (clean up)

```suggestion
```

jpivarski (approved):
Also remove the 7 output files before merging:

```
src/uproot/my-output/*.root
```

But, after they're removed (as well as the `if __name__ == "__main__"` part), this is ready to merge!

tests/test_1085_dask_write.py
```diff
@@ -0,0 +1,87 @@
...
+    assert len(file_2["tree"]["a"].arrays()) == int(len(arr) / partitions)
+    assert ak.all(file_2["tree"]["a"].arrays()["a"][0] == arr[1]["a"])
+
+
+def test_compute(tmp_path):
+    print("here")
+    partitions = 2
+    arr = uproot.open(skhep_testdata.data_path("uproot-HZZ.root"))["events"].arrays()
+    dask_arr = dask_awkward.from_awkward(ak.from_iter(arr), partitions)
+    with Client() as _:
```
jpivarski, line 72: Just as a Python language thing, this should be equivalent to the above:

```suggestion
    with Client():
```

jpivarski (approved):
Nevermind; I was looking at the wrong diffs. The temporary output files have already been removed.

Issue #288, opened 2021-03-02
Interpretation issue in extraction of vector of user-defined class objects


dano0014:
Our ROOT structure is as follows. Within the ROOT file there are a few TTrees, but the only one that is relevant is the Event TTree. Within that Event TTree there is a group that shows up with the typename "std::vector<ChannelData> channels" as each event has some # of channels. Each channel class has a member, "pulse". The pulse member shows up with the typename "std::vector<Pulse> pulses". I am looking for the information in the Pulse group, but it is showing up as a pointer (I think). The interpretation says "AsObjects(AsArray(True,False,None" but that is all that I can see. In ROOT the code would look like "event->channels[0].pulses[0].nph" and I would like to know how to convert that to python. I have also sent the Jupyter Notebook so you can better understand what I mean. Thank you for any help you can provide.

Also on the PyHEP-newcomers chat on Gitter, Mr. Pivarski had suggested writing "pulse.array()" which crashes my computer and "puls.array(library-"np")" which returns a "memberwise serialization of AsArray" NotImplementedError.

https://files.gitter.im/5c939a19d73408ce4fbb5e66/R1CK/uproot_problem.ipynb

https://northwestern.box.com/s/4x1w6ym8wsion6lgx49xwfuhkh7gvcdk
that is the link to the full data file (warning, it is pretty big, 3.5gb).

jpivarski:
Actually, if it says `NotImplementedError("memberwise serialization of AsArray")`, that's a duplicate of issue #38. I need to investigate and add deserialization code for ROOT's "memberwise" option—it simply doesn't exist at the moment.

Although the `"AsObjects(AsArray(True,False,None"` string also looks wrong: an Interpretation ought to return its string representation but is just returning (the string) `"None"` for some reason. I'll look at that, too.

jpivarski:
Sorry that I lost track of this one!

I've verified that you get the "memberwise serialization" error, but that is a duplicate of #38.

The other thing I was worried about, the interpretation name with `"None"` in it, is apparently gone:

```python
>>> uproot.open("uproot-issue-288.root:Events/event/channels/channels.pulses").interpretation
AsObjects(AsArray(True, False, AsVector(False, Model_Pulse)))
```

PR #1086, opened 2024-01-18
fix: pandas performance on files with many branches

ioanaif:


jpivarski:
The example file is small; it has 1000 branches, but they're not filled with much data. Unfortunately, we can't keep large test files in our CI, so this is something that can only be tested manually.

I found a CMS NanoAOD file (which I can share with you privately). By selecting `filter_typename="bool"`, we can get approximately 1000 branches with a simple data type that should be easy to wrap in Pandas. In `main` (with the file in warm cache using [vmtouch](https://anaconda.org/conda-forge/vmtouch)),

```python
import uproot, time
tree = uproot.open("Run2018D-DoubleMuon-Nano25Oct2019"
    "_ver2-v1-974F28EE-0FCE-4940-92B5-870859F880B1.root:Events")
tick = time.perf_counter()
for _ in tree.iterate(filter_typename="bool", step_size=100000, library="np"):
    tock = time.perf_counter()
    print(tock - tick)
    tick = tock
```

prints times on average 3.4 ± 0.1 sec, the time needed to read the arrays at all (using NumPy). Swapping `library="np"` for `library="pd"` prints lots and lots of warnings, but the times are 4.19 ± 0.08, so the overhead of `_pandas_memory_efficient` is 0.8 sec.

In `ioanaif/fix-pandas-memory-issue-1070`, the `library="np"` time is 3.0 ± 0.2 sec and the `library="pd"` time is 4.1 ± 0.2. The Pandas `PerformanceWarning` is gone, but there's no noticeable impact on the actual performance.

I found another file, from issue #288, which has a lot of large, simple-typed branches that can be selected with

```python
filter_typename=["double", "/double\[[0-9]+\]/", "bool", "/bool\[[0-9]+\]/"]
```

It takes `main` 2.50 ± 0.06 sec to read this file with `library="np"` and 2.89 ± 0.02 sec to read with `library="pd"`, but in `ioanaif/fix-pandas-memory-issue-1070`, it takes 2.58 ± 0.07 sec to read with `library="np"` and 9.7 ± 0.1 sec to read with `library="pd"`. **The Pandas overhead went from 0.4 sec to 7.1 sec by introducing this PR. That's a regression.**

The `PerformanceWarning` suggests using `pd.concat`, not the `pd.DataFrame` constructor. I wonder if that is relevant. With the following diff in this PR,

```diff
--- a/src/uproot/interpretation/library.py
+++ b/src/uproot/interpretation/library.py
@@ -856,7 +856,9 @@ class Pandas(Library):
 
         elif isinstance(how, str) or how is None:
             arrays, names = _pandas_only_series(pandas, arrays, expression_context)
-            return pandas.DataFrame(data=arrays, columns=names)
+            out = pandas.concat(arrays, axis=1, ignore_index=True)
+            out.columns = names
+            return out
 
         else:
             raise TypeError(
```

it now takes 2.6 ± 0.1 sec for `library="np"` and 2.96 ± 0.08 sec for `library="pd"`, which is no regression (0.4 sec of Pandas overhead is almost exactly what the `_pandas_memory_efficient` function in `main` had).

Going back to the NanoAOD file, it's 3.1 ± 0.2 sec with `library="np"` and 3.6 ± 0.2 sec with `library="pd"`, still acceptable.

So the bottom line is that using the `pandas.DataFrame` constructor introduces a performance regression (while eliminating Pandas's warning!) in one out of the two cases tried, but using the `pandas.concat` function does not. After this comment, I'll add it as a commit here.

jpivarski:
After adjusting the code so that it passes all tests (54cab87319329dba0a7498870c0168d7c910e8c6), the time with the file from from issue #288 is still 2.6 ± 0.1 sec with `library="np"` and 3.0 ± 0.1 sec with `library="pd"`. The 7-second issue didn't come back.

... until I manually reverted the code to

```python
pandas.DataFrame(data=arrays, columns=names)
```

just to be sure that it's persistent, that I'm not just imagining things. It is definitely the case that Pandas is doing something bad when we run its constructor. (Surely that was the first thing that I tried, way back when, and encountered some bad behavior that made me write `_pandas_memory_efficient` in the first place.)

Oh, if I run `gc.disable()` and then `gc.collect()` before the constructor, the time isn't as bad (6.0 ± 0.1 sec, rather than 9.6 ± 0.1 sec). Whatever bad thing Pandas is doing in its constructor, it involves creating a lot of short-lived objects.

jpivarski (approved):
I think this is ready to merge. I'll let you merge it, in case you want to do any more counter-edits or tests.

PR #1088, opened 2024-01-19
fix: recorrds -> records

jpivarski:


jpivarski:
I'm merging this for the same reason as https://github.com/scikit-hep/uproot5/pull/1104#issuecomment-1912432150.

Issue #518, opened 2021-12-02
Include an option to use the "deflate" library instead of Python's built-in "zlib"


jpivarski:
This one: https://pypi.org/project/deflate/

@Moelf and @aminnj have shown that using this third-party library gives a 30% performance improvement. Uproot and UnROOT used to have about the same speed, but replacing standard "zlib" with "deflate" in UnROOT sped it up by 30%. Compare Uproot:

![image](https://user-images.githubusercontent.com/1852447/144336866-cb7caa42-bf08-4606-8dfc-733c1446f0aa.png)

with UnROOT:

![image3](https://user-images.githubusercontent.com/1852447/144338067-92365ec8-b80d-4505-ba4f-a930524b1ec6.png)

(Cross-reference to an UnROOT issue?)

Moelf:
wait, the first screenshot is on two different branches, the point was to that it's not limited to flat branch (one might think that's the only place where decompression dominates)

jpivarski:
> the first screenshot is on two different branches of the TTree, the point was to that it's not limited to flat branch (one might think that's the only place where decompression dominates)

I updated the original message with a better presentation of the demo.

aminnj:
UnROOT issue: https://github.com/tamasgal/UnROOT.jl/pull/137
With some more discussion linked within

The benchmark file comes from http://opendata.cern.ch/record/12341

jpivarski:
This could piggyback on #1090's interface to use alternate ZLIB libraries. That PR introduces a back-door to add ISAL.

How does deflate compare with ISAL (in decompression time and compressed file size)? Do we want two back doors?

PR #1090, opened 2024-01-20
feat: use cramjam for lz4 and zstd

### LINKED ISSUE #1089: Use cramjam to provide compression utilities
@lgray mentioned https://pypi.org/project/cramjam/ to me and it looks like a nice solution to provide many of the ROOT compression algorithms in a single dependency-free package. Additionally, it allows to declare the output length so it can pre-allocate the buffer, which may provide some speedup for algorithms other than lz4, which is the only one currently using the uncompressed size hint:
https://github.com/scikit-hep/uproot5/blob/fd0637bf978d75619e125a7f624827348f757865/src/uproot/compression.py#L194

This is an internal feature and would not provide any user enhancement other than a potential speed-up

lgray:
Fixes #1089 

zlib remains the standard python library since the `deflate` interface of cramjam doesn't calculate the zlib headers for the streams.

I suppose I could implement that, but I also found a buffer size hinting option that was not being used for the standard python zlib, that may help. So I used that instead.

I have put in a feature request issue for lzma to the cramjam github, we'll see what happens! Looks like there's good package coverage for lzma in rust, including with hinted buffer size, so probably not too much of an ask.

Generally - we now have fewer dependencies for uproot through this package. Wheel coverage for cramjam looks sufficient for our platform requirements.

@jpivarski @nsmith- @martindurant


lgray:
Since it's only replacing lz4 and zstd, I suppose I could put this in the uproot_extras area and use it that way?

jpivarski:
(This is happening faster than I had been thinking it would. Thanks for opening a PR!)

Dependencies can only change at the minor version numbers, and since we're starting from a state in which `lz4`/`xxhash` and `zstd` are both optional dependencies, phase 1 would have `cramjam` be an optional dependency, instead (in uproot_extras). The thing I have to think about is whether there will be a phase 2 in which `cramjam` becomes a strict dependency (in the pyproject.toml).

Early on, the fact that Uproot was pure Python was considered very valuable. (Most people who answered my question about why they're using it mentioned pure Python portability.) But since Uproot 5, Awkward Array has been a strict dependency, and that's a compiled extension. Cramjam is a much "better behaved" compiled extension than lz4. (On some platforms, the lz4 package _contains_ its decompression code; on other platforms, it _expects to find it_ elsewhere in the operating system.)

I guess it would be okay. After all, [cramjam is on all platforms](https://pypi.org/project/cramjam/#files), including Pyodide.

There's usually 2 months between changes like this—at least, we do a cadence of 2 months between Awkward minor versions, and use the minor versions to apply deprecations and change dependencies. Uproot hasn't been as regular, but we don't want to rock peoples' boats unnecessarily.

lgray:
Ah, hmm, looks like we need to deal with the checksum!

jpivarski:
Since Python 3, LZMA has been in the standard library, so there's no reason to change that. Except possible performance.

Which reminds me, if cramjam has much worse performance, that would be a reason to stop and reconsider, but I highly doubt it. (They probably just plugged in the same algorithms, maybe translated to Rust.)

lgray:
The main performance advantage appears to be hinting the buffer sizes, which the python lzma library doesn't support.

lgray:
While cramjam does include benchmarks none of this demonstrate this particular feature.

lgray:
I'll convert this to a draft while we discuss the finer points.

lgray:
While we're at it: cramjam specifically recommends using https://github.com/pycompression/python-isal instead of itself for zlib compression.

Despite being the intel library for accelerated compression it supports all the standard conda architectures.

lgray:
Wow - the response for adding lzma was super quick https://github.com/milesgranger/cramjam/pull/127

Tested in a local branch and it immediately passes tests. No performance improvement with a small file though, but the interface to compression is much more uniform and clean and we can always use the output size hints.

lgray:
Added `zlib` to extras such that it tries to use `isal` if it is available but otherwise returns the standard python zlib.

The cost of failing the exception is miniscule compared to decompression time in realistic use cases. `isal` is typically 3x-5x faster than python's implementation, so this is definitely a useful optimization.

lgray:
Apologies for the noise. Aside from the forthcoming lzma feature in cramjam I do not expect to make any further major changes to this PR.

lgray:
@jpivarski a review of the present state of this PR would be appreciated!

src/uproot/compression.py
```diff
@@ -7,8 +7,8 @@
 
 from __future__ import annotations
 
+import math
 import struct
-import threading
```
jpivarski, line 11: The Python ZSTD library is thread-unsafe, so it's good to be able to remove this now.

src/uproot/compression.py
```diff
@@ -88,11 +88,15 @@ class _DecompressZLIB:
     name = "ZLIB"
     _2byte = b"ZL"
     _method = b"\x08"
+    zlib = uproot.extras.zlib()
```
jpivarski, line 91: ZLIB wasn't in extras before because it's part of the Python standard library; we can assume it's always there.

src/uproot/compression.py
```diff
@@ -88,11 +88,15 @@ class _DecompressZLIB:
...
 
     def decompress(self, data: bytes, uncompressed_bytes=None) -> bytes:
-        import zlib
-
-        return zlib.decompress(data)
+        zlib = self.zlib
```
jpivarski, line 94: The only reason it was imported like this was for symmetry with the other libraries.

src/uproot/extras.py
```diff
@@ -160,70 +160,55 @@ def xrootd_version():
...
-
-    pip install lz4 xxhash
-
-or
-
-    conda install lz4 python-xxhash"""
-        ) from err
-    else:
-        return lz4.block
+        from isal import isal_zlib as zlib
```
jpivarski, line 168: Oh, that's why it's coming from extras now.

It's a bad idea to choose a library based on what exists on the system—it will work slightly differently from `zlib`, and maybe it will have its own bugs. This would make a bug on one system non-reproducible on another system unless the bug-reporter uses environments correctly and reports the whole environment in which the test was performed.

If we're going to have this capability, it has to be opt-in, within Python so that you can do it in Jupyter (i.e. not based on environment variables). For writing, we can define a new `Compression` subclass with a different name, so that users can say

```python
with uproot.recreate("out.root", compression=uproot.ISAL_ZLIB(3)) as file:
    ...
```

and it would use `isal_zlib` to do the compression and report it as ZLIB in the ROOT file.

If `isal_zlib` is desired for reading, that's more complicated because the decompressor is chosen based on a two-byte sequence before each compressed block. (It has nothing to do with what `fCompress` is set to in the TFile, TTree, or TBranch level.) The only way to inject the user's choice would be to have a global setting,

```python
uproot.ZLIB.is_isal = True
```

Although now that I look at it, it's not a bad interface.

src/uproot/compression.py
```diff
@@ -115,6 +120,9 @@ def level(self):
         """
         The compression level: 0 is uncompressed, 1 is minimally compressed, and
         9 is maximally compressed.
+
+        This value to adapted to the ISAL compression levels if that library is used.
+        Note: with ISAL 0 is lowest compression, not uncompressed!
```
jpivarski, line 125: With ISAL being strictly opt-in, its compression levels (only relevant for writing) can be different from ZLIB, including this fact that ISAL 0 is not uncompressed. Fortunately, it's also possible to set `compression=None` in [uproot.recreate](https://uproot.readthedocs.io/en/latest/uproot.writing.writable.recreate.html) and [uproot.update](https://uproot.readthedocs.io/en/latest/uproot.writing.writable.update.html) to make this distinction.

jpivarski (commented):
Inline comments below!

src/uproot/compression.py
martindurant: > ISAL 0 is not uncompressed

I didn't know this! I suppose when no-compressor is also an option, it doesn't come up often.

jpivarski: (I hadn't known it either, before reading @lgray's docstring.)

Actually, the `fCompress` variable encoded in ROOT files can't express "ZLIB level 0, distinct from no compression," either. (There is no [EAlgorithm](https://root.cern.ch/doc/master/structROOT_1_1RCompressionSetting_1_1EAlgorithm.html#ad5ed4ef81888f5406ae0f018ff45fd96) for "no compression," so if you want no compression, you set the algorithm to "ZLIB" and the level to "0".) Fortunately, that declaration is meaningless (aspirational?). Whether a block of data is compressed or not is determined by whether the compressed size is equal to the uncompressed size (because a ROOT-writer would never write a compressed block if it happens to be larger than an uncompressed one). For small and other not-very-compressible data, blocks might even alternate arbitrarily between compressed and uncompressed.

When writing, Uproot (not ROOT) has an API that can distinguish between intending to use such-and-such an algorithm with such-and-such a level or intending to not compress at all. So we can make all blocks uncompressed and report it in `fCompress` as "ZLIB level 0."

martindurant: > whether the compressed size is equal to the uncompressed size

So actually zlib(level=0) will never happen anyway, since they will actually always be slightly bigger than the original, with the addition of the block header.

lgray:
Thanks for the comments - I'll clean things up.

src/uproot/extras.py
lgray: Yea I think `uproot.ZLIB.use_isal = True` or what you wrote would be a good way to go.

`use_isal` would be opt and and attempt only that import if set, and error if not available, and would properly affect compression and decompression.

lgray:
@jpivarski after these changes I think this PR is in a pretty good place - I'm going to pop it out of draft mode. Thanks!

jpivarski:
I see that cramjam 2.8.1 will be out in minutes: https://github.com/milesgranger/cramjam/issues/126#issuecomment-1912301225. As long as this PR smoothly transitions from the experimental namespace to the eventual namespace, then it will be future-proof and ready to go into Uproot.

It's something that should go into a minor release, Uproot 5.3.0, just as the fsspec stuff needed a minor release. For fsspec, we kept two "main" branches so that Luis could get all of his updates in without affecting the 5.1.x series. This is simpler: it's just one PR. (I also doubt there will be any merge conflicts in the near future—this PR only affects compression, which is separated out from everything else.)

Uproot doesn't have a minor release schedule the way that Awkward Array does, but Awkward Array's policy of giving each minor release (at least) 2 months is probably a good idea. [Uproot 5.2.0](https://github.com/scikit-hep/uproot5/releases/tag/v5.2.0) was released on December 14, so I just made a calendar reminder to myself to merge this PR and release Uproot 5.3.0 on February 14.

lgray:
OK - lzma from cramjam is now in. This one is ready for any final review.

src/uproot/compression.py
```diff
@@ -138,9 +158,21 @@ class _DecompressLZMA:
...
+        cramjam = uproot.extras.cramjam()
+        lzma = None
+        if hasattr(cramjam, "experimental") and hasattr(cramjam.experimental, "lzma"):
+            lzma = cramjam.experimental.lzma
+        elif hasattr(cramjam, "lzma"):
+            lzma = cramjam.lzma
+        else:
+            raise RuntimeError(
+                "lzma not found in the cramjam package! (requires cramjam >= 2.8.1)"
+            )
```
jpivarski, line 170: Slick version (also prioritizing primary/eventual namespace over the experimental namespace):

```suggestion
        lzma = (
            getattr(cramjam, "lzma", None)
            or getattr(getattr(cramjam, "experimental", None), "lzma", None)
        )
        if lzma is None:
            raise RuntimeError(
                "lzma not found in the cramjam package! (requires cramjam >= 2.8.1)"
            )
```

src/uproot/compression.py
```diff
@@ -175,8 +207,16 @@ def level(self, value):
...
```
jpivarski, line 219: Same.

```suggestion
        lzma = (
            getattr(cramjam, "lzma", None)
            or getattr(getattr(cramjam, "experimental", None), "lzma", None)
        )
        if lzma is None:
            raise RuntimeError(
                "lzma not found in the cramjam package! (requires cramjam >= 2.8.1)"
            )
```

jpivarski (approved):
I'm in favor of this. See https://github.com/scikit-hep/uproot5/pull/1090#issuecomment-1912316791 for the schedule of when it would get merged and into a new release.

lgray:
@jpivarski one little tweak - I moved cramjam to be a required dependency since it is used for lzma instead of the python standard library now.

jpivarski:
I came across issue #518, which is asking for a way to use another alternate compression algorithm compatible with ZLIB's format.

Before merging this PR, I'd like to replace the

```python
ZLIB.use_isal: bool
```

with

```python
ZLIB.library: Literal["zlib", "isal"]
```

so that libraries like `"deflate"` or `"awesomest_zlib_ever"` can be added to that list, just to keep it open to the possibility.

lgray:
Sure - please feel free to edit.

martindurant:
> libraries like "deflate"

I was about to say that! "deflate" is the compression method of the data itself (by whatever alforithms), and zlib adds the header. zlib in theory supports other compressors, but in practice it, zip and gzip always do DEFLATE.

jpivarski:
I had heard of the zlib-deflate distinction before, but I thought what was being discussed in #518 was a particular Python library called `"deflate"` (which is somehow better). ROOT has its own alternate headers, for some level of headers (the compressed blocks that it uses aren't like standalone ZIP files). Maybe I'm confusing headers with headers.

So this is for `"awesomest_zlib_ever"`, then.

jpivarski:
Now this should be ready to merge, except that we won't do that until close to Feb 14: https://github.com/scikit-hep/uproot5/pull/1090#issuecomment-1912316791.

Tangentially, if anyone runs the tests in parallel mode, it's possible for some of the ISAL tests to be executed as non-ISAL and vice-versa. Such a test run shouldn't fail because the two algorithms are supposed to be equivalent, but it wouldn't be testing what we want to test. Fortunately, CI runs non-parallel tests, so it always gets tested properly at some point.

PR #1094, opened 2024-01-23
feat: allow user to supply tuple of allowed exceptions

douglasdavis:
Change `uproot.dask`'s `allow_read_errors_with_report` argument to also allow a tuple of exceptions to be passed in. Original `=True` behavior keeps the `(OSError,)` only tuple

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -908,8 +908,8 @@ class UprootReadMixin:
 
     @property
     def allowed_exceptions(self):
-        if isinstance(allow_read_errors_with_report, tuple):
-            return allow_read_errors_with_report
+        if isinstance(self.allow_read_errors_with_report, tuple):
+            return self.allow_read_errors_with_report
         return (OSError,)
 
     @property
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -96,10 +96,12 @@ def dask(
         form_mapping (Callable[awkward.forms.Form] -> awkward.forms.Form | None): If not none
             and library="ak" then apply this remapping function to the awkward form of the input
             data. The form keys of the desired form should be available data in the input form.
-        allow_read_errors_with_report (bool): If True, catch OSError exceptions and return an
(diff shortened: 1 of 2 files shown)


src/uproot/_dask.py
```diff
@@ -96,10 +96,12 @@ def dask(
         form_mapping (Callable[awkward.forms.Form] -> awkward.forms.Form | None): If not none
             and library="ak" then apply this remapping function to the awkward form of the input
             data. The form keys of the desired form should be available data in the input form.
-        allow_read_errors_with_report (bool): If True, catch OSError exceptions and return an
-            empty array for these nodes in the task graph. The return of this function then
-            becomes a two element tuple, where the first return is the dask-awkward collection
-            of interest and the second return is a report dask-awkward collection.
+        allow_read_errors_with_report (bool or tuple or exceptions): If True, catch OSError
```
jpivarski, line 99:
```suggestion
        allow_read_errors_with_report (bool or tuple of exceptions): If True, catch OSError
```

src/uproot/_dask.py
```diff
@@ -902,15 +904,17 @@ class UprootReadMixin:
...
     @property
     def allowed_exceptions(self):
+        if isinstance(self.allow_read_errors_with_report, tuple):
+            return self.allow_read_errors_with_report
         return (OSError,)
 
     @property
     def return_report(self) -> bool:
-        return self.allow_read_errors_with_report
+        return bool(self.allow_read_errors_with_report)
```
jpivarski, line 917: So `allow_read_errors_with_report=()` should be interpreted as `allow_read_errors_with_report=False`? (It sounds reasonable to me.)

jpivarski (approved):
This looks good! Are you done with it? I can merge it when you give me the go-ahead (as a comment, rather than an emoji, so that I get the email). Thanks!

src/uproot/_dask.py
douglasdavis: yes!

douglasdavis:
yep this is good to go from my end!

Issue #1099, opened 2024-01-24
can't read a branch


ivukotic:
there is a strange issue when trying to read a branch from an ATLAS physlite file. 
I use uproot 5.2.1 and awkward 2.5.2.

Here the smallest reproducible example:

```
import uproot
import awkward as ak
fn='root://xcache.af.uchicago.edu:1094//'
fn+='root://dcgftp.usatlas.bnl.gov:1094//pnfs/usatlas.bnl.gov/LOCALGROUPDISK/rucio/data18_13TeV/04/9a/DAOD_PHYSLITE.34857549._000001.pool.root.1'
with uproot.open(fn) as f:
    tree=f['CollectionTree;1']
    a=tree["PrimaryVerticesAuxDyn.trackParticleLinks"].array()
```

this file is already cached so you should be able to access it without authentication.

this is the result:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[3], line 7
      5 with uproot.open(fn) as f:
      6     tree=f['CollectionTree;1']
----> 7     a=tree["PrimaryVerticesAuxDyn.trackParticleLinks"].array()
      8     print(a.to_list()[:2])

File /venv/lib/python3.9/site-packages/uproot/behaviors/TBranch.py:1815, in TBranch.array(self, interpretation, entry_start, entry_stop, decompression_executor, interpretation_executor, array_cache, library, ak_add_doc)
   1812                 ranges_or_baskets.append((branch, basket_num, range_or_basket))
   1814 interp_options = {"ak_add_doc": ak_add_doc}
-> 1815 _ranges_or_baskets_to_arrays(
   1816     self,
   1817     ranges_or_baskets,
   1818     branchid_interpretation,
   1819     entry_start,
   1820     entry_stop,
   1821     decompression_executor,
   1822     interpretation_executor,
   1823     library,
   1824     arrays,
   1825     False,
   1826     interp_options,
   1827 )
   1829 _fix_asgrouped(
   1830     arrays,
   1831     expression_context,
   (...)
   1835     ak_add_doc,
   1836 )
   1838 if array_cache is not None:

File /venv/lib/python3.9/site-packages/uproot/behaviors/TBranch.py:3142, in _ranges_or_baskets_to_arrays(hasbranches, ranges_or_baskets, branchid_interpretation, entry_start, entry_stop, decompression_executor, interpretation_executor, library, arrays, update_ranges_or_baskets, interp_options)
   3139     pass
[...]

PR #1100, opened 2024-01-24
fix: state of context["forth"] after an entire TBasket is incomplete

jpivarski:
The TBranch described in #1099 contains many empty lists, and its AwkwardForth code can't be discovered until it sees at least one non-empty list. That happens after cycling from one TBasket to the next (about a dozen TBaskets, actually), and the `context["forth"].vm` state is `None`, rather than nonexistent, due to

https://github.com/scikit-hep/uproot5/blob/e592ae333c2d8d2ee7d65f1a2de1fcb3a8fea13f/src/uproot/interpretation/objects.py#L237

So the check for "Does an AwkwardForth VM exist?" should check for both the missing attribute case and the attribute is equal to None case.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,6 +100,7 @@ log_cli_level = "info"
 markers = [
   "slow",
   "network",
+  "distributed",
   "xrootd"
 ]
 minversion = "6.0"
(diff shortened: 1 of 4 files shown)


jpivarski:
The rest of this PR is making it possible for me to skip Dask distributed tests, which complain about Jupyter (?!?) on my computer. I've conda-updated; I don't know what's wrong with Dask, distributed, or Jupyter.

jpivarski:
@SethBendigo, if this looks good to you, give me an approval and I'll merge it.

SethBendigo (approved):

jpivarski:
Thanks!

PR #1103, opened 2024-01-25
fix: any Locks in Models must be transient

jpivarski:


ioanaif:
LGTM. However, one of the other tests times out on the Ubuntu-latest 3.8 machine; It doesn't seem to be related to this fix though

jpivarski:
Requests for remote files sometimes hang.

Thanks!

PR #1104, opened 2024-01-25
fix: better path handling in uproot.dask_write

lgray:
This is a clone of https://github.com/dask-contrib/dask-awkward/pull/459 adapted to uproot.

### DIFF
--- a/src/uproot/writing/_dask_write.py
+++ b/src/uproot/writing/_dask_write.py
@@ -36,7 +36,7 @@ class _ToROOTFn:
         filename = f"part{str(block_index[0]).zfill(self.zfill)}.root"
         if self.prefix is not None:
             filename = f"{self.prefix}-{filename}"
-        filename = f"{self.protocol}://{self.path}/{filename}"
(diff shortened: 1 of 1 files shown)


lgray:
@jpivarski this one should go in with the next release of uproot.

lgray:
Agreed - probably good for a patch release now!

jpivarski:
fsspec-xrootd file-globbing of

```
root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_*.root
```

keeps killing the GitHub actions because it's timing out. That doesn't seem to be what this PR is about. I wonder if the server is just down or slow today. I ran it in a terminal: it takes about a minute for each of the two files, which is a lot for any one test, and most of the time was not spent receiving data in iftop.

If it's not what this is about, I'll merge this PR as-is, but I always do a test on `main` before releasing, and it might fail again there. If so, I'll add a `@pytest.mark.skip(reason="flaky server")` to that test.

PR #1105, opened 2024-01-26
chore: add dask_write to read-the-docs

zbilodea:


jpivarski (approved):
This is good; we want these fixes. And it's simple enough that I know I can just merge this.

(I'm going around and checking for easy-to-merge PRs.)

jpivarski:
I'm merging this even though it has these specific test failures because of the reasoning described in https://github.com/scikit-hep/uproot5/pull/1104#issuecomment-1912432150.

PR #1120, opened 2024-02-05
feat: expose decompression_executor and interpretation_executor in uproot dask

ioanaif:


ioanaif:
This PR doesn't add tests; however, it was tested locally to make sure the executors are exposed correctly. 

I used the tests in file `test_1058_dask_awkward_report.py` to which I added a `decompression_executor` and an `interpretation_executor` with the `submit` function modified to change a variable when called. 

All seems to be working well :)

jpivarski (commented):
I would expect this:

```python
class TestExecutor(uproot.source.futures.TrivialExecutor):
    def __init__(self):
        self.submit_count = 0
    def submit(self, task, /, *args, **kwargs):
        self.submit_count += 1
        super().submit(task, *args, **kwargs)

testexecutor = TestExecutor()

a = uproot.dask({
    skhep_testdata.data_path("uproot-Zmumu-zlib.root"): "events"
}, compression_executor=testexecutor)

a["pt1"].compute()
```

to result in a `testexecutor.submit_count` that is not `0`. (Same for `interpretation_executor`.)

Am I testing it incorrectly? What did you test to see that it worked?

In fact, if I replace the counter with `raise Exception`, `a["pt1"].compute()` returns a result without anything raising an exception. Uproot is not calling `submit` on the provided executor.

jpivarski (approved):
Okay, great! Looking at the differences since last time, it was just a matter of passing the arguments further down (into those "UprootOpenAndRead" classes), not that the executors aren't working even on eager reads.

This is good to merge! Squash-and-merge whenever you're ready!

nsmith-:
The Protocol was not updated alongside the implementations
https://github.com/scikit-hep/uproot5/pull/1120/files#diff-6780d7d16e83cfee3f1d68397d1301711fd5b0bbe56633a2fe17e0de6f18d8d3R830-R837
Are we doing any type-checking in the linting? I see a lot of typing in this file but `mypy` has a bad time running on it

PR #1123, opened 2024-02-07
fix: test suite creates files not in tmp_path

ioanaif:
In testing fsspec writing for configurations in scheme + slash_prefix a "\\" directory gets made with what should be temporary files. 

Moving to the tmp_path before creating the file doesn't work, thus we can simply remove the newly created folder. 

We can do this immediately after testing each file (`test_fsspec_writing_local_uri` in `test_0692_fsspec_writing`) or after all the versions of that test have been executed.


jpivarski (approved):
This is good; it's ready to be merged.

PR #1127, opened 2024-02-15
fix: allow colon in key names

ioanaif:
This adds fixes the infinite recursion issue when trying to add keys which have a colon in their name: 

https://github.com/scikit-hep/uproot5/issues/974

If we have a file with a key name "one:two", but also a TTree "one" which contains a "two" named branch:

```
 with uproot.recreate(newfile) as f:
        f["one:two"] = "together"
        array = ak.Array(["one", "two", "three"])
        f["one"] = {"two": array}

```
 the behaviour will be the following when reading the two:
```
 with uproot.open(newfile) as f:
        f["one:two"] == "together"
        f["one"]["two"].array() == ["one", "two", "three"]

```


jpivarski (approved):
We talked about it in our meeting, and this is a great solution. It's important that it prioritizes names with colons over parsing a colon as a separator, but allows both, since this makes access to all objects possible. (`f["one:two"]` versus `f["one"]["two"]` in your test example.)

It's ready to merge when you think it is.

PR #1128, opened 2024-02-16
fix: allow writing generic UHI-compatible histograms

YSelfTool:
Commit ec80b887 broke support for writing histograms that fulfill the Unified Histogram Interface, but are neither from hist nor from boost_histogram. UHI compatible histograms always have a function `variances`, but do not have a `storage_type` attribute, so that attribute should only be checked if the histogram is indeed from boost_histogram.

### DIFF
--- a/src/uproot/writing/identify.py
+++ b/src/uproot/writing/identify.py
@@ -265,7 +265,8 @@ def to_writable(obj):
             data = obj.values(flow=True)
             fSumw2 = (
                 obj.variances(flow=True)
-                if boost_histogram is None or obj.storage_type == boost_histogram.storage.Weight
+                if boost_histogram is None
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Thanks! We do want to ensure that arbitrary UHI histograms are writable, but this slipped in because we don't have any non-boost histogram types to test. (In principle, we could mock up a new histogram type that satisfies UHI, although that would be quite a bit more work.)

If you're done with this PR and it passes all tests, I'll merge it.

jpivarski:
Let me know if you're done and I'll merge it!

jpivarski:
I think this is done. I'm merging it now. Thanks!

jpivarski:
@all-contributors please add @YSelfTool for code

PR #1144, opened 2024-02-23
feat: implement TGraph-writing

jpivarski:
This is a draft. It needs TGraphErrors, TGraphAsymmErrors, tests of writing to actual files and reading them back with ROOT, and maybe some high-level interface so that we can

```python
output_file["name"] = XYZ
```

for a Python object XYZ that would be interpreted as a TGraph (or TGraphErrors, or TGraphAsymmErrors). What should that XYZ be? Not raw NumPy arrays, since a tuple of NumPy arrays is already interpreted as a histogram. Does it need to be some kind of `uproot.make_TGraph` function? It would be nice to have something more Pythonic.

Should any Pandas DataFrame with columns named `x` and `y` be interpreted as a TGraph? A Pandas DataFrame would supply a title.

Thoughts, @grzanka?

### DIFF
--- a/src/uproot/models/TGraph.py
+++ b/src/uproot/models/TGraph.py
@@ -338,16 +338,18 @@ class Model_TGraph_v4(uproot.behaviors.TGraph.TGraph, uproot.model.VersionedMode
         where = len(out)
         for x in self._bases:
             x._serialize(out, True, name, tobject_flags)
-        out.extend([
-            struct.pack(">i", self._members["fNpoints"]),
-            b"\x01",
-            self._members["fX"].astype(">f8").tobytes(),
-            b"\x01",
-            self._members["fY"].astype(">f8").tobytes(),
-            b"@\x00\x00\x1f\xff\xff\xff\xffTList\x00@\x00\x00\x11\x00\x05\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00",
-            struct.pack(">d", self._members["fMinimum"]),
-            struct.pack(">d", self._members["fMaximum"]),
-        ])
+        out.extend(
+            [
+                struct.pack(">i", self._members["fNpoints"]),
+                b"\x01",
+                self._members["fX"].astype(">f8").tobytes(),
+                b"\x01",
+                self._members["fY"].astype(">f8").tobytes(),
+                b"@\x00\x00\x1f\xff\xff\xff\xffTList\x00@\x00\x00\x11\x00\x05\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00",
+                struct.pack(">d", self._members["fMinimum"]),
+                struct.pack(">d", self._members["fMaximum"]),
+            ]
+        )
         if header:
             num_bytes = sum(len(x) for x in out[where:])
             version = 4
--- a/src/uproot/models/TGraph.py
+++ b/src/uproot/models/TGraph.py
@@ -338,7 +338,16 @@ class Model_TGraph_v4(uproot.behaviors.TGraph.TGraph, uproot.model.VersionedMode
         where = len(out)
         for x in self._bases:
             x._serialize(out, True, name, tobject_flags)
-        raise NotImplementedError("FIXME")
+        out.extend([
+            struct.pack(">i", self._members["fNpoints"]),
+            b"\x01",
+            self._members["fX"].astype(">f8").tobytes(),
+            b"\x01",
+            self._members["fY"].astype(">f8").tobytes(),
+            b"@\x00\x00\x1f\xff\xff\xff\xffTList\x00@\x00\x00\x11\x00\x05\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00",
+            struct.pack(">d", self._members["fMinimum"]),
+            struct.pack(">d", self._members["fMaximum"]),
+        ])
         if header:
             num_bytes = sum(len(x) for x in out[where:])
             version = 4


grzanka:
> Should any Pandas DataFrame with columns named `x` and `y` be interpreted as a TGraph? A Pandas DataFrame would supply a title.
> 
> Thoughts, @grzanka?

Another candidate may be a python dataclass with `x` and `y` fields containing arrays of numbers ? Or numpy recarray ?

grzanka:
As discussed here: https://github.com/scikit-hep/uproot5/discussions/1142#discussioncomment-8590182 we may create confusion by considering an object that can be converted both to TTree and TGraph.

jpivarski:
It can't be a dict because all dicts are already interpreted as TTree (as a set of TBranch names to arrays).

Adding attributes, as in point 5, is somewhat cumbersome—it would have to be extra lines of Python, hard to do interactively—and hidden. You would just need to know the name, and even if you know that there is a special name, you might need to look it up to get the right spelling.

How about a DataFrame? It can be recognized if it has column names like `x`, `y`, `errors_x`, `errors_y`, `errors_x_high`, `errors_x_low`, `errors_y_high`, `errors_y_low`, with rules for how `errors_x_high` and `errors_x_low` overrule `errors_x` (or raise an error if both are present), and rules for whether to make a TGraph, TGraphErrors, or TGraphAsymmErrors.

DataFrames (just Pandas or all of the new ones, Polars, CuDF, Modin? Can they be recognized in bulk?) are the basic way of describing scatter-point data in Python, and the TGraph* classes are the basic way of describing scatter-point data in ROOT.

Pepesob:
Isn't there the same problem with DataFrame as it is with Dictionariy? Right now pandas DataFrame is transformed into dictionary and then interpreted as TTree.
https://github.com/scikit-hep/uproot5/blob/0a997cd40e7661850f1d426f596cab615926a854/src/uproot/writing/identify.py#L53-L59
If I understand correctly you suggest to add logic, so whenever there are columns in DataFrame with given names `{x, y, errors_x, etc... }`, then recognize DataFrame as TGraph (or TGraphErrors, TGraphAsymmErrors)?

jpivarski:
We could catch the DataFrame type before asking if it's a Mapping. This would take away some of the types that are currently being identified as TTrees, but accidentally. There needs to be simple, easy-to-remember criteria for what will become a TTree and what will become a TGraph*, and DataFrame vs other Mapping could be that criteria. Presence of a special attribute is more subtle.

Pepesob:
I think that always converting DataFrame into TGraph is a good idea, because usually DataFrame do not have entries of variable size in column (at least in pandas), so they could be rightly interpreted as [x,y] (+errors) coordinates. However wouldn't that change break people existing code that transforms pandas DataFrame into TTree? I would still propose the flag `implicit_DataFrame_TGraph_convertion` to give choice to people.

jpivarski:
I had forgotten that DataFrame was a major, recommended way to make TTrees (not just an accident of DataFrames being mappings).

Okay.

But still, setting a flag like `implicit_DataFrame_TGraph_convertion` on the DataFrame (can't do it to a Python dict) is really obscure. It's hard to memorize; if I type `implicitly_DataFrame_TGraph_convertion`, it won't work.

What about this?

```python
f = uproot.recreate("/tmp/stuff.root")
df = pd.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})
f["ttree"] = df
f["tgraph"] = uproot.as_TGraph(df)
```

Then the user only needs to remember a function named `uproot.as_TGraph`, and it might be discoverable by tab-complete. ("What was it? `uproot.<tab>` (lots of things), `uproot.as_<tab>`, ah, there it is!")

This function _could_ just add the flag, and a DataFrame/namedtuple with that flag is recognized as a TGraph. So it's possible to do both.

Pepesob:
Okay there is slight misconception with the flag. What i mean was something like this:
```python
implicit_DataFrame_TGraph_convertion = True

def set_implicit_DataFrame_TGraph_convertion(value: bool):
    if not isinstance(value, bool):
        raise ValueError("Flag must be boolean")
    
    global implicit_DataFrame_TGraph_convertion
    implicit_DataFrame_TGraph_convertion = value

```

and then in the add_to_directory function
```python
if implicit_DataFrame_TGraph_convertion and uproot._util.is_dataframe(obj):
    if uproot._util.from_module(obj, "pandas"):
        import pandas
        [...]
```
So it is up to the user what behaviour he wants, implicit DataFrame conversion to TGraph or not. But let's move on.

Now I will refer to my 2. proposition point. 
```python
f = uproot.recreate("/tmp/stuff.root")
df = pd.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})
f["ttree"] = df
f["tgraph"] = uproot.as_TGraph(df)
```

`uproot.as_TGraph(df)` would return something like `TGraphConvertable` and then this object would be written to file.

On a second thought, there is something like this `uproot.models.TGraph.Model_TGraph_v4`
https://github.com/scikit-hep/uproot5/blob/0a997cd40e7661850f1d426f596cab615926a854/src/uproot/models/TGraph.py#L29-L344

So `uproot.as_TGraph(df)` would just return `uproot.models.TGraph.Model_TGraph_v4` and this object would be written to file. Moreover this solution won't require any changes in [identify.py](https://github.com/scikit-hep/uproot5/blob/main/src/uproot/writing/identify.py) because `uproot.models.TGraph.Model_TGraph_v4` is Model object, and for those there is already logic in [identify.py](https://github.com/scikit-hep/uproot5/blob/main/src/uproot/writing/identify.py) . At least i think so. I need to doulbe check it but i think this would work.

jpivarski:
I didn't realize that you were suggesting a _global_ flag, but we shouldn't do that for multiple reasons. (What if the user wants to write TTrees _and_ TGraphs? What if the user is not using Uproot directly, but through a library and that library changes this state? What if the user is using a library of a library, and one level does it and another doesn't know that it has happened?)

But having `uproot.as_TGraph` create a `uproot.models.TGraph.Model_TGraph_v4` is a direct way to do it, if you're not interested in the per-instance flag.

Pepesob:
Hey @jpivarski! I implemented `to_TGraph()` function. I also wrote test for this feature. What should be my next step? Should I create pull request to this branch (jpivarski/start-implementation-of-TGraph-writing) or directly to main?

jpivarski:
To simplify the git topology, please open a PR for inclusion into `main`.

Is your branch derived from this branch? If so, then I can close this PR in favor of yours and @grzanka's commits should still be in it, so that the final PR will count as being co-authored by the two of you.

Pepesob:
Yes I made branch derived from this one

PR #1146, opened 2024-02-23
feat: split ranges for large files read over http

ioanaif:
Since uproot 5.2.0 a request is sent for every chunk (with fsspec). Due to this, when working with large files, the server disconnects immediately. 

This PR adds (to the http source) the option to set the length of the `range_header` which will then divide the request into smaller chunks of length = `range_header_len`.

This option works with the source being set to http:
```f = uproot.open(fname, handler=uproot.source.http.HTTPSource, range_header_len=21786)```

### CODE REVIEW

File: src/uproot/reading.py [hunk 1]
```diff
@@ -163,6 +163,7 @@ def open(
     "num_fallback_workers": 10,
     "begin_chunk_size": 403,  # the smallest a ROOT file can be
     "minimal_ttree_metadata": True,
+    "range_header_len": 21784,
```
jpivarski, line 166: Just one thing: since this only applies to HTTP, the option should have `http_` in its name. Also, it's a maximum length, rather than a length (measured in bytes), so maybe

```suggestion
    "http_max_header_bytes": 21784,
```

with appropriate changes to pick up the name modification elsewhere?

src/uproot/source/http.py
```diff
@@ -579,6 +579,8 @@ def __init__(self, file_path: str, **options):
         self._fallback_options = options.copy()
         self._fallback_options["num_workers"] = self._num_fallback_workers
 
+        self._range_header_len = options["range_header_len"]
```
jpivarski, line 582:
```suggestion
        self._http_max_header_bytes = options["http_max_header_bytes"]
```

src/uproot/source/http.py
```diff
@@ -624,29 +626,63 @@ def chunk(self, start: int, stop: int) -> uproot.source.chunk.Chunk:
...
+
+            i, j = 1, 0
+            range_header = {"Range": "bytes=" + f"{ranges[0][0]}-{ranges[0][1] - 1}"}
+            last_batch_appended = False
+
+            while i < len(ranges):
+                new_range_to_append = ", " + f"{ranges[i][0]}-{ranges[i][1] - 1}"
+                if len(range_header["Range"]) < self._range_header_len - len(
+                    new_range_to_append
+                ):
```
jpivarski, line 663:
```suggestion
                if len(range_header["Range"]) < self._http_max_header_bytes - len(
                    new_range_to_append
                ):
```

tests/test_9999_split_ranges_for_large_files_over_http.py
```diff
@@ -0,0 +1,28 @@
...
+        "Jet_phi",
+        "Jet_btagCSVV2",
+        "Jet_pt",
+        "Jet_eta",
+    ]
+
+    f = uproot.open(
+        fname, handler=uproot.source.http.HTTPSource, range_header_len=21786
+    )
+    assert f.file.options["range_header_len"] == 21786
```
jpivarski, line 26: I think this is the last one (by search-and-replace):

```suggestion
    f = uproot.open(
        fname, handler=uproot.source.http.HTTPSource, http_max_header_bytes=21786
    )
    assert f.file.options["http_max_header_bytes"] == 21786
```

jpivarski (approved):
Nice! I'm glad that `ResourceClass.multifuture` can be repeatedly called like this, which simplifies this implementation.

I think this is ready to be merged, once the name for the new option changes (as described below).

PR #1149, opened 2024-02-24
chore: drop dep that never is valid

henryiii:
The condition here is exactly the package condition. :)


henryiii:
FWIW, you can now get rid of the rename https://github.com/scikit-hep/uproot5/blob/ef251ef8a1eb98fa6ee35c1c747f00e82af228b6/src/uproot/extras.py#L13 too.

jpivarski (approved):
Okay! If you're done, I'll enable auto-merge.

Thanks!

henryiii:
Sure!

PR #1154, opened 2024-02-29
feat: allow paths in classname_of, streamer_of, class_of and title_of

ioanaif:
This PR allows passing paths into `classname_of`, `streamer_of`, `class_of`  and `title_of`. Thus, the following path sent as argument will return "TBranch": 

```
with  uproot.open("test.root") as f:
        assert f.classname_of("Tree/x") == "TBranch"
        assert f.classname_of("Tree:x") == "TBranch"
        
```


jpivarski (approved):
It's good! I made some edits:

* elsewhere (`keys`, `values`, `items`), an argument very much like this is called `recursive`, so I changed `path_lookup` to `recursive`
* the test was making test.root files in the working directory; need to use `tmp_path`.

If you have no counter-edits, go ahead and merge.

PR #1155, opened 2024-02-29
feat: add branches to existing TTree

### LINKED ISSUE #862: Take an existing TTree and add a new TBranch
This would be a major project, but it's frequently asked for. Users want to be able to do this:

```python
with uproot.update("existing_file.root") as file:
    assert file["existing_tree"].num_entries == len(array)
    file["existing_tree"].new_branch(array)
```

such that the `"existing_tree"` has the same number of entries but a new TBranch.

This will involve overwriting the entire TTree metadata object (or creating one with a new cycle number), but it can reuse all of the TBasket data from the previous TTree, only adding new ones.

zbilodea:
Fixes #862 (when it's done).


ariostas:
Superseded by #1669

PR #1156, opened 2024-03-01
feat: Record source performance counters in uproot.dask report

nsmith-:
Report type is now
```
type: 7 * {
    call_time: ?unknown,
    duration: float64,
    performance_counters: {
        num_requested_bytes: int64,
        num_requests: int64,
        num_requested_chunks: int64
    },
    args: var * string,
    kwargs: var * unknown,
    exception: ?unknown,
    message: ?unknown,
    fqdn: ?unknown,
    hostname: ?unknown
}
```

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -6,6 +6,8 @@ import socket
 import time
 from collections.abc import Callable, Iterable, Mapping
 
+from uproot.source.chunk import SourcePerformanceCounters
+
 try:
     from typing import TYPE_CHECKING, Final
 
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
I scanned over all of the changes. It looks like it does what you need, and I don't see any way that this could negatively impact any other parts of Uproot. So I'd say it's ready to merge!

In your example type, there were a lot of `?unknown` fields. They must have been always `None` in the example that you ingested with `ak.from_iter` and checked the type of.

nsmith-:
Yes, those are fields that are only set for failed files. I could force a failure to show the full type; it is a union of types as defined by `_report_success` and `_report_failure`.

PR #1158, opened 2024-03-03
feat: Add extras for optional remote access dependencies

### LINKED ISSUE #1153: Use extras so downstream packages can depend on `uproot[xrootd]`
I couldn't find anything discussing this but I'd find it useful if there was an extra that let me declare the need for `xrootd` support. I could also add a dependency on `fspec-xrootd` but it feels a bit wrong to specify an implementation detail of `uproot`.

I guess the same might also make sense for `s3` and maybe a generic option for all compression related dependencies (less sure about the name for this one, maybe `[compression]`?).

chrisburr:
This makes it possible to install/depend on `uproot[xrootd,http,s3]` to decouple the user facing feature from the implementation details in uproot.

Closes #1153

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,8 @@ dev = [
   "pandas",
   "awkward-pandas"
 ]
+http = []
+s3 = ["minio"]
 test = [
   "isal",
   "xxhash",
@@ -77,6 +79,7 @@ test = [
   "scikit-hep-testdata",
   "rangehttpserver"
 ]
+xrootd = ["fsspec-xrootd"]
 
 [project.urls]
 Download = "https://github.com/scikit-hep/uproot5/releases"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ dev = [
   "awkward-pandas"
 ]
 http = []
-s3 = ["minio"]
+s3 = ["aiobotocore"]
 test = [
   "isal",
   "xxhash",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,8 +62,8 @@ dev = [
   "pandas",
   "awkward-pandas"
 ]
-http = []
-s3 = ["aiobotocore"]
+http = ["aiohttp"]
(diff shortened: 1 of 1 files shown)


nsmith-:
Since fsspec provides an s3 solution through aiobotocore, maybe we should be installing aiobotocore rather than minio? Minio is what is used by the older s3 source.

pyproject.toml
```diff
@@ -62,6 +62,8 @@ dev = [
   "pandas",
   "awkward-pandas"
 ]
+http = []
+s3 = ["minio"]
```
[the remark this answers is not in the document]
nsmith-: Ah sorry, I just realized it is actually `s3fs` that fsspec requires installed, which in turn depends on `aiobotocore`:
```
$ python3 -m venv tmpvenv
$ . tmpvenv/bin/activate
(tmpvenv) $ pip install fsspec
...
(tmpvenv) $ python
Python 3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import fsspec
>>> fsspec.fs("s3")
...
ImportError: Install s3fs to access S3
```

pyproject.toml
```diff
@@ -62,6 +62,8 @@ dev = [
   "pandas",
   "awkward-pandas"
 ]
+http = []
```
jpivarski, line 65: I think this is what fsspec needs for HTTP access.

```suggestion
http = ["aiohttp"]
```

jpivarski (approved):

pyproject.toml
```diff
@@ -62,6 +62,8 @@ dev = [
...
+s3 = ["aiobotocore"]
```
chrisburr, line 66:
```suggestion
s3 = ["s3fs"]
```

chrisburr:
Is this still waiting for anything?

jpivarski:
You're right—this is done; thanks for reminding me!

PR #1160, opened 2024-03-04
feat: special case for reading std::string from a TDirectory

jpivarski:
~~Fixes #1159.~~ (That's a discussion.)

What do you think, @chrisburr? Is `std::string` truly exceptional, such that a special case is warranted, or is this the tip of an iceberg that should be handled in a more systematic way?

(If it's to be handled in a systematic way, I don't know when or by whom.)

### DIFF
--- a/src/uproot/models/TString.py
+++ b/src/uproot/models/TString.py
@@ -104,3 +104,4 @@ in file {self.file.file_path}"""
 
 
 uproot.classes["TString"] = Model_TString
+uproot.classes["string"] = Model_TString
--- a/src/uproot/models/TString.py
+++ b/src/uproot/models/TString.py
@@ -104,4 +104,3 @@ in file {self.file.file_path}"""
 
 
 uproot.classes["TString"] = Model_TString
-uproot.classes["string"] = Model_TString
--- a/src/uproot/reading.py
+++ b/src/uproot/reading.py
@@ -2473,12 +2473,8 @@ class ReadOnlyKey:
         else:
             chunk, cursor = self.get_uncompressed_chunk_cursor()
             start_cursor = cursor.copy()
-            context = {"breadcrumbs": (), "TKey": self}
-
-            if self._fClassName == "string":
-                return cursor.string(chunk, context)
-
             cls = self._file.class_named(self._fClassName)
+            context = {"breadcrumbs": (), "TKey": self}
 
             try:
                 out = cls.read(chunk, cursor, context, self._file, selffile, parent)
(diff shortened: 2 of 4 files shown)


jpivarski:
Actually, this second commit is less invasive, and it's setting up less of a precedent for special cases. The serialization of the `std::string` is just like a `TString` (not `TObjString`), so maybe we can put it in the lookup table under both names.

The downside of this approach is that it returns the data as an instance of a class named `Model_TString`. Although this is a subclass of `str` and you can easily get at the data, it could be misleading to call this "`TString`", given that it's a `std::string` in C++.

jpivarski:
Can we include your file as a test? Is https://cburr.web.cern.ch/string-example.root a stable address?

chrisburr:
Thank you for following up with this so quickly. I've tried this out and it works exactly the way I'd expect:

```python
In [1]: import uproot
   ...: import json
   ...: uproot.classes["string"] = uproot.classes["TString"]
   ...:
   ...: fn = "https://cburr.web.cern.ch/string-example.root"
   ...:
   ...: f2 = uproot.open(fn)
   ...: json.loads(f2["FileSummaryRecord"])
   ...:
Out[1]:
{'LumiCounter.eventsByRun': {'counts': {},
  'empty': True,
  'type': 'LumiEventCounter'},
 'guid': '5FE9437E-D958-11EE-AB88-3CECEF1070AC'}
```

I suspect `string` is exceptional and at a glance there seems to be dedicated code for serialising it:

https://github.com/root-project/root/blob/e7d7ee19abd630b0df3a16d84d61bb59e2e9b357/core/meta/src/TClass.cxx#L648-L724

> Can we include your file as a test? Is https://cburr.web.cern.ch/string-example.root a stable address?

It can absolutely be included as a test though I'd suggest putting it somewhere else if you can as that URL likely won't last forever. It's only ~5KB.

jpivarski:
Well, if it's a special case in ROOT, it can be a special case in Uproot.

I'm still not sure if I like the

```python
            if self._fClassName == "string":
                return cursor.string(chunk, context)
```

or the

```python
uproot.classes["string"] = Model_TString
```

better. The first option is "more hacky," or at least, more special of a special case, and the second is natural but has the downside of declaring that a `std::vector` is a `TString`.

Actually, that made my decision for me: we can't have Uproot insinuating that a `std::vector` is a `TString`. That kind of (deliberate!) confusion might mess with someone's debugging someday. At least the explicit if statement calls itself out as an obvious wart.

ioanaif (approved):

PR #1161, opened 2024-03-05
feat: add option to use deflate lib

ioanaif:


src/uproot/extras.py
```diff
@@ -168,7 +168,7 @@ def isal():
         import isal
     except ModuleNotFoundError as err:
         raise ModuleNotFoundError(
-            """install the 'cramjam' package with:
+            """install the 'isal' package with:
```
jpivarski, line 171: Thanks for catching that!

jpivarski (approved):
Great, thanks! Everything here looks good, and this is ready to merge.

PR #1166, opened 2024-03-09
fix: change expected xz/lzma module location on cramjam

milesgranger:
I've [pre-released v2.8.3-rc1](https://pypi.org/project/cramjam/2.8.3rc1/#files), which has the XZ/LZMA stuff now as `cramjam.xz`. 
Would rather not break things for you, so until releasing `2.8.3`, I think this change, or something like it, should happen here or >= 2.8.3 will look like it doesn't have xz/lzma support.

xref https://github.com/milesgranger/cramjam/issues/126

PR #1167, opened 2024-03-09
fix: lzma -> xz now that it is out of experimental

lgray:
small name change

see https://github.com/milesgranger/cramjam/pull/133


lgray:
Closing as duplicate of #1166

PR #1180, opened 2024-03-21
feat: add support for reading free floating std::vector

ioanaif:


jpivarski (approved):
Great! I went back and forth on the priority order; now it's

1. check for any globally- or per-file-defined Models
2. load the file's streamers and look for any Models the streamers might have defined
3. check for explicit STL classes, like `std::vector<T>` (where the `T` might have been defined by one of the streamers, so we had to check
4. give up and call it UnknownClass.

The regex is now

```
(std\s*::\s*)?(vector|map|set|string|bitset)
```

because we want exactly one of the STL class names to match.

But it looks good, the test does the right thing, and it's ready to merge.

jpivarski:
Another alteration: this is similar to #1160, but that implementation checked for `std::string` in `ReadOnlyKey.get`. It has to be in a different place: since free-floating `std::string` has to return a `str` type and not a Model, `ReadOnlyFile.class_named` is too late. The other cases, `std::vector`, `std::map`, `std::set`, and `std::bitset` (which are the only ones with parameters; `std::string` wouldn't be followed by a `<`) return container types such as `STLVector`, `STLMap`, `STLSet`, and `STLBitSet`.

However, I could make the code that handled the #1160 case more symmetric. It can check for both `string` and `std::string`.

Now it's ready to merge.

PR #1181, opened 2024-03-22
feat: add support for std::list

ioanaif:


jpivarski (approved):
Nice! The tests show that `std::list` really is serialized like a `std::vector`.

Go ahead and merge this PR!

PR #1186, opened 2024-03-26
fix: `np.dtype(branch_type)` might raise ValueError, not just TypeError

### LINKED ISSUE #1185: Error creating fixed-length array branch in new TTree
```python
import uproot
uproot.__version__
'5.3.1'
```

I am trying to make an empty TTree with both fixed-length array branches as well as variable-length array branches, and fill it in a loop with `extend`:

```python
import awkward as ak
import uproot as ur

pt = ak.Array([[0.0, 11, 22], [], [33, 44], [55], [66, 77, 88, 99]])
event_params = [[0, 0, 0], [0, 0, 1], [0, 1, 0], [0, 1, 1], [1, 0, 0]]

f = ur.recreate("f.root")
f.mktree("t", {"event_params": "3 * uint32", "pt": "var * float64"})
for i in range(5):
    f["t"].extend({"event_params": [event_params[i]], "pt": ak.Array([pt[i]])})
f.close()
```

However, this script exits with the error `ValueError: format number 1 of "3 * uint32" is not recognized`. Interestingly, if the `event_params` branch is removed the script will work with the variable length `pt` branch only.

A quick fix seems to be to catch `ValueError` thrown by `numpy.dtype()` when the `3 * uint32` datashape string is passed to it, and throw a `TypeError` instead, which seems to fix the problem. However, I don't know if this causes any side effects. Any advice on fixing this error is appreciated, thanks in advance.

```diff
diff --git a/src/uproot/writing/_cascadetree.py b/src/uproot/writing/_cascadetree.py
index 5bd2201..d25a25a 100644
--- a/src/uproot/writing/_cascadetree.py
+++ b/src/uproot/writing/_cascadetree.py
@@ -121,7 +121,10 @@ class Tree:
                         raise TypeError
                     if isinstance(branch_type, str) and branch_type.strip() == "bytes":
                         raise TypeError
-                    branch_dtype = numpy.dtype(branch_type)
+                    try:
+                        branch_dtype = numpy.dtype(branch_type)
+                    except ValueError:
+                        raise TypeError
 
                 except TypeError as err:
                     try:
```

djw9497:
Fixes #1185

### DIFF
--- a/src/uproot/writing/_cascadetree.py
+++ b/src/uproot/writing/_cascadetree.py
@@ -123,7 +123,7 @@ class Tree:
                         raise TypeError
                     branch_dtype = numpy.dtype(branch_type)
 
-                except TypeError as err:
+                except (TypeError, ValueError) as err:
                     try:
                         awkward = uproot.extras.awkward()
                     except ModuleNotFoundError as err:
--- a/tests/test_1186_dtype_might_raise_ValueError.py
+++ b/tests/test_1186_dtype_might_raise_ValueError.py
@@ -1,0 +1,23 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot5/blob/main/LICENSE
+
+import os
+
+import pytest
+import numpy as np
+
+import uproot
+
+ak = pytest.importorskip("awkward")
+
+
+def test(tmp_path):
+    filename = os.path.join(tmp_path, "whatever.root")
+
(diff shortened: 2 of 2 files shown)


jpivarski (approved):
Great, thanks!

I made a test using your minimal reproducer, so that the code stays fixed.

Let me know if you have any counter-edits; otherwise, I'll merge it when it passes tests.

jpivarski:
@all-contributors please add @djw9497 for code

PR #1189, opened 2024-03-28
fix: dask failing for TTrees with duplicate TBranch names

ioanaif:


src/uproot/_dask.py
```diff
@@ -1586,7 +1589,7 @@ def _get_dak_array_delay_open(
     ffile_path, fobject_path = files[0][0:2]
 
     if known_base_form is not None:
-        common_keys = list(known_base_form.fields)
+        common_keys = set(known_base_form.fields)
```
jpivarski, line 1592: We want to drop duplicates but also preserve order. This does it, but it's not concise.

```suggestion
        already_seen = set()
        common_keys = []
        for x in known_base_form.fields:
            if x not in already_seen:
                common_keys.append(x)
            already_seen.add(x)
```

jpivarski: ChatGPT suggests this:

```suggestion
        common_keys = list(dict.fromkeys(known_base_form.fields))
```

and I rather like it. It relies on the dict order being stable, which is true for all Python versions that Uproot supports. (If set order were stable, too, that would be awesome.) I hadn't known about the `dict.fromkeys` constructor, but it makes a dict in which all values are `None`. A dict in which all values are `None` is essentially an order-preserving set. Converting it back to a list is a good idea because the else branch of this if-else makes `common_keys` a list, and it's less error-prone to have a consistent type.

jpivarski (approved):
This only changes an internal interface, and even if Coffea is relying on this internal interface, it only adds a new keyword argument at the end of an argument list, whose default value is no change in behavior. It's completely safe.

Moreover, it ought to solve the problem of duplicates in `uproot.dask`, and the test demonstrate it.

I'll merge this PR after committing my one suggestion below.

PR #1193, opened 2024-04-05
ci: try pixi

henryiii:
Just curious to see if this works.


ariostas:
@henryiii do you think the [pixi-lock](https://github.com/parcels-code/pixi-lock) action would help here? (it's mentioned in https://pixi.prefix.dev/latest/workspace/lock_file/#additional-considerations-for-libraries).

Issue #1157, opened 2024-03-03
Performance degradation reading files via root://


chrisburr:
Since the move to using fsspec-xrootd the performance of reading files via XRootD has dropped. For example, here is a script which pulls a few bytes of metadata from an LHCb file:

<details><summary><code>uproot-perf.py</code></summary>
<p>

```python
#!/usr/bin/env python
import argparse
import time

import uproot
import uproot.source.xrootd


def run_test(pfn, handler):
    f = uproot.open(pfn, handler=handler)

    start = time.perf_counter()
    f["Refs"]["Params"].array()
    print("Took", time.perf_counter() - start, "seconds")


def parse_args():
    parser = argparse.ArgumentParser()
    parser.add_argument("pfn")
    parser.add_argument("handler", choices=["default", "xrootd"])
    args = parser.parse_args()

    handler = {
        "default": None,
        "xrootd": uproot.source.xrootd.XRootDSource,
    }[args.handler]
    run_test(args.pfn, handler)


if __name__ == "__main__":
    parse_args()
```

</p>
</details> 

```bash
# Use some LHCb open data as an example
$ export pfn='root://eospublic.cern.ch//eos/opendata/lhcb/Collision12/CHARM/LHCb_2012_Beam4000GeV_VeloClosed_MagDown_RealData_Reco14_Stripping21_CHARM_MDST/00041836/0000/00041836_00009718_1.charm.mdst'

# Using the default
$ ./uproot-perf.py $pfn default
Took 0.17654120799999995 seconds

# Using the old XRootDSource
$ ./uproot-perf.py $pfn xrootd                                                                                                                                                                    
Took 0.07812191700000004 seconds
```

The effect is less obvious for me with eospublic as I'm only a few milliseconds away from it. If I use data stored further away from me, I get 0.69s vs 2.7s.

I can also see the overhead when pulling larger amounts of data, though it's a smaller effect.

I suspect this is because "vector reads" are no longer being used so you have the roundtrip latency and overhead.

chrisburr:
From a bit more debugging, it looks like it's reading way more data than it needs to, see the `size:` parts of below:

```bash
$ XRD_LOGLEVEL=Debug python /tmp/uproot-perf.py $pfn default 2>&1 | rg '(read command|Took )' -A 1
[...]

PR #1198, opened 2024-04-17
feat: Implement read coalescing algorithm

nsmith-:
Tuning to be investigated


nsmith-:
The removal of the stateful handle is motivated by https://github.com/scikit-hep/uproot5/issues/1157#issuecomment-1981832282

nsmith-:
Only in python 3.11+ do we see
```
E               ValueError: not a ROOT file: first four bytes are b'\xec_wu'
E               in file s3://pivarski-princeton/pythia_ppZee_run17emb.picoDst.root
```
looking into it...

nsmith-:
The problem is the signature of `fs.cat_file`. For the file
```python
import fsspec

fs, path = fsspec.url_to_fs("s3://pivarski-princeton/pythia_ppZee_run17emb.picoDst.root", anon=True)
```

If I create a file-object handle and read the first 4 bytes, I get:
```python
fh = fs.open(path)
fh.seek(0)
assert fh.read(4) == b"root"
```

if I instead use `cat_file`, I find:
```python
data = fs.cat_file(path, 0, 4)
assert len(data) == 4  # fails, len(data) is 978299156
assert data == b"root" # also fails, data[:4] is b'\x00\x00\xf1l'
```

So there is some mismatch in the behavior that I was initially puzzled by. Then I realized that, depending on the implementation, `fs.cat_file` can have different signatures and the s3fs one has a different order of keyword arguments. I made a UX issue in https://github.com/fsspec/filesystem_spec/issues/1600 and used keywords here.

jpivarski (approved):
As I understand it, src/uproot/source/coalesce.py is a suite of utility functions used only by FSSpecSource to merge nearby but not necessarily adjacent byte requests—download a bit too much data, but have (far?) fewer requests and responses in flight, which is a net savings, depending on how a network's throughput and latency compare.

Presumably, it would need to be tuned for each network. How are the default settings? (It looks to me like the coalesce functions are always used by FSSpecSource, so this will affect the default case.) The most conservative default is to only coalesce adjacent byte ranges, but the most common network scenario (grad student in a basement, trying to get data from CERN) is more latency-bound than throughput-bound. Maybe the default could accept 25% over-read or so.

I see that this also only affects FSSpecSource. It could, in principle, be applied to the pre-FSSpecSources, but there's less reason to do that. FSSpecSource has the many-requests-in-flight model built into it, and the old sources attempt to do vector reads (only one request in flight).

I'm in favor of this. If @lobis is available, it would be great to hear what you think! If we don't hear from you by Thursday, I think we should go ahead with this PR.

nsmith-:
The defaults are in spirit similar to the CMSSW defaults, though the use case is certainly different. I was hoping to gain some experience in the wild how different these are.

As for implementing it in pre-fsspec sources, indeed that is an option, and one I think may be useful. For example, even in the case of vector reads (which fsspec is now doing if the protocol supports, through the `cat_ranges` interface) there may be value in issuing separate reasonably-sized vector read requests rather than one large one.

jpivarski:
It's Thursday; I think we should go ahead with this PR. I'll update to `main` and then merge.

PR #1199, opened 2024-04-24
ci: macos-latest is changing to macos-14 ARM runners

henryiii:
Committed via https://github.com/asottile/all-repos


jpivarski:
`macos-latest` works if we disable `deflate` (#1203), so I'm going to go with that instead of downgrading.

PR #1207, opened 2024-05-07
fix: https://root-forum.cern.ch/t/wrong-branch-type-detection/59236/25

jpivarski:


umute97:
Is there a way to make this available for python3.7? Unfortunately, I have to develop for this rather antique version...

jpivarski:
Uproot doesn't declare itself to be usable in Python 3.7, but it probably is. (At one point, I made it work in Python 2.6! That's no longer true, though.)

I just tried to hack it by checking out the repo from GitHub, changing the line

https://github.com/scikit-hep/uproot5/blob/6d35e1a466861211de44ffdf6e71f65ae6b6d699/pyproject.toml#L54

to say `3.7`, and then `pip install -e .` (local installation). It didn't work because there's no compatible Awkward Array version for Python 3.8, and that would require a custom installation, so it's turning into a rabbit hole.

I think you don't need the above for your analysis, right? It's just changing the title of the TBranches that Uproot writes. ROOT still recognizes the data with slightly different TBranch titles. (I experimented on that yesterday.)

umute97:
> It didn't work because there's no compatible Awkward Array version for Python 3.8, and that would require a custom installation, so it's turning into a rabbit hole.
Bummer ¯\_(ツ)_/¯
> I think you don't need the above for your analysis, right? It's just changing the title of the TBranches that Uproot writes. ROOT still recognizes the data with slightly different TBranch titles. (I experimented on that yesterday.)
Yes, don't worry! While experimenting I just realized that I'm still running an old version of `uproot` and was wondering why and update it if I can. Now I know. Thanks for your continued effort for providing this awesome interface!

PR #1218, opened 2024-05-14
fix: add support for new RNTuple switches

ariostas:
This PR fixes reading RNTuple switch data, which is now 96 bits as opposed to 64 bits. These are used to encode `std::variant` types. I re-enabled a test for them, which I didn't realize was being skipped.

Also, as a minor thing, I disabled another RNTuple test because the test file takes a relatively long file to download, and sometimes the server seems to get stuck.

### DIFF
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -480,9 +480,15 @@ in file {self.file.file_path}"""
         pagelist = linklist[ncol]
         dtype_byte = self.column_records[ncol].type
         dtype_str = uproot.const.rntuple_col_num_to_dtype_dict[dtype_byte]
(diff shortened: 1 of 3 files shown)


ariostas:
I'm going to make it a bit cleaner by using structured `numpy.dtype` for switches. I'll mark it for review when it's ready.

ariostas:
Thanks @jpivarski, it's definitely a lot nicer with a structured numpy dtype than reading a switch as 3 int32. Now it's ready for review.

jpivarski (approved):
Oh, yeah: now that I see the context, the structured dtype is much better. I thought we were talking about three equal-sized integers, which can just as easily be a reshaped array, but it's one 64-bit integer and one 32-bit integer. Reinterpreting as a structured array does the same job with no intermediate arrays, so it's both cleaner to read and more efficient.

Go ahead and merge if you have no additional changes to make!

PR #1222, opened 2024-05-21
ci: Add GitHub artifact attestations to package distribution

matthewfeickert:
* Add generation of GitHub artifact attestations to built sdist and wheel before upload. c.f.:
   - https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/
   - https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds


jpivarski (approved):
Looks good; thanks! I'll merge this now, because it looks like a straightforward addition (no pending updates).

PR #1223, opened 2024-05-23
feat: add support for more RNTuple types

ariostas:
This PR adds support for a couple of RNTuple types that are currently not supported, namely `std::atomic<T>` and `std::bitset`. I'll make a PR to `scikit-hep-testdata` to add those, and then set up some tests here.

### DIFF
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -351,11 +351,13 @@ in file {self.file.file_path}"""
             # n.b. the split may happen in column
             return self.col_form(this_id)
         elif structural_role == uproot.const.rntuple_role_leaf:
-            # std::array it only has one child
             if this_id in self._related_ids:
+                # std::array has only one subfield
                 child_id = self._related_ids[this_id][0]
-
-            inner = self.field_form(child_id, seen)
+                inner = self.field_form(child_id, seen)
+            else:
+                # std::bitset has no subfields, so we use it directly
+                inner = self.col_form(this_id)
             keyname = f"RegularForm-{this_id}"
             return ak.forms.RegularForm(inner, this_record.repetition, form_key=keyname)
         elif structural_role == uproot.const.rntuple_role_vector:
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -338,6 +338,15 @@ in file {self.file.file_path}"""
             structural_role == uproot.const.rntuple_role_leaf
             and this_record.repetition == 0
         ):
+            # deal with std::atomic
+            # they have no associated column, but exactly one subfield containing the underlying data
+            tmp_id = self._alias_columns_dict.get(this_id, this_id)
+            if (
+                tmp_id not in self._column_records_dict
+                and len(self._related_ids[tmp_id]) == 1
+            ):
+                this_id = self._related_ids[tmp_id][0]
+                seen.add(this_id)
             # base case of recursion
             # n.b. the split may happen in column
             return self.col_form(this_id)
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -397,7 +397,8 @@ in file {self.file.file_path}"""
             newids = []
             if this_id in self._related_ids:
                 newids = self._related_ids[this_id]
-            recordlist = [self.field_form(i, seen) for i in newids]
+            recordlist = [ak.forms.RecordForm([], [], form_key="inv_var")]
+            recordlist.extend([self.field_form(i, seen) for i in newids])
             return ak.forms.UnionForm("i8", "i64", recordlist, form_key=keyname)
         else:
             # everything should recurse above this branch
@@ -585,7 +586,7 @@ in file {self.file.file_path}"""
 
 # Supporting function and classes
 def _split_switch_bits(content):
-    tags = content["tag"].astype(numpy.dtype("int8")) - 1
+    tags = content["tag"].astype(numpy.dtype("int8"))
     kindex = content["index"]
     return kindex, tags
 
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -549,7 +549,9 @@ in file {self.file.file_path}"""
             [c.num_entries for c in clusters[start_cluster_idx:stop_cluster_idx]]
         )
 
-        form = self.to_akform().select_columns(filter_names)
+        form = self.to_akform().select_columns(
+            filter_names, prune_unions_and_records=False
+        )
         # only read columns mentioned in the awkward form
         target_cols = []
         container_dict = {}
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -579,10 +579,10 @@ in file {self.file.file_path}"""
                         tags = numpy.delete(tags, invalid)
                         invalid -= numpy.arange(len(invalid))
                         optional_index = numpy.insert(
-                            numpy.arange(len(kindex)), invalid, -1
+                            numpy.arange(len(kindex), dtype=numpy.int64), invalid, -1
                         )
                     else:
-                        optional_index = numpy.arange(len(kindex))
+                        optional_index = numpy.arange(len(kindex), dtype=numpy.int64)
                     container_dict[f"{key}-index"] = optional_index
                     container_dict[f"{key}-union-index"] = kindex
                     container_dict[f"{key}-union-tags"] = tags
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -30,7 +30,7 @@ _rntuple_locator_format = struct.Struct("<iQ")
 _rntuple_cluster_summary_format = struct.Struct("<QQ")
 _rntuple_checksum_format = struct.Struct("<Q")
 _rntuple_envlink_size_format = struct.Struct("<Q")
-_rntuple_page_num_elements_format = struct.Struct("<I")
+_rntuple_page_num_elements_format = struct.Struct("<i")
 _rntuple_column_group_id_format = struct.Struct("<I")
 _rntuple_first_ele_index_format = struct.Struct("<I")
 
@@ -605,9 +605,9 @@ def _recursive_find(form, res):
 class PageDescription:
     def read(self, chunk, cursor, context):
         out = MetaData(type(self).__name__)
-        out.num_elements = cursor.field(
-            chunk, _rntuple_page_num_elements_format, context
-        )
+        num_elements = cursor.field(chunk, _rntuple_page_num_elements_format, context)
+        out.has_checksum = num_elements < 0
+        out.num_elements = abs(num_elements)
         out.locator = LocatorReader().read(chunk, cursor, context)
         return out
 
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -397,9 +397,11 @@ in file {self.file.file_path}"""
             newids = []
             if this_id in self._related_ids:
                 newids = self._related_ids[this_id]
-            recordlist = [ak.forms.RecordForm([], [], form_key="inv_var")]
-            recordlist.extend([self.field_form(i, seen) for i in newids])
-            return ak.forms.UnionForm("i8", "i64", recordlist, form_key=keyname)
+            recordlist = [self.field_form(i, seen) for i in newids]
+            inner = ak.forms.UnionForm(
+                "i8", "i64", recordlist, form_key=keyname + "-union"
+            )
+            return ak.forms.IndexedOptionForm("i64", inner, form_key=keyname)
         else:
             # everything should recurse above this branch
             raise AssertionError("this should be unreachable")
@@ -558,7 +560,7 @@ in file {self.file.file_path}"""
         container_dict = {}
         _recursive_find(form, target_cols)
         for key in target_cols:
-            if "column" in key:
+            if "column" in key and "union" not in key:
                 key_nr = int(key.split("-")[1])
                 dtype_byte = self.column_records[key_nr].type
                 content = self.read_col_pages(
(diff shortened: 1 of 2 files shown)


ariostas:
There's two more fixes that I'll add to this PR. Currently, if you have an `std::variant` where one of the types is an empty struct, something goes wrong and it ends up discarding that part of the variant. The other one, is that if you try to read a variant in an invalid state it just doesn't work.

ariostas:
> There's two more fixes that I'll add to this PR. Currently, if you have an std::variant where one of the types is an empty struct, something goes wrong and it ends up discarding that part of the variant. The other one, is that if you try to read a variant in an invalid state it just doesn't work.

To follow up, it's not straightforward to fix this. For empty structs, one can make some progress using the `prune_unions_and_records=False` argument of `select_columns`, but `Awkward` can't really deal with empty structs, so there's no way to deal with this unless we implement it in a hacky way that turns empty structs into `None`. For invalid `std::variant`s, Jim suggested using the `UnionArray.simplified` constructor, but that has some issues including it removing `form_key` fields and merging forms that we don't want to merge.

So given that these are probably uncommon edge cases, I'd say it's probably fine to not support them for now and if in the future there is a real need for them then we will figure out how to implement them.

jpivarski:
Currently not covered:

* records with no fields (allowed in Awkward, just needs RNTuple → Awkward translation)
* invalid variants: ???

ariostas:
> Currently not covered:
> 
>   * records with no fields (allowed in Awkward, just needs RNTuple → Awkward translation)
>   * invalid variants: ???

I added a test file for these cases in https://github.com/scikit-hep/scikit-hep-testdata/pull/155

ariostas:
@jpivarski Here's a follow up on the above discussion.

For empty structs, ROOT has no issues. Following our discussion on Slack from a while ago, all I had to do to fix them on Uproot was to use `prune_unions_and_records=False` for the `select_columns` function, since, unlike Parquet, these are fine and don't need to be pruned. So now there's no issue with these.

For invalid variants, both ROOT and Uproot end up having very similar behavior. Both are able to read the data with no issues, but in both cases one needs to be careful using it.

In ROOT:
- Checking `variant_field->index()` returns `(unsigned long) 255`.
- Doing something like `auto v = *variant_field;` throws an exception.
- Weirdly enough, `variant_field->valueless_by_exception()` returns `(bool) false`.

In Uproot:
- The tag index is set to `(int8) -1`.
- Doing something like `var_array[index_of_inv_var]` raises `ak._errors.index_error`.

So given that the behavior is similar and in either case one needs to be careful, I think it's fine to have this level of support for invalid variants.

I'll add tests for all the new types and cases supported, and then this PR is ready for review.

jpivarski (commented):
Sounds good!

I don't see any tests associated with these changes. Is there any test that would be sensitive to them? You have files with these features, right?

> In Uproot:
> 
> * The tag index is set to `(int8) -1`.

That's invalid, and Awkward Array may complain about it at any time. The only reason it doesn't complain right away is because it's an $\mathcal{O}(n)$ check that it only discovers at runtime or if an array is constructed with `check_valid=True`.

There's another way to deal with this: the invalid cases could be turned into missing values. If you construct the UnionArray inside of `ByteMaskedArray.simplified` or `IndexedOptionArray.simplified`, this function will do the work of adjusting the option-type and union-type into a canonical form. The UnionArray can temporarily have `-1` tag indexes and this process of adjustment would eliminate them (because they're unreachable, and the process doesn't need to preserve unreachable elements).

The only problem that I see with this is that the type (including an option-type) can't be determined from data values. If you don't know whether there will be any invalid variants until you look at its column data, that's too late. One reason is that a Dask process will need to make decisions based on types, but it doesn't look at the data.

If there's no way of knowing whether a variant in RNTuple will have `-1` in its tags or not, then we are forced to consider _all_ variants to potentially have missing values, even if most of them don't.

There's still time to ask the ROOT team to add metadata to the type schema indicating whether it's possible for a particular variant to have `-1` in its tags or not. I suspect that this only applies to class types in C++ and not to value types.

ariostas:
I'll add tests once https://github.com/scikit-hep/scikit-hep-testdata/pull/155 gets merged and released.

> That's invalid, and Awkward Array may complain about it at any time.

Yeah, the thing is that type tags are 1-indexed, with 0 indicating an invalid variant. Since there's already `\mathcal{O}(n)` work being done to convert them to the 0-indexed tags that Awkward expects, then I guess we might as well check if there are any invalid tags. Probably a better alternative would be to avoid the `\mathcal{O}(n)` work and include the invalid case by having Awkward treat 0 as missing values. Would that be doable with current Awkward functionality?

As a side note, is there a reason why Awkward uses int8 instead of uint8 for tags? I think RNTuple is supposed to support variants with up to 255 types (+ invalid). I think no one in their right mind would use even close to that many types, but it could technically be an issue.

jpivarski:
(Oops; pressed "enter" too early.)

The reason that `ak.contents.UnionArray` constructors avoid anything other than $\mathcal{O}(1)$ is because they are called frequently in the process of manipulating data. Unwrapping and re-wrapping layouts is common, and if we start in a valid state, do nothing, and finish in a valid state, checking validity is wasted computation time. A data-manipulation process (in the Awkward codebase) should only do necessary work, assuming that the input is valid, and only check for validity that it _relies_ on in on-the-fly tests. (The kernels that transform Indexes check for constraints that they need, such as list `stops[i] >= starts[i]`, during the kernel.)

None of this applies to the original construction phase in Uproot, so go ahead and do $\mathcal{O}(n)$ or $\mathcal{O}(n log n)$ work, as needed. In fact, considering that Awkward _assumes_ validity, it's worse to knowingly produce something invalid, since that would show up as an error downstream or worse—incorrect results.

jpivarski:
> As a side note, is there a reason why Awkward uses int8 instead of uint8 for tags?

There isn't a good reason for that. In retrospect, we should have used uint8, since we don't use negative values for anything meaningful and if we're not careful in checking for validity on the fly (in the Awkward codebase), negative values might end up invoking Python's wrap-around indexing (counting negative indexes from the end of a list of `contents`).

But it didn't seem like a relevant point because 127 is already _way_ too many union tags. If an array has a hundred distinct types, Awkward manipulations are going to be spending a lot of time messing with metadata (in Python), rather than number-crunching on the array buffers (in C or CUDA).

Unions are tricky to work with in general—there are a lot of surprising corner-cases that had to be left as `NotImplementedError`. At most, I can imagine someone doing data analysis with a handful of different types, such as different versions of a record with more or fewer fields. Some analysts have attempted to combine "electron" and "muon" types into a "lepton" array as a union, but ran into some issues downstream. In fact, I've found it useful to go in the opposite direction: instead of putting disparate data into one array, I've found it useful to split arrays by particle type as much as possible, even to make separate arrays for positive and negative charges of a particle type, since it's less confusing to do [ak.cartesian](https://awkward-array.org/doc/main/reference/generated/ak.cartesian.html) on non-overlapping sets than [ak.combinations](https://awkward-array.org/doc/main/reference/generated/ak.combinations.html) on a single set and then filter the unwanted combinations (like wrong-sign pairs). If I were doing an analysis with tagged and untagged particles (like b-tags), I would also start by partitioning the data into separate arrays. So, in general: un-unioning is a better practice than unioning.

Considering that, we still want to have a union type because input data comes in union form. Converting union-type data into simpler data is a likely pre-processing step, just as padding and truncating irregular lists is a typical pre-processing step.

If an RNTuple gives us more than 127 union choices, you can raise `NotImplementedError` for now. If this is an issue in the future, we can make some minimal modifications to allow Awkward Array to interpret tags as uint8, even if they're technically int8. Since their meaning is categorical, all we need are distinct bit-patterns. We'll just need to be careful to reinterpret-cast as uint8 before applying it as an index of the `contents` list.

jpivarski:
Sorry; I missed the PR in scikit-hep-testdata. I put it into a release now: [0.4.46](https://github.com/scikit-hep/scikit-hep-testdata/releases/tag/v0.4.46).

ariostas:
Sounds good, thank you! I'll go ahead and implement this and I'll let you know once it's ready.

ariostas:
@jpivarski how do you feel about the changes in ba422a5 to support invalid variants? The advantages over what we discussed over Slack are that it is very straightforward and doesn't require any $\mathcal{O}(n)$ operations. The invalid variants would simply show up as empty records since the type and index of those entries are always zero.

jpivarski:
Instead of invalid variants appearing as `None`, they'd be appearing as empty records, `{}`?

Well, that would reduce its usefulness. There are some operations that can't be performed on records, even if they have no fields, such as ufuncs and reducers. For example, consider these two unions:

```python
>>> one = ak.Array([[1, 4, 9], None, 25])
>>> two = ak.Array([[1, 4, 9], {}, 25])
```

The first one has a missing value and the second has an empty record in index `1`.

```python
>>> one.type.show()
3 * union[
    option[var * int64],
    ?int64
]
>>> two.type.show()
3 * union[
    var * int64,
    {
        
    },
    int64
]
```

The first consists entirely of numbers, lists, and missing values, which can be passed to numerical functions like `np.sqrt`:

```python
>>> np.sqrt(one)
<Array [[1, 2, 3], None, 5] type='3 * union[option[var * float64], ?float64]'>
```

But the second also has a record type, and records are forbidden in ufuncs unless there's an explicit behavior defined for it:

```python
>>> np.sqrt(two)
```

```
  File "/home/jpivarski/irishep/awkward/src/awkward/highlevel.py", line 1511, in __array_ufunc__
    return ak._connect.numpy.array_ufunc(ufunc, method, inputs, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_connect/numpy.py", line 466, in array_ufunc
    out = ak._broadcasting.broadcast_and_apply(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1108, in broadcast_and_apply
    out = apply_step(
          ^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1086, in apply_step
    return continuation()
           ^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1055, in continuation
    return broadcast_any_list()
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 623, in broadcast_any_list
    outcontent = apply_step(
                 ^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1086, in apply_step
    return continuation()
           ^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1044, in continuation
    return broadcast_any_union()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 952, in broadcast_any_union
    apply_step(
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1086, in apply_step
    return continuation()
           ^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1040, in continuation
    return broadcast_any_indexed()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 990, in broadcast_any_indexed
    return apply_step(
           ^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1086, in apply_step
    return continuation()
           ^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 1059, in continuation
    return broadcast_any_record()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_broadcasting.py", line 468, in broadcast_any_record
    raise ValueError(f"cannot broadcast records{in_function(options)}")
ValueError: cannot broadcast records in sqrt

This error occurred while calling

    numpy.sqrt.__call__(
        <Array [[1, 4, 9], {}, 25] type='3 * union[var * int64, {}, int64]'>
    )
```

Also, RNTuple's use of invalid variants is meant to cover cases like `nullptr` in C++, right? If so, the Python/Awkward equivalent of that _should_ be `None`.

ariostas:
Thank you @jpivarski, that makes sense. I made the changes so that now invalid variants end up as `None`. I think this PR is ready for review now.

ariostas:
Hmm for some reason it sometimes doesn't work on Windows. I'll look into it.

src/uproot/models/RNTuple.py
```diff
@@ -538,13 +552,15 @@ def arrays(
...
+        form = self.to_akform().select_columns(
+            filter_names, prune_unions_and_records=False
+        )
         # only read columns mentioned in the awkward form
         target_cols = []
         container_dict = {}
         _recursive_find(form, target_cols)
         for key in target_cols:
-            if "column" in key:
+            if "column" in key and "union" not in key:
```
jpivarski, line 563: Does `key` have the word `"union"` in it by construction? Is it because the form analyzed by

https://github.com/scikit-hep/uproot5/blob/25fbb3a85aa87b6d817c40a362ef581c693c5cea/src/uproot/models/RNTuple.py#L608-L617

is a Form that you made and deliberately added the string `"union"`?

I'm asking because this is not something that can be generically expected of given data, only if you built it that way in this code.

jpivarski (approved):
This is great! It's fine that there are $\mathcal{O}(n)$ _vectorized_ operations involved in constructing the array—similar things are needed for other formats. Just as long as there aren't any $\mathcal{O}(n)$ Python loops.

I don't see any test failures with Windows. Is that something that you encountered outside of CI?

Other than that, I think this is ready to merge. Thanks!

src/uproot/models/RNTuple.py
ariostas: Since for variants now it initially uses nested `IndexedOptionForm(UnionForm(...))`, I added a `"-union"` suffix to differentiate the key of the inner form. Do you think this could cause issues?

https://github.com/scikit-hep/uproot5/blob/25fbb3a85aa87b6d817c40a362ef581c693c5cea/src/uproot/models/RNTuple.py#L401-L404

ariostas:
> I don't see any test failures with Windows. Is that something that you encountered outside of CI?

The failures on Windows were for a98dc19. It turned out that NumPy 1.X on Windows defaults to `dtype=int32` when using `np.arange`. So the dtype was explicitly added in 25fbb3a.

PR #1225, opened 2024-05-23
feat: changes for dask-awkward one-pass optimize

martindurant:
This makes the basic uproot tests in https://github.com/dask-contrib/dask-awkward/pull/491 pass

cc @lgray


src/uproot/_dask.py
```diff
@@ -889,7 +888,9 @@ def keys_for_buffer_keys(self, buffer_keys: frozenset[str]) -> frozenset[str]:
         keys: set[str] = set()
         for buffer_key in buffer_keys:
             # Identify form key
-            form_key, attribute = buffer_key.rsplit("-", maxsplit=1)
+            form_key, attribute = buffer_key.replace("@.", "<root>.").rsplit(
+                "-", maxsplit=1
```
martindurant, line 892: Would this fail for any of the styles of buffer name? We could use `[0]` or `*attribute` to guard against this (we don't actually use the attribute).

In fact, this whole loop could be a single comprehension.

src/uproot/_dask.py
```diff
@@ -1026,84 +1031,19 @@ def read_tree(
...
-                ),
-                "form_info": self.form_mapping_info,
-            },
-        )
+    def project(self, columns) -> T:
+        from dask_awkward.lib.utils import _buf_to_col
 
-    def project(self: T, *, report: TypeTracerReport, state: dict) -> T:
-        keys = self.necessary_columns(report=report, state=state)
+        keys = [_buf_to_col(c).replace(".", "_") for c in columns]
```
martindurant, line 1041: @lgray : this is the simplistic way to unmap the column names. This line for the simple case, the if-block for nanoAOD, adding in implicit columns like nJet. Then, this doesn't actually make use of the schema and keys_for_buffer_keys

src/uproot/_dask.py
jpivarski: It's unclear what characters are _not_ allowed in a TBranch name, which gets passed over to the form_keys. It would be hard to use a `:` in a TBranch name, since that gets interpreted by ROOT (in some constructors, not all). On the other hand, `.` is very common and can't be a delimiter.

I tried to see what would break it:

```c++
root [0] TTree* t = new TTree("name", "title")
(TTree *) 0x6360d13c1720
root [1] (new TBranch(t, "something", nullptr, "leaf"))->GetName()
(const char *) "something"
root [2] (new TBranch(t, "some.thing", nullptr, "leaf"))->GetName()
(const char *) "some.thing"
root [3] (new TBranch(t, "some:thing", nullptr, "leaf"))->GetName()
(const char *) "some:thing"
root [4] (new TBranch(t, "some/thing", nullptr, "leaf"))->GetName()
(const char *) "some/thing"
root [5] (new TBranch(t, "some\"thing", nullptr, "leaf"))->GetName()
(const char *) "some"thing"
root [6] (new TBranch(t, "some`thing", nullptr, "leaf"))->GetName()
(const char *) "some`thing"
root [7] (new TBranch(t, "some!thing", nullptr, "leaf"))->GetName()
(const char *) "some!thing"
root [8] (new TBranch(t, "some@thing", nullptr, "leaf"))->GetName()
(const char *) "some@thing"
root [9] (new TBranch(t, "some#thing", nullptr, "leaf"))->GetName()
(const char *) "some#thing"
root [10] (new TBranch(t, "some+thing", nullptr, "leaf"))->GetName()
(const char *) "some+thing"
root [11] (new TBranch(t, "some(thing", nullptr, "leaf"))->GetName()
(const char *) "some(thing"
root [12] (new TBranch(t, "some{thing", nullptr, "leaf"))->GetName()
(const char *) "some{thing"
root [13] (new TBranch(t, "some[thing", nullptr, "leaf"))->GetName()
(const char *) "some[thing"
root [14] (new TBranch(t, "some\\thing", nullptr, "leaf"))->GetName()
(const char *) "some\thing"
root [15] (new TBranch(t, "some|thing", nullptr, "leaf"))->GetName()
(const char *) "some|thing"
root [16] (new TBranch(t, "some;thing", nullptr, "leaf"))->GetName()
(const char *) "some;thing"
root [17] (new TBranch(t, "some\0thing", nullptr, "leaf"))->GetName()
(const char *) "some"
```

Here's something we can count on: a ROOT TBranch name is not going to have a null byte (Python `"\x00"`) in the _middle_ of the string, whereas this is legal and would be properly carried around in Python. It seems pretty drastic, though.

src/uproot/_dask.py
jpivarski: (Reminder to @lgray to take a look at this.)

jpivarski (approved):

Issue #1229, opened 2024-06-10
`parse_typename` fails on const pointer members


veprbl:
```shell
wget -O uproot-issue-XXX.root https://github.com/scikit-hep/scikit-hep-testdata/raw/7387822903e0c0de02cee4b98dbb9677f7bfba2a/src/skhep_testdata/data/uproot-issue-1229.root
```

```python
>>> import uproot
>>> uproot.__version__
'5.3.7'
>>> uproot.open("uproot-issue-XXX.root")["tree"]["branch/pointer"].typename
'TFooMember*'
>>> uproot.open("uproot-issue-XXX.root")["tree"]["branch/const_pointer"].typename
Traceback (most recent call last):
  File "/nix/store/bz0s89q09d1jxb96bcpjad38sqw6nimk-python3.11-uproot-5.3.7/lib/python3.11/site-packages/uproot/interpretation/identify.py", line 385, in interpretation_of
    from_dtype = _leaf_to_dtype(leaf, getdims=False).newbyteorder(">")
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/bz0s89q09d1jxb96bcpjad38sqw6nimk-python3.11-uproot-5.3.7/lib/python3.11/site-packages/uproot/interpretation/identify.py", line 103, in _leaf_to_dtype
    return numpy.dtype((_ftype_to_dtype(leaf.member("fType")), dims))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/bz0s89q09d1jxb96bcpjad38sqw6nimk-python3.11-uproot-5.3.7/lib/python3.11/site-packages/uproot/interpretation/identify.py", line 61, in _ftype_to_dtype
    raise NotNumerical()
uproot.interpretation.identify.NotNumerical

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/nix/store/bz0s89q09d1jxb96bcpjad38sqw6nimk-python3.11-uproot-5.3.7/lib/python3.11/site-packages/uproot/behaviors/TBranch.py", line 1942, in typename
    if self.interpretation is None:
       ^^^^^^^^^^^^^^^^^^^
[...]

PR #1231, opened 2024-06-12
fix: support `const` in typename

veprbl:
Resolves: #1229

### DIFF
--- a/src/uproot/interpretation/identify.py
+++ b/src/uproot/interpretation/identify.py
@@ -921,19 +921,6 @@ def _parse_node(tokens, i, typename, file, quote, header, inner_header):
             _parse_maybe_quote(f"uproot.containers.AsString({header})", quote),
         )
 
-    elif (
-        has2
-        and tokens[i].group(0) == "const"
-        and (
-            tokens[i + 1].group(0) == "string"
-            or _simplify_token(tokens[i + 1]) == "std::string"
-        )
-    ):
-        return (
-            i + 2,
-            _parse_maybe_quote(f"uproot.containers.AsString({header})", quote),
-        )
-
     elif tokens[i].group(0) == "TString":
         return (
             i + 1,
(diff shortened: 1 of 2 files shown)


veprbl:
This is ready for review. I could not resolve one of the issues with a test (see b2928b3891a3d8f0715f19a0b1f26c77b5e088ee), and instead found a reasonable alternative way to do the same test.

jpivarski:
Thanks for getting back to this! The strategy of checking for `"const"` at the very beginning and ignoring it thereafter sounds good to me, although when users ask for the `typename`, it will lack the `const` keywords, which they might be expecting. But I guess that's fine: the `const` keyword doesn't mean much for _serialized_ data.

It should not be necessary to apply the workaround described in b2928b3891a3d8f0715f19a0b1f26c77b5e088ee. I checked out this branch and

```python
>>> parse_typename("TH1I*") == AsPointer(Model_TH1I)
True
>>> parse_typename("const TH1I*") == AsPointer(Model_TH1I)
True
```

Moreover, this is not how the tests are failing—they're failing because they can't access test files through HTTP. It seems like some server is down. I'm [running the tests again in main](https://github.com/scikit-hep/uproot5/actions/runs/12822586124), and we'll see if there's an error there.

I should also look through the `parse_typename` for other places where we explicitly check for `"const"` because your change would have made them dead code (you can't get `"const"` later if you've already left at the beginning of the function). Pushing that modification would trigger the tests again. If it was a momentary HTTP server failure, they would pass.

jpivarski:
The tests in `main` are going to fail. It's because example.com has changed how it works, and we had been using it to test what (we thought) should be HTTP 206 errors, but it's now returning 200 (no error).

#1362 is an adjustment for that, and if it works, it will merge into main and then can be merged into this PR branch, which will fix the tests.

jpivarski:
In the two now-dead code blocks that I removed, I noticed that our old policy had always been to ignore the `const` keyword in type names, so that aspect is not changing.

jpivarski (approved):
I wrote a lot of inline comments about the test failures and other updates. But when all test pass (by merging #1362 into main and main into this PR), I think this can be merged. Let me know, @ianna, if you disagree!

ianna (approved):
Thanks! Looks great. The tests should pass now.

veprbl:
I appreciate the fixups to the existing code. My recollection was that `const` types would outright not parse without my change, so I did not think of looking for those cases.

jpivarski:
The only two `const` cases that had been handled before were strings. Your
solution takes the `const` off of any type.

Issue #1220, opened 2024-05-16
RNtuple `__len__` should return the number of keys, and have a `num_entries` property for the number of entries


jpivarski:
Although it may be counterintuitive, `uproot.TTree` and `uproot.RNTuple` should return the number of branches/top-level columns from `__len__`, rather than the number of entries. This is because they satisfy the [Mapping](https://docs.python.org/3/glossary.html#term-mapping) protocol: `__len__` is the length of the output of `keys`, `values`, and `items`.

For the number of entries, `uproot.RNTuple` should follow `uproot.TTree` in having a [num_entries](https://uproot.readthedocs.io/en/latest/uproot.behaviors.TTree.TTree.html#num-entries) property. Users will need this information to know how to set `entry_start` and `entry_stop`.

---------

Maybe `uproot.RNTuple` needs an [iterate](https://uproot.readthedocs.io/en/latest/uproot.behaviors.TTree.TTree.html#iterate) method as well. For TTree, this is _functionally_ equivalent to

```python
def iterate(self, *args, step_size=<default>, **kwargs):
    step_size = regularize_step_size(step_size)
    for start in range(0, self.num_entries, step_size):
        yield self.arrays(*args, entry_start=start, entry_stop=start + step_size, **kwargs)
```

but TTree does a little more and the partially read TBaskets that were trimmed to yield one array are reused if they're part of the next array (taking advantage of the fact that `iterate` is sequential: we know what's coming next). Here's where that happens:

https://github.com/scikit-hep/uproot5/blob/13087b0f6d179bb8982249d8fda8cda5f3b8e44c/src/uproot/behaviors/TBranch.py#L1028-L1114

I don't know if the RNTuple code is organized in such a way that it would be easy to do this. (Not that I would call the TTree implementation "easy": this feature is obfuscating the code quite a bit.) I'm sympathetic to the argument that this isn't worth doing, since parallel chunked-access methods are more important than sequential ones.

PR #1250, opened 2024-07-15
chore: miscellaneous RNTuple improvements

ariostas:
This PR addresses the issues in #1220. I fixed the `__len__` method so that it returns the number of top-level columns instead of the number of entries, which is now the `num_entries` property.

I'll work on implementing an `iterate` method and update this description when it's ready.


src/uproot/const.py
```diff
@@ -212,7 +208,29 @@
...
+rntuple_field_flag_checksum = 0x04
+
+rntuple_col_flag_deferred = 0x08
+
+rntuple_extra_type_identifier_root = 0x00
+
+rntuple_user_metadata_type_int = 0x01
+rntuple_user_metadata_type_bool = 0x02
+rntuple_user_metadata_type_double = 0x03
+rntuple_user_metadata_type_string = 0x04
```
ariostas, line 236: Maybe these should be enums

ariostas:
I made some more updates to keep up to date with the spec. I'll wrap up this PR next week and possibly just postpone the implementation of `iterate` to another PR.

Closes #1288, closes #1302, closes #1303

ariostas:
@jpivarski this PR is ready for review now. Here are some highlights/comments:

- I fixed the `__len__` vs `num_entries` issue from #1220.
- I implemented a simple `iterate` method. It supports step sizes in terms of memory, but the current estimation method might not be optimal.
- I added an `RNTupleField` class, so that now you can do things like `obj["nMuon"].array()` and `obj["AntiKt4TruthDressedWZJetsAux:"]["pt"].array()`, but I haven't implemented path support yet for things like `obj["AntiKt4TruthDressedWZJetsAux:/pt"].array()`.
- I updated some things to match the spec, including fixes to #1288 #1302 #1303.
- I made some other minor improvements.

jpivarski (approved):
Very nice—I like the grouped enums! As far as I can see, these are maintainability updates, not changes in how the code functions. (Okay, I see that the PR title contains "chore:" rather than "feat:" or "fix:".)

Again, I'm forwarding you (email) the information about the RC3 format change, which is likely to be the last one ever.

Feel free to merge this PR!

PR #1254, opened 2024-07-23
fix: set _max_workers in ThreadPoolExecutor with non-None initializer

NJManganelli:
Looks like this was missing when the function got rebuilt. A test added with None and 1 max_workers arguments in the style of PR#1120, let me know if names should be changed


jpivarski (approved):
Yeah, that's right! If `max_workers` is given, you want the ThreadPoolExecutor to actually use it.

With the PR title fixed, this should pass all cuts, and I'll merge it as soon as it does.

PR #1256, opened 2024-07-24
feat: implement TGraph-writing

Pepesob:
Implemented function `as_TGraph()` that gives to user possibilty to save DataFrame (or any dict like object) as TGraph. 
Example of usage:
```python
import uproot
f = uproot.recreate("example.root")
x = np.array([-50,0,1, 2, 3, 4, 5, 17, 100,110], dtype='float64')
y = np.array([-50,0,1, 2, 3, 4, 5, 10, 100,120], dtype='float64')
d = {'x': x, 'y': y}
df = pd.DataFrame(d)
f["mytgraph"] = uproot.as_TGraph(df)
f.close()
```

List of arguments:
```

- df (DataFrame or and dict like object): DataFrame object with column names as follows:
     x (float): x values of TGraph.
     y (float): y values of TGraph.
     x_errors (float or left unspecified): Symethrical error values for corresponding x value
     y_errors (float or left unspecified): Symethrical error values for corresponding y value
     x_errors_low (float or left unspecified): Asymmetrical lower error values for corresponding x value
     x_errors_high (float or left unspecified): Asymmetrical upper error values for corresponding x value
     y_errors_low (float or left unspecified): Asymmetrical lower error values for corresponding y value
     y_errors_high (float or left unspecified): Asymmetrical upper error values for corresponding y value
     (other column names will be ignored!)
- title (str): Title of the histogram.
- xAxisLabel (str): Label of the X axis.
- yAxisLabel (str): Label of the Y axis.
- minY (None or float): Minimum value on the Y axis to be shown, if set to None then minY=min(y)
- maxY (None or float): Maximum value on the Y axis to be shown, if set to None then maxY=max(y)
- lineColor (int): Line color. (https://root.cern.ch/doc/master/classTAttLine.html)
- lineStyle (int): Line style.
- lineWidth (int): Line width.
- markerColor (int): Marker color. (https://root.cern.ch/doc/master/classTAttMarker.html)
- markerStyle (int): Marker style.
- markerSize (float): Marker size.

```
How does it work:
```
  Converts arguments into TGraph, TGraphErrors or TGraphAsymmErros based on the given arguments.
  When all errors are unspecified, detected object is TGraph.
  When x_errors, y_errors are specified, detected object is TGraphErrors.
  When x_errors_low, x_errors_high, y_errors_low, y_errors_high are specified, detected object is TGraphAsymmErrors.
  Note that both x_errors, y_errors need to be specified or set to None.
  The same rule applies to x_errors_low, x_errors_high, y_errors_low, y_errors_high.
  Also can't specify x_errors, y_errors and x_errors_low, x_errors_high, y_errors_low, y_errors_high at the same time.
  All rules are designed to remove any ambiguity.
```


tests/test_1128_TGraph_writing.py
```diff
@@ -0,0 +1,57 @@
...
+
+        assert tgraph.GetTitle() == f"{title};{xLabel};{yLabel}"
+        assert tgraph.GetXaxis().GetTitle() == xLabel
+        assert tgraph.GetYaxis().GetTitle() == yLabel
+
+        for i in range(len(x)):
+            xAxis = tgraph.GetX()
+            yAxis = tgraph.GetY()
+            assert abs(xAxis[i] - x[i]) < EPS
+            assert abs(yAxis[i] - y[i]) < EPS
```
jpivarski, line 57:
```suggestion
            assert xAxis[i] == pytest.approx(x[i])
            assert yAxis[i] == pytest.approx(y[i])
```

tests/test_1128_TGraph_writing.py
```diff
@@ -0,0 +1,57 @@
+import uproot
+import os
+import pandas as pd
+import pytest
+import ROOT
+import numpy as np
+
+EPS = 1e-6
```
jpivarski, line 8:
```suggestion
```

src/uproot/writing/interpret.py
```diff
@@ -0,0 +1,306 @@
...
+        if df.get("y_errors_low", None) is not None
+        else None
+    )
+    y_errors_high = (
+        numpy.array(df["y_errors_high"])
+        if df.get("y_errors_high", None) is not None
+        else None
+    )
+
+    return _to_TGraph(
```
jpivarski, line 284: I'm not sure why this function has been split into a public part and a private part, but it's fine.

src/uproot/writing/interpret.py
```diff
@@ -0,0 +1,306 @@
...
+
+    asym_errors = [x_errors_low, x_errors_high, y_errors_low, y_errors_high]
+    asym_errors_bool = [err is not None for err in asym_errors]
+
+    tgraph_type = "TGraph"
+
+    # Detecting which type of TGraph to chose
+    if any(sym_errors_bool):
+        if not all(sym_errors_bool):
+            raise ValueError("Have to specify both x_errors and y_errors")
```
jpivarski, line 78: Error messages are lower case with no end-punctuation (period), but I'm also rewriting it to tell the user which high-level function has failed. If this is called by a library that the user didn't write, they might not have called `uproot.to_TGraph` themselves, and they might not even know that it exists. Mentioning it here helps them to debug.

```suggestion
            raise ValueError("uproot.to_TGraph requires both x_errors and y_errors")
```

src/uproot/writing/interpret.py
```diff
@@ -0,0 +1,306 @@
...
+        if any(asym_errors_bool):
+            raise ValueError(
+                "Can's specify both symetrical errors and asymetrical errors"
```
jpivarski, line 81:
```suggestion
                "uproot.to_TGraph can accept symmetrical errors OR asymmetrical errors, but not both"
```

src/uproot/writing/interpret.py
```diff
@@ -0,0 +1,306 @@
...
+            )
+        tgraph_type = "TGraphErrors"
+
+    elif any(asym_errors_bool):
+        if not all(asym_errors_bool):
+            raise ValueError(
+                "Have to specify all: x_errors_low, x_errors_high, y_errors_low, y_errors_high"
```
jpivarski, line 88:
```suggestion
                "uproot.to_TGraph requires all of the following: x_errors_low, x_errors_high, y_errors_low, y_errors_high"
```

src/uproot/writing/interpret.py
```diff
@@ -0,0 +1,306 @@
...
+            )
+        if any(sym_errors_bool):
+            raise ValueError(
+                "Can's specify both symetrical errors and asymetrical errors"
```
jpivarski, line 92:
```suggestion
                "uproot.to_TGraph can accept symmetrical errors OR asymmetrical errors, but not both"
```

src/uproot/writing/interpret.py
```diff
@@ -0,0 +1,306 @@
...
+        raise ValueError("Arrays x and y must have the same length!")
+    if len(x) == 0:
+        raise ValueError("Arguments and values arrays can't be emty")
+    if len(x.shape) != 1:
+        raise ValueError(f"x has to be 1D, but is {len(x.shape)}D!")
+    if len(y.shape) != 1:
+        raise ValueError(f"y has to be 1D, but is {len(y.shape)}D!")
+
+    if minY is None:
+        new_minY = min(x, default=0)
```
jpivarski, line 140: If `x` is an array, Python's `min` and `max` (several lines below) is much slower than `np.min` and `np.max`. Even if we don't particularly care about performance for something as small as the number of data points in a TGraph, it would be less confusing if we always use vectorized routines with arrays and non-vectorized routines with Python sequences.

src/uproot/writing/interpret.py
```diff
@@ -0,0 +1,306 @@
...
+    elif not isinstance(minY, (int, float)):
```
jpivarski, line 141:
```suggestion
    elif not isinstance(minY, numbers.Real):
```

and `import numbers` at the top of the file. (It's a Python standard library.) If you only check for `(int, float)`, you'll miss NumPy integers:

```python
>>> isinstance(np.int64(123), (int, float))
False
```

src/uproot/writing/interpret.py
```diff
@@ -0,0 +1,306 @@
...
+        raise ValueError(f"fMinium has to be None or a number! But is {type(minY)}")
+    else:
+        new_minY = minY
+
+    if maxY is None:
+        new_maxY = max(x, default=0)
+    elif not isinstance(maxY, (int, float)):
```
jpivarski, line 148:
```suggestion
    elif not isinstance(maxY, numbers.Real):
```

jpivarski (changes requested):
I didn't carefully check all the cases as described in lines 246 through 252 of interpret.py (the docstring of `to_TGraph`). But it's good that there's an effective flow-chart/decision tree for these cases because the user will need to internalize it.

I only have a few code clean-up things below, and a likely solution to the test failures.

src/uproot/writing/interpret.py
Pepesob: When I started to make this function I wanted all the arrays to be passed as arguments, but later I just changed it to accept dict like object.

src/uproot/writing/interpret.py
```diff
@@ -130,23 +132,23 @@ def _to_TGraph(
     if len(x) != len(y):
         raise ValueError("Arrays x and y must have the same length!")
     if len(x) == 0:
-        raise ValueError("Arguments and values arrays can't be emty")
+        raise ValueError("x and Y arrays can't be emty")
```
jpivarski, line 135:
```suggestion
        raise ValueError("uproot.as_TGraph x and y arrays can't be empty")
```

src/uproot/writing/interpret.py
```diff
@@ -130,23 +132,23 @@ def _to_TGraph(
...
     if len(y.shape) != 1:
         raise ValueError(f"y has to be 1D, but is {len(y.shape)}D!")
 
     if minY is None:
-        new_minY = min(x, default=0)
-    elif not isinstance(minY, (int, float)):
-        raise ValueError(f"fMinium has to be None or a number! But is {type(minY)}")
+        new_minY = np.min(x)
+    elif not isinstance(minY, numbers.Real):
+        raise ValueError(f"minY has to be None or a number! But is {type(minY)}")
```
jpivarski, line 144:
```suggestion
        raise ValueError(f"uproot.as_TGraph minY has to be None or a number, not {type(minY)}")
```

src/uproot/writing/interpret.py
```diff
@@ -130,23 +132,23 @@ def _to_TGraph(
...
     else:
         new_minY = minY
 
     if maxY is None:
-        new_maxY = max(x, default=0)
-    elif not isinstance(maxY, (int, float)):
-        raise ValueError(f"fMinium has to be None or a number! But is {type(maxY)}")
+        new_maxY = np.max(x)
+    elif not isinstance(maxY, numbers.Real):
+        raise ValueError(f"minY has to be None or a number! But is {type(maxY)}")
```
jpivarski, line 151:
```suggestion
        raise ValueError(f"uproot.as_TGraph minY has to be None or a number, not {type(maxY)}")
```

jpivarski (approved):
Very nice! I'll make the suggested changes (below) myself, and then merge this PR if the tests pass. Thank you very much!

@grzanka said,

> This is a nice design ! Can it be extended (maybe in future PR) to handle TMultiGraph as well ? I would like to save plots with different data series. Each serie would be separate collection of x,y pairs. Each serie may have different number of points.

That sounds like a good future PR. Would that be similar enough to include in the `as_TGraph` function (which already covers three different classes), or would the interface be different enough to warrant a new function?

jpivarski:
@all-contributors please add @Pepesob for code

Issue #1267, opened 2024-08-09
Inconsistent Reading of Grouped Branches 


SamKelson:
**Describe the bug:**
When attempting to access branches of a TTree in a ROOT file using uproot, an error occurs if grouped branches are included in a set with other branches. Specifically, the `arrays` method throws an error about grouped branches when attempting to access them alongside other branches. However, the error thrown is inconsistent as there are certain combinations of branches with the grouped branch that do not throw an error.

**To Reproduce:**
1. Download public file from [cernbox](https://cernbox.cern.ch/remote.php/dav/public-files/wJGWzAyirlWE6QV/DAOD_PHYSLITE.37233417._000052.pool.root.1)
2. Use uproot to open the file and access the `CollectionTree` TTree:
```
file = uproot.open("DAOD_PHYSLITE.37233417._000052.pool.root.1")['CollectionTree']
```
3. Use the `arrays` method of TTree to access the following keys provided via `set` `s:
```
tracking_keys = set({'GSFTrackParticlesAuxDyn.z0', 'GSFTrackParticlesAuxDyn.numberOfNextToInnermostPixelLayerHits', 'GSFTrackParticlesAuxDyn.definingParametersCovMatrixDiag', 'GSFTrackParticlesAuxDyn.expectNextToInnermostPixelLayerHit', 'GSFTrackParticlesAuxDyn.numberOfSCTHits', 'GSFTrackParticlesAuxDyn.numberOfPixelHits', 'GSFTrackParticlesAuxDyn.phi', 'GSFTrackParticlesAuxDyn.theta', 'GSFTrackParticlesAuxDyn.expectInnermostPixelLayerHit', 'GSFTrackParticlesAuxDyn.qOverP', 'GSFTrackParticlesAuxDyn.numberOfInnermostPixelLayerHits', 'GSFTrackParticlesAuxDyn.d0', 'GSFTrackParticlesAuxDyn.originalTrackParticle', 'GSFTrackParticlesAuxDyn.vz', 'GSFTrackParticlesAuxDyn.definingParametersCovMatrixOffDiag', 'GSFTrackParticlesAuxDyn.chiSquared'})
[...]

PR #1269, opened 2024-08-12
fix: apply `TBranch.array`'s veto of `AsGrouped` from `ranges_or_baskets` to `HasBranches.arrays` and `HasBranches.iterate`

jpivarski:
Very rarely, TBranches with subbranches contain data, usually (unnecessary/ignorable) counts of the subbranches' data per entry. The interpretation for any TBranch with subbranches is `uproot.AsGrouped`, which ignores any of the TBranch's own data and makes it a RecordArray of its subbranches instead.

This is correctly accounted for in the `arrays` (dict of outputs) by assigning the corresponding value to `None` (so that [this event loop](https://github.com/scikit-hep/uproot5/blob/80e780347c2dc40c7004a41c22dfc743ba33f008/src/uproot/behaviors/TBranch.py#L3091-L3110) won't wait for any data to be read).

In the usual case, the TBranch with subbranches does not have any data, so no ranges are added to `ranges_or_baskets`. (`branch.entries_to_ranges_or_baskets` returns an empty list.) However, if the TBranch with subbranches has data, then it will be added to `ranges_or_baskets` for reading. Whether it actually gets read or not depends on the order that the physical disk or network returns TBasket data (since [this event loop](https://github.com/scikit-hep/uproot5/blob/80e780347c2dc40c7004a41c22dfc743ba33f008/src/uproot/behaviors/TBranch.py#L3091-L3110) stops watching after it receives the number of arrays it is expecting, and `arrays` has a `None` for `uproot.AsGrouped` arrays, so they don't count against the expected number).

#1267 encountered hit this issue because the ROOT file does have TBranches with subbranches that have TBaskets (the unusual case). The error message was probably intended to be internal, or it was used to be user-facing but was leftover from an era before the policy was to ignore any data in `uproot.AsGrouped`, rather than report it as an error to the user. The code that intends to ignore the data can be found in `TBranch.array`:

https://github.com/scikit-hep/uproot5/blob/80e780347c2dc40c7004a41c22dfc743ba33f008/src/uproot/behaviors/TBranch.py#L1807-L1816

(i.e. skip any TBranches with `uproot.AsGrouped` interpretation, regardless of whether they might contain data), but the equivalent hadn't been applied to `HasBranches.arrays` or `HasBranches.iterate`:

https://github.com/scikit-hep/uproot5/blob/80e780347c2dc40c7004a41c22dfc743ba33f008/src/uproot/behaviors/TBranch.py#L810-L820

https://github.com/scikit-hep/uproot5/blob/80e780347c2dc40c7004a41c22dfc743ba33f008/src/uproot/behaviors/TBranch.py#L1036-L1056

(Note that the `if` statement skips cached data but not `uproot.AsGrouped`.)

This PR makes all three array-fetching functions consistent in their exclusion of `uproot.AsGrouped` interpretations from `ranges_or_baskets`. (Other array-fetching functions call these, rather than preparing their own `ranges_or_baskets`.)

### DIFF
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -811,7 +811,10 @@ class HasBranches(Mapping):
         checked = set()
         for _, context in expression_context:
             for branch in context["branches"]:
-                if branch.cache_key not in checked:
+                if branch.cache_key not in checked and not isinstance(
+                    branchid_interpretation[branch.cache_key],
+                    uproot.interpretation.grouped.AsGrouped,
+                ):
                     checked.add(branch.cache_key)
                     for (
                         basket_num,
@@ -1035,7 +1038,10 @@ class HasBranches(Mapping):
                 checked = set()
                 for _, context in expression_context:
                     for branch in context["branches"]:
-                        if branch.cache_key not in checked:
+                        if branch.cache_key not in checked and not isinstance(
+                            branchid_interpretation[branch.cache_key],
+                            uproot.interpretation.grouped.AsGrouped,
+                        ):
                             checked.add(branch.cache_key)
                             for (
                                 basket_num,


matthewfeickert (commented):
This indeed seems to resolve things: https://github.com/scikit-hep/uproot5/issues/1267#issuecomment-2284986203

jpivarski:
> (More reason for "us" (ATLAS people) to get a **small** PHYSLITE file in `scikit-hep-testdata`.)

(I found out just how slow my hotel wifi was by downloading that file...)

PR #1280, opened 2024-08-27
perf: if first whole TBasket has inconclusive AwkwardForth

jpivarski:
@nikoladze found a performance bug: if the first whole TBasket has inconclusive AwkwardForth, the code (before this PR) concatenated everything and ran the whole array through a generic iterable → Awkward Array transformation. The problem is that almost all of the data had been an Awkward Array already, all but the one TBasket with inconclusive AwkwardForth.

This PR adds an extra check for this case. If the list of basket-arrays is only partially Awkward, it converts the non-Awkward ones to Awkward individually so that the concatenated array doesn't have to be passed through the unnecessary generic iterable → Awkward Array transformation.

(There aren't any tests for this because it's based on private data.)

### DIFF
--- a/src/uproot/interpretation/objects.py
+++ b/src/uproot/interpretation/objects.py
@@ -512,6 +512,27 @@ input stream
 
             start = stop
 
+        # If *some* of the baskets are Awkward and *some* are not,
+        # convert the ones that are not, individually.
+        if any(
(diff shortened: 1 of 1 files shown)


matthewfeickert:
@jpivarski Could we get another patch release with this in? :grimacing: Sorry, I know we've been asking for them frequnetly.

PR #1285, opened 2024-09-06
fix: RNTuple ug-fixing offset array concatenation, adding filter_name

giedrius2020:
Changes made in RNTuple model:

- Bug-fix: when reading arrays from more than one cluster, data was not identical to TTree. The reason for this was improper array concatenation in case of offset values. More information in here: https://gist.github.com/giedrius2020/c1ca87784779418ee3e938aaca31ff75 
- Added a function to filter wanted key values in `def keys()`
- Minimal change: changed existing "filter_names" variables to "filter_name", because that is how the variable was called in TTree model.

### DIFF
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -497,8 +497,10 @@ in file {self.file.file_path}"""
         arrays = [self.read_col_page(ncol, i) for i in cluster_range]
 
         # Check if column stores offset values for jagged arrays (splitindex64) (applies to cardinality cols too):
-        if dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex64"] or\
-           dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex32"]:
+        if (
+            dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex64"]
+            or dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex32"]
+        ):
             # Continue new cluster values from the last value of previous cluster.
 
             # Extract the last offset values:
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -506,7 +506,7 @@ in file {self.file.file_path}"""
                 arr[-1] for arr in arrays[:-1]
             ]  # First value always zero, therefore skip first arr.
             # Compute cumulative sum using itertools.accumulate:
-            last_offsets = [*list(accumulate(last_elements))]
+            last_offsets = list(accumulate(last_elements))
             # Add the offsets to each array
             for i in range(1, len(arrays)):
                 arrays[i] += last_offsets[i-1]
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -501,14 +501,13 @@ in file {self.file.file_path}"""
             dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex64"]
             or dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex32"]
         ):
-            # Continue new cluster values from the last value of previous cluster.
-
             # Extract the last offset values:
-            last_elements = [arr[-1] for arr in arrays[:-1]]
+            last_elements = [arr[-1] for arr in arrays[:-1]]  # First value always zero, therefore skip first arr.
             # Compute cumulative sum using itertools.accumulate:
-            last_offsets = [0, *list(accumulate(last_elements))]
+            last_offsets = [*list(accumulate(last_elements))]
             # Add the offsets to each array
-            arrays = [arr + offset for arr, offset in zip(arrays, last_offsets)]
+            for arr, offset in zip(arrays[1:], last_offsets):
+                arr += offset
             # Remove the first element from every sub-array except for the first one:
             arrays = [arrays[0]] + [arr[1:] for arr in arrays[1:]]
 
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -7,9 +7,9 @@ from __future__ import annotations
 
 import struct
 from collections import defaultdict
+from itertools import accumulate
 
 import numpy
-from itertools import accumulate
 
 import uproot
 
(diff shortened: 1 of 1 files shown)


src/uproot/models/RNTuple.py
```diff
@@ -60,8 +61,21 @@ def _keys(self):
...
-        return self._keys
+    def keys(
+        self,
+        *,
+        filter_name=None,
+        filter_typename=None,
+        filter_branch=None,
+        recursive=False,
+        full_paths=True,
+        ignore_duplicates=False,
```
ariostas, line 72: I don't think it makes sense to have these extra arguments already if they don't do anything.

`ignore_duplicates` shouldn't exist since that's not an issue with RNTuples, and `filter_branch` also shouldn't exist since there are no "branches". The rest would be good to implement at some point.

src/uproot/models/RNTuple.py
```diff
@@ -563,9 +589,14 @@ def arrays(
             if "column" in key and "union" not in key:
                 key_nr = int(key.split("-")[1])
                 dtype_byte = self.column_records[key_nr].type
+
+                # Check if column stores offset values for jagged arrays (applies to cardinality cols too):
+                is_offset_col = dtype_byte == 14
```
ariostas, line 594: I feel like it might make more sense to have this logic inside `read_col_pages`

ariostas, line 594: Instead of hardcoding 14, it would be more readable to use `uproot.const.rntuple_col_type_to_num_dict["splitindex64"]`. I imagine this also applies to `splitindex32`.

ariostas:
Thank you for catching this @giedrius2020! I think the issue could be solved in a simpler way by adjusting some of the logic in here. I'll think back at what my reasoning was for this piece.

https://github.com/scikit-hep/uproot5/blob/012df94ccace9147d12aae8710585318abacda3d/src/uproot/models/RNTuple.py#L511-L521

ariostas:
Could you try checking if changing the last few lines to

```python
if delta:
    if tracker > 0:
        res[tracker] -= cumsum
    cumsum += numpy.sum(res[tracker:tracker_end])
tracker = tracker_end 
```
fixes the issue?

davidlange6:
perhaps best that we separate the api change to keys() from the bug fix?

otherwise, perhaps beyond this PR, is there some test in uproot intended to check the physics correctness of reading back files?

ariostas:
> I tried your suggestion (while disabling my changes) and it did not help.

Ah okay, thanks

> perhaps best that we separate the api change to keys() from the bug fix?

Since the RNTuple stuff isn't stable yet, I wouldn't worry much about separating things

> otherwise, perhaps beyond this PR, is there some test in uproot intended to check the physics correctness of reading back files?

I've added some tests, but the issue was that I didn't have a test file that was large enough to have multiple clusters, so that's why I hadn't seen this bug

src/uproot/models/RNTuple.py
giedrius2020: I did a lazy copy-paste from TTree keys() in this case, to solve missing **recursive** argument in dask.
I agree that they should be removed to avoid confusion. I will edit PR and remove the ones you mentioned.

src/uproot/models/RNTuple.py
giedrius2020: Good idea, I wanted to do that, but did not find which variables to use.
I will edit PR.

giedrius2020: Good idea, I wanted to do that, but did not find which variables to use.
I will edit PR.

giedrius2020: Ok, I will it **read_col_pages** scope

src/uproot/models/RNTuple.py
ariostas: Ah okay, that makes sense. Then let's just remove the ones that don't make sense and leave a TODO comment for the ones that haven't been implemented.

src/uproot/models/RNTuple.py
```diff
@@ -480,10 +493,27 @@ def read_pagedesc(self, destination, desc, dtype_str, dtype, nbits, split):
...
+            or dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex32"]
+        ):
+            # Continue new cluster values from the last value of previous cluster.
+
+            # Extract the last offset values:
+            last_elements = [arr[-1] for arr in arrays[:-1]]
+            # Compute cumulative sum using itertools.accumulate:
+            last_offsets = [0, *list(accumulate(last_elements))]
+            # Add the offsets to each array
+            arrays = [arr + offset for arr, offset in zip(arrays, last_offsets)]
```
ariostas, line 511: It would be better to try to do operations in place if possible.
```suggestion
            for arr, offset in zip(arrays, last_offsets):
                arr += offset
```

Also, you could skip the first one since the offset is always 0

giedrius2020: I see, indeed, your suggestion helps to save used memory. 

I have just committed both of these suggestions.

src/uproot/models/RNTuple.py
```diff
@@ -480,10 +493,28 @@ def read_pagedesc(self, destination, desc, dtype_str, dtype, nbits, split):
...
+        ):
+            # Extract the last offset values:
+            last_elements = [
+                arr[-1] for arr in arrays[:-1]
+            ]  # First value always zero, therefore skip first arr.
+            # Compute cumulative sum using itertools.accumulate:
+            last_offsets = [*list(accumulate(last_elements))]
+            # Add the offsets to each array
+            for arr, offset in zip(arrays[1:], last_offsets):
+                arr += offset
```
ariostas, line 512: It's a bit annoying that the linter is complaining about this. I guess it will have to be
```suggestion
            for i in range(1, len(arrays)):
                arrays[i] += last_offsets[i-1]
```

src/uproot/models/RNTuple.py
```diff
@@ -480,10 +493,28 @@ def read_pagedesc(self, destination, desc, dtype_str, dtype, nbits, split):
...
+        if (
+            dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex64"]
+            or dtype_byte == uproot.const.rntuple_col_type_to_num_dict["splitindex32"]
+        ):
+            # Extract the last offset values:
+            last_elements = [
+                arr[-1] for arr in arrays[:-1]
+            ]  # First value always zero, therefore skip first arr.
+            # Compute cumulative sum using itertools.accumulate:
+            last_offsets = [*list(accumulate(last_elements))]
```
ariostas, line 509: Also, this can now be simplified

```suggestion
            last_offsets = list(accumulate(last_elements))
```

davidlange6:
can one control cluster size in rntuple writing? indeed, i guess one doesn't want a few 100 MB of test file around

ariostas:
> can one control cluster size in rntuple writing?

I'm not sure, but that would be nice. I'll look into it

ariostas (approved):
Thank you @giedrius2020, this looks great. The test that is failing is unrelated.

I'm going to try to generate a small RNTuple with multiple clusters so that we can add a test for this.

ariostas:
I added a new test file in https://github.com/scikit-hep/scikit-hep-testdata/pull/159. Let's add a new test once that gets merged and released.

jpivarski:
That worked! I'm updating again to be sure that we get all of the 3.13 tests.

jpivarski:
Now this should include the pyodide-build, and I'll enable auto-merge because I'm sure it will pass.

PR #1287, opened 2024-09-12
test: disable test that uses UNL file

ariostas:
The UNL server started requiring authentication to access [this file](https://xrootd-local.unl.edu:1094//store/user/AGC/nanoAOD/TT_TuneCUETP8M1_13TeV-powheg-pythia8/cmsopendata2015_ttbar_19980_PU25nsData2015v1_76X_mcRun2_asymptotic_v12_ext3-v1_00000_0000.root"), so the test is timing out.

Maybe this was just an accidental change in the configuration, but I'm not sure who to contact so I disabled the test for now.

### DIFF
--- a/tests/test_1146_split_ranges_for_large_files_over_http.py
+++ b/tests/test_1146_split_ranges_for_large_files_over_http.py
@@ -1,8 +1,12 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/uproot5/blob/main/LICENSE
 
+import pytest
 import uproot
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
"Large file in HTTP"; we can't put it on scikit-hep-testdata. Also, this is directly a test of XRootD, so if it were a local file, it wouldn't be testing what it's supposed to test.

I'll merge this PR because it doesn't look like it needs any additions.

PR #1289, opened 2024-09-16
chore(deps): add xxhash dependency

### LINKED ISSUE #1271: `xxhash` is currently a strict dependency for reading RNTuples
`xxhash` is currently a strict dependency for reading RNTuples even though it's not listed on `pyproject.toml`. The dependent code is here:

https://github.com/scikit-hep/uproot5/blob/ba3269b31cd791589bbd3f6d4634dbe606b9f997/src/uproot/compression.py#L433-L439

which could simply be removed to remove the dependency.

For writing, there will be a strict dependency and there is no way to get around it, but maybe we can make `uproot.extras.xxhash` pick between `xxhash` and `ppxxh` (which is pure Python, so it works in WASM).

ariostas:
`xxhash` is currently a strict dependency for reading RNTuples (see #1271). We could simply remove the line that uses `xxhash` to verify checksums, but given that it is now available in Pyodide and it will unquestionably be needed for writing RNTuples, then it makes sense to add it as a dependency now.

Once this is merged, I will also start using it more extensively for RNTuple reading.

Closes #1271.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,7 @@ classifiers = [
 dependencies = [
   "awkward>=2.4.6",
   "cramjam>=2.5.0",
+  "xxhash",
   "numpy",
   "fsspec",
   "packaging",
@@ -67,7 +68,6 @@ s3 = ["s3fs"]
 test = [
   "isal",
   "deflate",
-  "xxhash",
   "minio",
   "aiohttp",
   "fsspec-xrootd",


ariostas:
The new version of Dask is breaking things. I'll look into it.

jpivarski (approved):
Let's do it! I'll merge this and then release Uproot 5.4.0.

ariostas:
It seems like the Pyodide test that reads a TTree over http is a little finicky. I think the server is not fast enough and it ends up timing out since there's some overhead from running things in Pyodide. I've seen it fail a few times in the CI already. I'll replace it with one that gets a smaller file from GitHub to ensure it's more stable.

PR #1293, opened 2024-09-19
docs: Uproot only supports Pythons 3.8 through 3.12

jpivarski:
Okay, so I'm leaving this as manual because the automated version doesn't show a range—it lists every Python version separately. This would make the badge too big, and I don't want the badges to use so much space. They're carefully arranged in two rows just to keep them from being overwhelming.


jpivarski:
The auto-merge will fail if the dask-awkward circular import problem (dask-contrib/dask-awkward#544) is still outstanding (not checked into a release for CI to pick up).

PR #1304, opened 2024-09-25
chore(ci): switch to miniforge

ariostas:
Miniforge and Mambaforge are now equivalent, so the latter has been deprecated. There are a bunch of warning messages about this on CI runs, so this PR is mainly to clean those up. The rest of the warnings are out of our control, but I've seen that there's open PRs that will fix some of them.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -21,11 +21,6 @@ jobs:
     runs-on: ${{ matrix.platform }}
     timeout-minutes: 30
 
-    # Required for miniconda to activate conda
-    defaults:
-      run:
-        shell: bash -l {0}
-
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


jpivarski (approved):

.github/workflows/build-test.yml
```diff
@@ -34,14 +34,16 @@ jobs:
         with:
           auto-update-conda: true
           python-version: ${{ matrix.python-version }}
-          miniforge-variant: Mambaforge
-          use-mamba: true
+          miniforge-version: latest
+
+      - name: Check active Python version
+        run: python -c "import sys; assert '.'.join(str(s) for s in sys.version_info[:2]) == '${{ matrix.python-version }}', f'{version} incorrect!'"
```
ariostas, line 40: Will this work with `3.13t`?

henryiii: With modifications it could.

PR #1308, opened 2024-10-03
ci: test on 3.13

henryiii:
Also using uv for the vanilla job


henryiii:
@jpivarski, it would be good to get patch releases before 3.13 is released, on Tuesday!

jpivarski (approved):

PR #1319, opened 2024-10-20
fix: don't let Awkward Array attempt to compare a big-endian array

jpivarski:


jpivarski:
The initial fix passed tests and the merge with main didn't add anything that might conflict with the initial fix, so I'm just going to accept this without waiting for all the tests to run. (Maybe I'm running too many things, but GitHub seems slow to provide runners.)

PR #1320, opened 2024-10-23
perf: improve performance for read coalescing

### LINKED ISSUE #1257: uproot4 to uproot5 TTree to DataFrame arrays slower
Hello.

I run a script to load a TTree as a DataFrame with 5 million rows and 3 columns, and measure the time taken. Here's an MWE for ipython:
```python
%timeit uproot.open('tree.root')['time'].arrays(['year','month','day'],library='pd')
```

For uproot 5.3.7 with pandas 2.2.2, 19.2 s ± 177 ms per loop
For uproot 4.1.9 with pandas 1.3.5, 1.61 s ± 9.27 ms per loop

The value reported is mean ± std. dev. of 7 runs, 1 loop each. We can see uproot5 is 4 times slower than uproot4. The discrepancy increases further when more columns are loaded, leading to more than ten minutes of time for what previously took seconds.

I just want to know if there has been any major change that can cause such a reduction of load time for large TTree to DataFrame. Also, if any more checks should be done before I reach a conclusion, that is super helpful.

Thanks a lot for your time.

nikoladze:
This fixes #1257 - the current read coalescing algorithm has a loop over ranges (to determine `Cluster.stop`) for every range added to a `Cluster` leading to a somewhat quadratic scaling. This can be fixed by having a private `_stop` attribute that is updated in an `append` method. The downside is that now `.append` has to be used to append new ranges to a `Cluster`. Since this is not really a user facing class i think that should be acceptable.

Using the example TTree from #1257 

```python
import uproot
import numpy as np

num_baskets = 1_000
num_per_basket = 1_000
num_branches = 50
with uproot.recreate("test.root") as f:
    for i_chunk in range(num_baskets):
        chunk = {f"branch{i}": np.random.rand(num_per_basket) for i in range(num_branches)}
        if "tree" in f:
            f["tree"].extend(chunk)
        else:
            f["tree"] = chunk
```

To be read via
```python
import uproot

def load(filename):
    with uproot.open({filename: "tree"}) as tree:
        return tree.arrays(library="np")
```

**Before** this PR i get on my laptop
```python
In [6]: %timeit -n1 -r2 __ = load("test.root")
1min 19s ± 3.2 s per loop (mean ± std. dev. of 2 runs, 1 loop each)
```

**After** the changes i get
```python
In [2]: %timeit __ = load("test.root")
4.62 s ± 90.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/src/uproot/source/coalesce.py
+++ b/src/uproot/source/coalesce.py
@@ -49,6 +49,7 @@ class RangeRequest:
 @dataclass
 class Cluster:
     ranges: list[RangeRequest]
+    _stop: int | None = None
 
     @property
     def start(self):
@@ -57,7 +58,13 @@ class Cluster:
 
     @property
     def stop(self):
-        return max(range.stop for range in self.ranges)
+        # Note: in order for this to work one has to use `append` for adding new ranges
+        return self._stop
+
+    def append(self, range):
+        if self._stop is None or range.stop > self._stop:
+            self._stop = range.stop
+        self.ranges.append(range)
 
     def __len__(self):
         return self.stop - self.start
@@ -88,7 +95,7 @@ def _merge_adjacent(ranges: list[RangeRequest], config: CoalesceConfig):
         if cluster.ranges and current_range.start - cluster.stop > config.max_range_gap:
             yield cluster
             cluster = Cluster([])
-        cluster.ranges.append(current_range)
+        cluster.append(current_range)
     if cluster.ranges:
         yield cluster
 


nsmith- (approved):
LGTM

nsmith-:
After the change, is the read coalescing algorithm time negligble?

nikoladze:
> After the change, is the read coalescing algorithm time negligble?

yes, `_merge_adjacent` for the 50k ranges in the example now takes in the order of 30ms

jpivarski (approved):
It looks good to me, too. As we discussed in person, if `stop` had previously been accessed every time, this replaces an _O(n²)_ algorithm with an _O(n)_ one, and we always want to do that.

It's also possible that the local-fsspec (default `handler`) versus `uproot.MemmapSource` performance difference that I showed today is affected by this. Before we use that as an argument to change the default `handler` to `uproot.MemmapSource`, I need to reproduce that plot to see whether the effect entirely goes away. (If it only partially goes away, and `uproot.MemmapSource` is still better in many-thread systems, then it would still make sense to change the default for local files.)

Since we're all happy with this change, I'll merge it. Thanks!

jpivarski:
Although I'd consider a _O(n²)_ → _O(n)_ fix to be a serious bug-fix, I'm labeling this commit as "perf" for clarity, to indicate that user-visible behavior doesn't change.

Issue #1326, opened 2024-10-30
pyodide-build test stopped working


jpivarski:
@ariostas, could you look at the [output of the pyodide-build test](https://github.com/scikit-hep/uproot5/actions/runs/11600993444/job/32302709575)?

It's failing because there's a deprecation warning:

```
============================= test session starts ==============================
platform linux -- Python 3.12.7, pytest-8.3.3, pluggy-1.5.0 -- /opt/hostedtoolcache/Python/3.12.7/x64/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase(PosixPath('/home/runner/work/uproot5/uproot5/.hypothesis/examples'))
rootdir: /home/runner/work/uproot5/uproot5
configfile: pyproject.toml
plugins: hypothesis-6.115.6, asyncio-0.24.0, anyio-4.6.2.post1, timeout-2.3.1, pyodide-0.58.3
asyncio: mode=Mode.STRICT, default_loop_scope=None
timeout: 600.0s
timeout method: signal
timeout func_only: False
collecting ... collected 5 items
tests-wasm/test_1272_basic_functionality.py::test_read_ttree[chrome] ERROR [ 20%]
tests-wasm/test_1272_basic_functionality.py::test_write_ttree[chrome] ERROR [ 40%]
tests-wasm/test_1272_basic_functionality.py::test_read_rntuple[chrome] ERROR [ 60%]
tests-wasm/test_1272_basic_functionality.py::test_read_ttree_http[chrome] ERROR [ 80%]
client_config = None
ignore_certificates = False
ignore_proxy = False
init_args_for_pool_manager = None
keep_alive = True
remote_server_addr = 'http://localhost:49715'
self       = <selenium.webdriver.chromium.remote_connection.ChromiumRemoteConnection object at 0x7f384eb73a70>
/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/selenium/webdriver/remote/remote_connection.py:302: DeprecationWarning
_______________ ERROR at setup of test_read_rntuple_http[chrome] _______________
request = <SubRequest 'selenium_module_scope' for <Function test_read_ttree[chrome]>>
runtime = 'chrome'
web_server_main = ('127.0.0.1', 52009, PosixPath('/tmp/tmpkyznra1g/http-server.log'))
playwright_browsers = {}
    @pytest.fixture(scope="module")
[...]

PR #1327, opened 2024-10-31
fix(ci): pin selenium version to fix pyodide tests

ariostas:
There was a bug in the new Selenium version that is causing the Pyodide tests to fail (see #1326). I'm pinning Selenium to the previous version until they make a new release.


jpivarski (approved):
Sounds good! (We talked about this in our meeting.) I'll merge it now.

PR #1332, opened 2024-11-07
chore: remove Python 3.8, ensure Python 3.13 is included

jpivarski:


ariostas:
There's a bit of clean up that can be done here.
https://github.com/scikit-hep/uproot5/blob/a331467c117a337c7336e8b593f26177444dd708/src/uproot/source/fsspec.py#L136-L153

I'm not sure if it would be better to do it here or in another PR.

jpivarski:
If that section of code can be cleaned up _because_ we aren't supporting Python 3.8 anymore, then it makes sense to include it in this PR. I think you have write access to this PR, so go ahead and do what you think should be done. I won't merge it until you say you're done.

ariostas:
Yeah, Python 3.9 introduced `asyncio.to_thread`, so that chunk of code to support 3.8 can now be dropped. I'll add a commit.

I tried to find other potential cleanups, but I think that was the only one.

ariostas:
I fixed the tests that were failing. I introduced a fix for numpy v1, but I think it's small enough that it should be fine to include in here. It should now be good to go, at least from my side.

jpivarski:
This looks great; thank you! I can't approve it as a PR because I started it, but I approve anyway. Since you're done, I'll merge it and make Uproot 5.5.0.

(I just checked to see if any tests need to be added to the branch protection—no, thanks to the DAG that ends in a test called "Pass", we can change which tests run and still have them all be required now.)

PR #1334, opened 2024-11-07
chore(ci): Disable RNTuple tests for now and unpin selenium version

ariostas:
Next week is the planned release of version 1.0 of the RNTuple spec. So soon I'll be starting to delete old RNTuple test files from `scikit-hep-testdata` and generate new ones with the v1.0 spec. This will break the CI tests, so I'm disabling the RNTuple tests so that it doesn't affect other PRs.

I'm also unpinning the selenium version that was pinned in #1327 since they released a fixed version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,6 @@ test = [
 ]
 test-pyodide = [
   "pytest>=6",
-  "selenium<=4.25.0",  # unpin once >4.26.0 is available
   "pytest-pyodide",
   "pytest-timeout",
   "scikit-hep-testdata"
--- a/tests-wasm/test_1272_basic_functionality.py
+++ b/tests-wasm/test_1272_basic_functionality.py
@@ -70,6 +70,7 @@ def test_write_ttree(selenium):
 
 
 # Taken from test_1191_rntuple_fixes.py
+@pytest.mark.skip(reason="Skipping until test files are available with RNTuple v1.0")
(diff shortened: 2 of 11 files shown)


pyproject.toml
```diff
@@ -82,7 +82,6 @@ test = [
 ]
 test-pyodide = [
   "pytest>=6",
-  "selenium<=4.25.0",  # unpin once >4.26.0 is available
```
jpivarski, line 85: This just removes `"selenium<=4.25.0"` without replacing it with a generic `"selenium"`. Did you mean to do that? The tests still need to have selenium installed, right?

jpivarski (approved):
These are good things to do (and, in my opinion, fine that they're both in the same PR; they're similar enough).

pyproject.toml
ariostas: Yeah, selenium is a dependency of `pytest-pyodide`, so I was thinking to just leave dependency/version management to them

jpivarski: I guess that makes sense. I never know what to do about transitive dependencies. Perhaps if you use it directly (`import selenium`), then you should declare it as a dependency, just in case pytest-pyodide ever removes it or switches, but not if you don't use it directly.

ariostas: Hmm that's a good point. It is never used directly, so it should be fine.

PR #1338, opened 2024-11-14
chore: update reading support for RNTuple v1

ariostas:
This PR adds support for the first stable version of the RNTuple binary format (v1.0.0.0). It's mostly just rearranging some things and adjusting some constants. Tests will fail until https://github.com/scikit-hep/scikit-hep-testdata/pull/164 is merged and released. This is still a work in progress, but it seems to mostly work already.

**Note that this will completely break support for any files produced before v1**


ariostas:
This is now ready for review. All the previous functionality works again, and I fixed a small bug. All the tests should pass after https://github.com/scikit-hep/scikit-hep-testdata/pull/164 is merged and released.

The remaining things that are not supported are:
 - Real32Trunc and Real32Quant columns
 - Suppressed columns
 - Objects serialized with the ROOT streamer (which will probably never work)

Instead of waiting until I finish implementing the first two, I think it would be good to get this out so that people can try it out and we can get some feedback.

PR #1347, opened 2024-12-06
feat: support reading of custom-length RNTuple floats

ariostas:
This PR implements the reading of `Real32Trunc` and `Real32Quant` fields, which have a variable number of bits in the ranges 10-31 and 1-32, respectively.

This already works, but I'll clean it up and add a test. I might also add support for suppressed columns in this PR. I'll update the title and description if I do so.

### DIFF
--- a/src/uproot/const.py
+++ b/src/uproot/const.py
@@ -242,6 +242,10 @@ rntuple_delta_types = (
     rntuple_col_type_to_num_dict["splitindex32"],
     rntuple_col_type_to_num_dict["splitindex64"],
 )
+rntuple_custom_float_types = (
+    rntuple_col_type_to_num_dict["real32trunc"],
+    rntuple_col_type_to_num_dict["real32quant"],
+)
 
 
 class RNTupleLocatorType(IntEnum):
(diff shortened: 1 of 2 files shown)


ariostas:
It ended up only taking a few lines to support suppressed columns. I'll improve the reading of floats and add some tests.

ariostas:
This PR is ready for review, but it needs https://github.com/scikit-hep/scikit-hep-testdata/pull/167 for the tests to pass.

Also, since there is still an issue with Dask, we'll just have to wait until that is resolved.

ariostas:
@jpivarski I fixed the issue with Numpy 1 and I improved the tests. I still had to use `np.isclose` for some of the comparisons due to small differences, probably from ROOT using a different way to round things.

jpivarski (approved):
It's fine for comparisons with ROOT to involve `isclose` with tight tolerances (around the scale of numerical precision). What I meant was that if you know you'll be truncating at the Nth digit, it's better to test against an expected value that's truncated at the Nth digit, with `isclose` picking up any small errors, rather than using `isclose` to check for differences with respect to the original value. My reasoning was just that the Nth digit might be a huge error and leaving a window wide open like that could fail to identify some errors.

As a whole, this PR looks great and I'd say it's ready to be merged. This is the PR that brings RNTuple-reading up to 100% coverage, right?

ariostas:
> This is the PR that brings RNTuple-reading up to 100% coverage, right?

Yes, as far as I can tell, after this PR we'll have 100% coverage of the current spec.

PR #1352, opened 2024-12-16
fix: uproot was exposed in one place to dask's _task_spec overhaul

lgray:
c.f. https://github.com/dask-contrib/dask-awkward/pull/556

There don't appear to be any other places where uproot directly accesses dask innards.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ requires-python = ">=3.9"
 [project.optional-dependencies]
 dev = [
   "boost_histogram>=0.13",
-  "dask-awkward>=2023.12.1",
+  "dask-awkward>=2024.12.1",
   "dask[array,distributed]",
   "hist>=1.2",
   "pandas",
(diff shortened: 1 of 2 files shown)


lgray:
@jpivarski please merge this as soon as it's all passing, we will need a timely release as well.

lgray:
ah - of course this waits on the dask-awkward release first :-p

pfackeldey:
@lgray there's no reason to increase the `dask-awkward` version requirement, or is there? The change checks for an update in dask for the new task_spec, but still handles the old case. (I think this PR is independent of the dask-awkward PR aswell in this sense)

lgray:
And it's only for the dev extra as well, and since we check back to py39 we test all code paths.

jpivarski (approved):
I see that this follows upstream changes in Dask and has Uproot taking a new minimum dask-awkward version.

Looks good to me! I'll make a new release right after merging this PR.

PR #1356, opened 2025-01-02
feat: basic RNTuple writing

ariostas:
This PR adds basic RNTuple writing functionality. Currently, it can write an RNTuple with some simple fields, but no data. My plan for this PR is to get it to write flat columns of basic types, which it seems like it won't take too much work.

After this PR, there will still be a lot of work to fully implement RNTuple writing, but it will be nice to have the most basic functionality working so that we can test things and build up from there.


ariostas:
The writing functionality still needs a lot of work, but in the interest of keeping PRs reasonably sized, I will follow up with other PRs.

With the changes in this PR, it is now possible to write flat arrays of native data types (except bool). The files produced are able to be correctly read by ROOT, so the most basic structure and functionality is working.

Unfortunately, we are currently not able to test reading the files with ROOT in the CI, since ROOT 6.34 is still not available on conda-forge. However, I did run the tests locally and everything passes.

ianna (approved):
@ariostas - Fantastic! This PR is an essential step in moving forward! I agree with your reasoning to split further development to proceed with smaller PRs. Building docs fails, but it's unrelated to this PR. Please, go ahead and merge it when you are ready.

PR #1359, opened 2025-01-09
fix: support RNTuples with multiple cluster groups

### LINKED ISSUE #1357: RNTuple reading doesn't work when there are multiple cluster groups
While working on RNTuple reading, I realized that the reading code makes the wrong assumption that there is a single cluster group.

It should be an easy fix, but I might not get to it until the end of next week.

ariostas:
This PR fixes #1357. Now multiple cluster groups are handled correctly. Internally, it just groups all clusters together into a single group. I couldn't think of any reasons to keep track of different groups, but we can implement that later if necessary.


ariostas:
The test fails because it needs https://github.com/scikit-hep/scikit-hep-testdata/pull/168. I'll rerun the CI when it's ready. Otherwise, it's ready for review.

jpivarski (approved):
It looks great! And there's a test, which passes, so you can merge this whenever you're ready.

PR #1362, opened 2025-01-17
test: adjust for changes in example.com

jpivarski:
It returns 200, rather than 206, for our wrong-page examples now. There might be a good way to replace this test, but for now I'm disabling it. We have plenty of other tests that check the HTTP backend. (This test is from 0001, which tested HTTP without any ROOT files because it was written at a time when only the backend handlers existed.)

### DIFF
--- a/tests/test_0001_source_class.py
+++ b/tests/test_0001_source_class.py
@@ -117,6 +117,7 @@ def test_memmap_fail(use_threads, tmp_path):
             ...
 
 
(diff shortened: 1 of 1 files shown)


jpivarski:
@ianna, if you approve this PR, feel free to merge it. (I'm done!)

ianna (approved):
@jpivarski - thanks! It looks good to me. I will merge it.

PR #1365, opened 2025-01-23
chore: update pyodide in CI and add more tests

ariostas:
I updated the version of Pyodide that is used in the CI. I'm also adding an extra test to check if `AwkwardForth` works properly, given that it is needed to read some RNTuple files.


ariostas:
This PR will have to wait until the next version of Pyodide is released (which hopefully includes the fixed `awkward-cpp`). Alternatively, I could change things so that `awkward-cpp` is built every time, but I'm not sure it's worth it.

ariostas:
The new Pyodide 0.27.3 that was just released includes `awkward-cpp` 44, so now the `awkward-forth` tests pass. So this PR is now ready for review.

PR #1368, opened 2025-01-27
fix: pin dask to upper limit

ianna:


ariostas:
I've noticed that the miniconda action has been a little flaky lately. I just reran it and the test passed.

ariostas (approved):
Looks good to me. Thanks for the fix!

PR #1373, opened 2025-02-04
fix: Revert "fix: pin dask to upper limit"

ianna:
Reverts scikit-hep/uproot5#1368

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,6 @@ requires-python = ">=3.9"
 dev = [
   "boost_histogram>=0.13",
   "dask-awkward>=2024.12.1",
-  "dask[dataframe] >=2022,<2025",
   "dask[array,distributed]",
   "hist>=1.2",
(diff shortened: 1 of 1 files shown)


ianna (commented):
@pfackeldey - given that all tests pass, I think, we need to remove the upper version limit.

pyproject.toml
```diff
@@ -58,7 +58,6 @@ requires-python = ">=3.9"
 dev = [
   "boost_histogram>=0.13",
   "dask-awkward>=2024.12.1",
-  "dask[dataframe] >=2022,<2025",
```
ianna, line 61: "dask[dataframe] >=2025",

pfackeldey: "dask-awkward >=2025.2.0"

pfackeldey (approved):

pfackeldey:
Looks good to me!

PR #1377, opened 2025-02-07
test: fix broken test

ariostas:
~ROOT was updated to 6.32.8 in `conda-forge` and it seems to have some issue with Python 3.9. Given that the issue is with ROOT and not Uproot, and also that Python 3.9 will reach its eand-of-life later this year, I'm just disabling this test for Python 3.9.~

EDIT: See comments below


ariostas:
Hold on, I just realized that ROOT is only being installed in Python 3.9. Let me see what happens if I install it in other Python versions.

ariostas:
Turns out it was ROOT's fault. It probably internally changed something, but I fixed the test.

ianna (approved):
@ariostas - Great! Thanks! The tests pass, I'll merge it. Thanks.

PR #1386, opened 2025-02-22
perf: make `dask_write` a tree reduction

ikrommyd:
Similarly with https://github.com/dask-contrib/dask-awkward/pull/549, it's better if dask writing is a tree reduction for the memory of the workers. This PR does just that.


pfackeldey:
Thanks looks good :+1:

pfackeldey:
@all-contributors please add @ikrommyd for code

PR #1393, opened 2025-02-26
feat: add tree to virtual array conversion

pfackeldey:
This waits for https://github.com/scikit-hep/awkward/pull/3364 (and a corresponding awkward release). 

I may likely not have attempted the most optimal solution here. Happy for feedback & input.

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -950,11 +950,93 @@ class TrivialFormMappingInfo(ImplementsFormMappingInfo):
 
         return container
 
+    def load_virtual_arrays(
+        self,
+        tree: HasBranches,
+        keys: frozenset[str],
+        start: int,
+        stop: int,
+        decompression_executor,
+        interpretation_executor,
+        options: Any,
+    ) -> Mapping[str, AwkArray]:
+        awkward = uproot.extras.awkward()
+
+        class Generator:
+            def __init__(self, branch, attribute):
+                self.branch = branch
+                self.attribute = attribute
+
+            def __repr__(self):
+                return f"Generator({self.branch}, {self.attribute})"
+
+            def __call__(self):
+                layout = self.branch.array(
+                    entry_start=start,
+                    entry_stop=stop,
+                    interpretation_executor=interpretation_executor,
+                    decompression_executor=decompression_executor,
+                    library="ak",
+                    ak_add_doc=options.get("ak_add_doc"),
+                ).layout
+                # this is a bit of a hack, but it works for now
+                if isinstance(layout, awkward.contents.NumpyArray):
+                    return layout.data
+                elif isinstance(layout, awkward.contents.ListOffsetArray):
+                    if self.attribute == "data":
+                        return layout.content.data
+                    elif self.attribute == "offsets":
+                        return layout.offsets.data
+                    else:
+                        raise NotImplementedError()
+                else:
+                    raise NotImplementedError()
+
+        container = {}
+        for buffer_key, _ in self._form.expected_from_buffers().items():
+            form_key, attribute = self.parse_buffer_key(buffer_key)
+            branch_name = self._form_key_to_key[form_key]
+            branch = tree[branch_name]
+            container[buffer_key] = Generator(branch, attribute)
+
+        return container
+
+
+def form_with_unique_keys(form: Form, key: str) -> Form:
+    awkward = uproot.extras.awkward()
+
+    def impl(form: Form, key: str) -> None:
+        # Set form key
+        form.form_key = key
+
+        # If the form is a record we need to loop over all fields in the
+        # record and set form that include the field name; this will keep
+        # recursing as well.
+        if form.is_record:
+            for field in form.fields:
+                impl(form.content(field), f"{key}.{field}")
+
+        elif form.is_union:
+            for i, entry in enumerate(form.contents):
+                impl(entry, f"{key}#{i}")
+
+        # NumPy like array is easy
+        elif form.is_numpy or form.is_unknown:
+            pass
+
+        # Anything else grab the content and keep recursing
+        else:
+            impl(form.content, f"{key}.content")
+
+    # Perform a "deep" copy without preserving references
+    form = awkward.forms.from_dict(form.to_dict())
+    impl(form, key)
+    return form
+
 
 class TrivialFormMapping(ImplementsFormMapping):
     def __call__(self, form: Form) -> tuple[Form, TrivialFormMappingInfo]:
-        dask_awkward = uproot.extras.dask_awkward()
-        new_form = dask_awkward.lib.utils.form_with_unique_keys(form, "<root>")
+        new_form = form_with_unique_keys(form, "<root>")
         return new_form, TrivialFormMappingInfo(new_form)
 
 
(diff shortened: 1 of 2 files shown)


ianna:
@pfackeldey - what is the plan for this PR? thanks!

ikrommyd:
> @pfackeldey - what is the plan for this PR? thanks!

I think we were discussing with Peter to add some caching support. Uproot will deserialize each electron branch for example separately with its own offsets. All those will have the same `count_branch` however so it's probably best to not deserialize the same offsets dozens of times. It's probably best to cache `count_branch` deserialization result (length) and use it for the other branches that have the same `count_branch`.

ariostas:
> I think we were discussing with Peter to add some caching support.

Isn't there already some caching being done in Uproot? When trying to read the count_branch multiple times it should already be hitting the cache

ikrommyd:
> > I think we were discussing with Peter to add some caching support.
> 
> Isn't there already some caching being done in Uproot? When trying to read the count_branch multiple times it should already be hitting the cache

Yeah I need to try if it's hitting it, haven't done that yet. Will do today.

ariostas:
> Do you the best way to log that (the number of deserializations per branch)?

I'm not sure. I've just skimmed the code since at some point I'll have to do that for RNTuples

pfackeldey:
> @pfackeldey - what is the plan for this PR? thanks!

I'm not sure. I'm not a big fan of this implementation, but I also don't know how it can be done in a better way. I was hoping for some input here.

pfackeldey:
Hi @ianna,
Finally, I found a good implementation! 
This is now handling every awkward Content case in a programmatic way. I could reuse a similar logic as for `uproot.dask`.
Could you have a look?

src/uproot/_dask.py
```diff
@@ -971,11 +971,111 @@ def load_buffers(
...
+
+        container = {}
+        for buffer_key, _ in self._form.expected_from_buffers().items():
+            container[buffer_key] = generator(tree, buffer_key)
+
+        return container
+
+
+def form_with_unique_keys(form: Form, key: str) -> Form:
+    awkward = uproot.extras.awkward()
```
ianna, line 1045: If this function is performance-sensitive or called multiple times, moving this import/lookup to the module scope would be better.
Or at least as a default parameter to avoid repeated calls.

ianna: oh, wait - apart from that line this function is a copy of `form_with_unique_keys`: https://github.com/dask-contrib/dask-awkward/blob/3440ce07c2c2e14657b1e48b42f4437b6d7e7849/src/dask_awkward/lib/utils.py#L140

Why?

ianna (changes requested):
@pfackeldey - Thanks! I would rather avoid duplicating the code. What was the reason for copying the `form_with_unique_keys` utility function here? Thanks.

src/uproot/_dask.py
pfackeldey: It's not performance sensitive, it's just called once per `.virtual_arrays()` call.

> oh, wait - apart from that line this function is a copy of form_with_unique_keys

Because there should not be a dependency on dask-awkward here. 
It's an unfortunate duplicate right now, but it would need a whole more rethinking and restructuring the whole form-manipulation logic in dask-awkward and uproot.dask. I'm happy to look into this (I planned to anyway), but I did not want to block this feature from being added for now.

ianna: I would assume the name `_dask` implies the dependency on `dask_awkward`. Besides, there is still other dependency in this module: https://github.com/scikit-hep/uproot5/blob/56eebf6fdd44f01b66d91b3c6acf346dcecf1f6b/src/uproot/_dask.py#L1228 

Would using the `utils` from `dask_awkward` block the new feature? Thanks.

pfackeldey: There's no dependency on dask-awkward: 
It's only imported when `uproot.extras.dask_awkward()` is called, and those are always inside a function and not at the top module level. The line you've linked is never called in `.virtual_arrays()`.

pfackeldey:
Needs: https://github.com/scikit-hep/awkward/pull/3482/

src/uproot/behaviors/TBranch.py
```diff
@@ -728,6 +728,143 @@ def show(
 
             stream.write(formatter.format(name, typename, interp).rstrip(" ") + "\n")
 
+    def virtual_arrays(
+        self,
+        *,
+        filter_name=no_filter,
```
ariostas, line 734: Do you think people will complain about not having expression support like the regular `arrays` method?

(I don't blame you for not wanting to do it. This is why I want to fix `formulate` so that I can use it for RNTuples.)

ariostas:
> I'd prefer doing this via requirements and not parsing versions in the code.

Also, I think it's fine since Awkward and Uproot are co-maintained, so if one can be upgraded, so can the other one. It's not like Awkward dropped support for Python 3.9, while Uproot still supported it or something like that.

src/uproot/behaviors/TBranch.py
pfackeldey: We can add it in the future once formulate if fixed I'd say. I'd argue it's better to restrict functionality and open it up later once it's in a working state.

ianna: To avoid API sprawl, we can introduce a single, unified API, like:
```python
def arrays(
    self,
    expressions=None,
    cut=None,
    *,
    filter_name=no_filter,
    filter_typename=no_filter,
    filter_branch=no_filter,
    aliases=None,
    language=uproot.language.python.python_language,
    entry_start=None,
    entry_stop=None,
    decompression_executor=None,
    interpretation_executor=None,
    array_cache="inherit",
    library="ak",
    ak_add_doc=False,
    how=None,
    virtual=False,     # <--- NEW
    access_log=None    # <--- NEW (only applies if virtual=True)
)
```

pfackeldey: that's a great idea! I think I tried this in the beginning, but for some reason decided for a new method... can't remember why.

I'd be in favor to make it work like you proposed @ianna

ariostas: That's true, but maybe at least trivial expressions? (which is probably what most people use anyway)

ikrommyd: If by trivial means just "which branches" then sure. In my opinion, that's the only selections that should be allowed with virtual arrays. Just "which branches" -style selections. No math.

ariostas: Also, for non-trivial expressions, maybe it would make sense to just materialize. I'm assuming that's what happens when you try to manipulate virtual arrays, so it would make sense to have the same behavior.

ariostas: I agree with not supporting expressions, but I think in that case having `virtual_arrays` would make more sense to make it clear that the api is a bit different.

pfackeldey: > I agree with not supporting expressions, but I think in that case having `virtual_arrays` would make more sense to make it clear that the api is a bit different.

that may have been my original motivation for a separate method. there may be other args as well the make sense for eager arrays and not virtual arrays and vice-versa. Maybe we should understand how many weird/invalid argument combinations there are to see if it is valid to move it into a different method.

ianna:
> @pfackeldey - thanks! Looks good to me. Please merge it if you’re done with it. Thanks

Oh, it’s still in draft mode 😅

Issue #1387, opened 2025-02-23
entry_stop not working with concatenation?


acampove:
I am using version `5.5.2`. The reproducer is:

```python
import uproot
import numpy as np

def _make_file(fname : str):
    n_entries = 1000
    branch1_data = np.random.rand(n_entries)
    branch2_data = np.random.rand(n_entries)

    with uproot.recreate(fname) as f:
        f["tree"] = {
            "a": branch1_data,
            "b": branch2_data   
        }

def main():
    _make_file('file_1.root')
    _make_file('file_2.root')

    df = uproot.concatenate({'file_1.root': 'tree', 'file_2.root' : 'tree'}, library='pd', entry_stop=100)

    print(len(df))

if __name__ == "__main__":
    main()
```

I expect the dataframe to have 200 entries but I see 2000. Am I doing anything stupid here?

pfackeldey:
Hi @acampove,
`entry_start` and `entry_stop` are not supported for `uproot.concatenate`. The question here is what should these entry ranges correspond to? 
- To the `entry_start` and `entry_stop` of each input file? Then we need to have a different type of argument that recognizes these _per tree_.
- To the `entry_start` and `entry_stop` of the concatenated file? That still requires us to concatenate the original trees and then slice them down. If you want this behavior you can always slice with `df[:200]`.

I agree though that `uproot.concatenate` should complain if it gets an unrecognized argument (the drawbacks of `**kwargs`...). 

Let me know what you think about this, maybe it could be helpful if you can describe more your specific application?

acampove:
Hi @pfackeldey 

Thanks for your reply. The `entry_stop` and `entry_start` should correspond to the full, concatenated tree. That is intuitive and I would do that. 

I need to run a quick local test involving code using this functionality. What you need is a line of code saying:

```python

for key in kwargs:
     if key not in self.implemented_args:
         raise NotImplementedError(f'Argument {key} not implemented')
```

and I would put that in a function, sot that you only call one line. 

Is there any way to achieve what `entry_stop` would? If not, maybe it's a good idea to implement it

Cheers.

PR #1396, opened 2025-02-28
feat: add entry_start and entry_stop to uproot.concatenate

pfackeldey:
Closes https://github.com/scikit-hep/uproot5/issues/1387.

@acampove can you give this a try?

### DIFF
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -241,6 +241,8 @@ def concatenate(
     filter_branch=no_filter,
     aliases=None,
     language=uproot.language.python.python_language,
+    entry_start=None,
+    entry_stop=None,
     decompression_executor=None,
     interpretation_executor=None,
     library="ak",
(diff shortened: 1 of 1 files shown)


src/uproot/behaviors/TBranch.py
```diff
@@ -359,37 +367,70 @@ def concatenate(
...
-                    else:
-                        raise
+        if _hasbranches is not None:
+            all_hasbranches.append(_hasbranches)
+
+    total_num_entries = sum(hasbranches.num_entries for hasbranches in all_hasbranches)
+    entry_start, entry_stop = _regularize_entries_start_stop(
+        total_num_entries, entry_start, entry_stop
+    )
+    for hasbranches in all_hasbranches:
```
ianna, line 384: @pfackeldey - would it be a performance penalty to have the second loop?

ianna (commented):
@pfackeldey - Thanks for looking into it! Let's wait for the user feedback.

src/uproot/behaviors/TBranch.py
pfackeldey: no, unless the number of given files<->tree is _very_ large - I doubt someone would ever try to concatenate e.g. >1M trees.

acampove: No, we won't do that. The largest number of trees is going to be in the 10K-20K. But we would just merge the files when the number goes beyond a few hundred.

Thanks for the work BTW.

PR #1401, opened 2025-03-06
ci: switch to micromamba

### LINKED ISSUE #1391: CI is being fairly unreliable
Recently, the CI is being fairly flaky. Usually, at least one of the test jobs randomly fails. The issue is with the `conda-incubator/setup-miniconda` action, as it randomly fails to set up the environment properly. It seems like other people are having the same issue here: https://github.com/conda-incubator/setup-miniconda/issues/392 . They mention that explicitly specifying the mamba version fixes the issue, so I think we should just do that if this is not fixed within a couple weeks.

ariostas:
I'm pinning the Mamba version that the CI uses so that it stops being so unreliable. The issue was reported in conda-incubator/setup-miniconda#392 but it's been a while and it hasn't been fixed. The suggestion they have there is to pin the Mamba version, so that's what I'm doing. I'll run the tests a few times to make sure that this fix works and then I'll mark it as ready for review.

Closes #1391

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -36,7 +36,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
           miniforge-version: latest
           use-mamba: true
-          mamba-version: 2.0.5 # FIXME: Unpin when conda-incubator/setup-miniconda#392 is fixed
+          mamba-version: 2.0.4 # FIXME: Unpin when conda-incubator/setup-miniconda#392 is fixed
 
       - name: Check active Python version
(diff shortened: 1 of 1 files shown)


ariostas:
I'm not really sure why this is not working. It seems like up to 2.0.5 it's broken on Windows, and 2.0.6 should be fixed (at least from reading the release notes). However, it's not able to find 2.0.6 even though it's [on conda-forge](https://anaconda.org/conda-forge/mamba).

ariostas:
This is ready for review now. I ended up switching to micromamba and it seems to be stable now. I ran the tests a few times and there were no issues with mamba (there were a couple of unrelated issues, probably from me rerunning tests so much).

henryiii (approved):

ianna (approved):
@ariostas - Great! Looks good to me. Thanks.

PR #1406, opened 2025-03-17
feat: RNTuple reading improvements

ariostas:
This is still a work in progress. I'm adding more functionality for RNTuple reading so that methods are on par with what is available for TTrees. In particular, I'm adding missing functionality to `keys`, `arrays`, and similar methods. There's also other minor tweaks and fixes, and a lot more documentation.

I introduced a new `HasFields` class that plays a similar role for `RNTuple` and `RField`, as `HasBranches` does for `TTree` and `TBranch`. I'm basing things on the existing code for `TBranch`. However, since `RNTuple` is more "fundamental" than `RField`, and `RField` is a "fake" class that is not read directly, I'm structuring things a bit differently.


ariostas:
I ended up not adding all the functionality, since I didn't realize how robust those functions are for TTrees. However, I added all the most commonly used functionality, and I'll follow up implementing missing things on future PRs.

There were some significant changes to `keys` and related methods. Now nothing is hidden from the user, so they have access to the full RNTuple structure, even though some subfields are not really useful. You can also now see the full structure with `.show()`.

I also fixed the tupleness of tuple-like data (`std::pair` and `std::tuple`), and I added some more documentation.

ariostas:
Just out of curiosity, I'll have Copilot review this PR to see how it works.

ianna (approved):
@ariostas - Thanks! Looks good to me. The copilot review is quite detailed :-)
I'm merging it.

PR #1413, opened 2025-04-07
fix: compress RNTuple column data

ariostas:
RNTuple data was being stored fully uncompressed, which was pretty wasteful. This small PR makes it so that it is stored compressed with the compression setting from the file. It would be good to have the option of selecting compression settings in a more granular way (i.e. per field or column), but I'll have to think a bit more how to implement this since it would make more sense to specify it per column, but it would be a lot easier to specify it per field.

### DIFF
--- a/src/uproot/writing/_cascadentuple.py
+++ b/src/uproot/writing/_cascadentuple.py
@@ -964,7 +964,12 @@ class NTuple(CascadeNode):
             raw_data = col_data.reshape(-1).view("uint8")
             if col_data.dtype == numpy.dtype("bool"):
                 raw_data = numpy.packbits(raw_data, bitorder="little")
-            page_key = self.add_rblob(sink, raw_data, len(raw_data))
+            uncompressed_bytes = len(raw_data)
+            # Need better logic to specify per-column/field compression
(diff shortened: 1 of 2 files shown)


src/uproot/writing/_cascadentuple.py
```diff
@@ -661,7 +645,7 @@ def serialize(self):
                                         col[0][2]
                                     ),
                                     _rntuple_column_compression_settings_format.pack(
-                                        self.compression_settings
+                                        col[0][3]
```
ianna, line 648: why is it so specific?

ianna (approved):
@ariostas - looks good to me. I guess, I don't understand the RNTuple format good enough to tell if specific encoding is what one would expect. Please, merge it when you are finished. Thanks!

src/uproot/writing/_cascadentuple.py
ariostas: Yeah, this is pretty ugly. It's one of the things I'll have to refactor since right now it's hard to know what it's doing. But in the interest of getting this PR out since it's kind of important, I'll defer this to the big writing refactoring that I'll start doing soon. Thanks, Ianna!

PR #1415, opened 2025-04-07
chore: better error message when writing TBaskets that are too large

ariostas:
In #1135 there are some discussion about writing batches of data that are too big to fit in a TBasket. I added more detailed error messages so that it is clear what went wrong, since the way it currently crashes is not very informative.

### DIFF
--- a/src/uproot/writing/_cascadetree.py
+++ b/src/uproot/writing/_cascadetree.py
@@ -1395,6 +1395,12 @@ class Tree:
         fObjlen = len(uncompressed_data)
         fNbytes = fKeylen + len(compressed_data)
 
+        if fObjlen > 2**31 - 1 or fNbytes > 2**31 - 1:
+            raise ValueError(
+                f"Numpy array data of branch {branch_name} has an uncompressed size of {fObjlen} bytes "
+                f"and a compressed size of {fNbytes} bytes, which is too large to fit in a TBasket"
+            )
+
         parent_location = self._directory.key.location  # FIXME: is this correct?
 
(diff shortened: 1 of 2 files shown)


ariostas:
@ikrommyd what do you think of these messages?

ikrommyd:
> @ikrommyd what do you think of these messages?

I like them, I would probably also mention that the max size is 2GB if that's always the hard limit. I'm not aware of root file internals so I'm not sure if it's 100% correct but if it is, it's useful so that the user knows what they need to at least reduce the size to.

src/uproot/writing/_cascadetree.py
```diff
@@ -1395,6 +1395,13 @@ def write_np_basket(self, sink, branch_name, compression, array):
         fObjlen = len(uncompressed_data)
         fNbytes = fKeylen + len(compressed_data)
 
+        if fObjlen > 2**31 - 1 or fNbytes > 2**31 - 1:
```
ianna, line 1398: would it be better to define something like `MAX_INT32 = numpy.iinfo(np.int32).max`?

ianna (changes requested):
@ariostas - I think, it's better to define a constant to rely on. I feel the calculations can be confusing.

src/uproot/writing/_cascadetree.py
ariostas: Yeah, that's definitely a lot better. Thanks, Ianna!

ianna (approved):
@ariostas - Thanks! Looks great, please, merge it if you finished with it. Thanks!

PR #1416, opened 2025-04-09
feat: support writing ListArrays to RNTuples

ariostas:
This PR adds support for writing `ListArray`s to RNTuples. RNTuples have an analogue for this, so they are converted to `ListOffsetArray`s before writing them.

There are still some array types that can't be written, but each of them requires special handling since again there are no analogues in RNTuple. I'll continue to add them over time.

### DIFF
--- a/src/uproot/writing/_cascadentuple.py
+++ b/src/uproot/writing/_cascadentuple.py
@@ -90,7 +90,7 @@ def _cpp_typename(akform, subcall=False):
         typename = _ak_primitive_to_typename_dict[ak_primitive]
         for arr_size in inner_shape[::-1]:
             typename = f"std::array<{typename},{arr_size}>"
-    elif isinstance(akform, awkward.forms.ListOffsetForm):
+    elif isinstance(akform, (awkward.forms.ListOffsetForm, awkward.forms.ListForm)):
         content_typename = _cpp_typename(akform.content, subcall=True)
         typename = f"std::vector<{content_typename}>"
         override_typename = akform.parameters.get("__array__", "")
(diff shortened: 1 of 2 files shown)


src/uproot/writing/_cascadentuple.py
```diff
@@ -921,6 +952,24 @@ def extend(self, file, sink, data):
...
+                stops = awkward.index.Index(data_buffers[f"{barekey}-stops"])
+                next_barekey = f"node{int(barekey[4:])+1}"
+                content = awkward.contents.NumpyArray(
+                    data_buffers[f"{next_barekey}-data"]
+                )
+                tmp_buffers = awkward.to_buffers(
+                    awkward.contents.ListArray(
+                        starts, stops, content
+                    ).to_ListOffsetArray64()
+                )[2]
```
ariostas, line 968: @ianna do you know of a better way to convert ListArrays to ListOffsetArrays?

ianna: @ariostas - It looks like you have a very simple structure here: a vector of vectors? I think, you could avoid copy to the temporary buffers and back if you know that your `data_buffers` are contiguous. Yes, `to_ListOffsetArray64` is convenient and will take care of it and run more checks, but it will also convert the `Index`'s buffers to `int64`. Would something like this do the job?
```python
offsets = [starts[0], *stops]
```

ianna (approved):
@ariostas - It is a correct way to convert `ListArray` to `ListOffsetArray`, though it may not be the most efficient. I'll leave it up to you to decide if you want to leave it as is since it is the internal code, and you may want to optimize it later.

src/uproot/writing/_cascadentuple.py
ariostas: Thank you, Iason and Ianna!

@ianna transforming to a `ListOffsetArray` is not always straightforward. Here's an example (which is the one I included in the test)
```
>>> import awkward as ak
>>> a = ak.contents.ListArray(
...     ak.index.Index([1, 0, 3]),
...     ak.index.Index([1, 2, 4]),
...     ak.contents.NumpyArray([0, 1, 2, 3, 4, 5]),
... )
>>> print(a)
<ListArray len='3'>
    <starts><Index dtype='int64' len='3'>[1 0 3]</Index></starts>
    <stops><Index dtype='int64' len='3'>[1 2 4]</Index></stops>
    <content><NumpyArray dtype='int64' len='6'>[0 1 2 3 4 5]</NumpyArray></content>
</ListArray>
>>> print(a.to_ListOffsetArray64())
<ListOffsetArray len='3'>
    <offsets><Index dtype='int64' len='4'>[0 0 2 3]</Index></offsets>
    <content><NumpyArray dtype='int64' len='3'>[0 1 3]</NumpyArray></content>
</ListOffsetArray>
```
So you can see that the offsets and even the data buffer are changed non-trivially.

ariostas: Later on I'll look to see if there is a more straightforward way to do this with an internal function in Awkward that might avoid some copying around, but at least it's good to know that this is not terrible.

PR #1421, opened 2025-04-18
fix: safer branch title access

pfackeldey:
This happens when writing a length-0 coffea NanoEvents array with uproot and then reading it again (reported on coffea Mattermost channel). uproot wants to add the branch title as the `__doc__` parameter to the awkward array, but fails if it isn't present. This PR looses this requirement a bit and adds an empty string instead of a failure.

### DIFF
--- a/src/uproot/interpretation/library.py
+++ b/src/uproot/interpretation/library.py
@@ -460,7 +460,7 @@ def _awkward_json_to_array(awkward, form, array):
 
 def _awkward_add_doc(awkward, array, branch, ak_add_doc):
     if ak_add_doc:
-        return awkward.with_parameter(array, "__doc__", branch.title)
+        return awkward.with_parameter(array, "__doc__", getattr(branch, "title", ""))
     else:
         return array
 


ianna (changes requested):
@pfackeldey - Good catch! Thanks. The previous code assumes branch definitely has a title attribute. If not, it will raise an `AttributeError`.

I think it is better to add something meaningful instead of an empty string, especially for documentation or metadata like `__doc__`.

Here’s why:

- An empty string doesn’t help users or tools that rely on metadata to understand or display the data.
-  It can make debugging or documentation harder.
-  It doesn’t signal why the title is missing (was it forgotten, unavailable, not relevant?).

Please, consider an alternative. For example, if it can handle `None`:
```python
getattr(branch, "title", None)
```
if not, something like:
```python
getattr(branch, "title", "no title")
```

Thanks!

pfackeldey:
:+1: I changed it to `None`.

Initially I didn't go for:
```python
getattr(branch, "title", None)
```
because that's a change of types (it's not guaranteed to be a str anymore).

and I don't like:
```python
getattr(branch, "title", "no title")
```
because if a `branch.title` is in fact exactly "no title" there's a double meaning behind this string, which is problematic. That's always a problem of default strings, and I think they should always be avoided if possible.

I was also considering to only add a parameter if the title attribute exists.

In the end I decided to go for `None` because it's the least invasive of all these changes and likely has the least consequences for downstream libraries or users.

Hope that change does not affect anyone and looks good now :+1:

ianna (approved):
@pfackeldey - Thanks!

PR #1420, opened 2025-04-17
fix: issue with empty big_endian array

pfackeldey:
Fixes problem with 0-length arrays and writing, e.g.:

```python
import uproot

events = uproot.dask({"coffea/tests/samples/nano_dy.root": "Events"})

uproot.dask_write(
  events[events.nJet > 10],  # cuts away all events
  destination="skimtest/",
  prefix="nano_dy/skimmed",
  compute=True,
)
```

PR #1422, opened 2025-04-18
ci: pin Chrome version for Pyodide tests

ariostas:
I'm updating the version of Pyodide that the CI uses to see if it fixes the issue we started to see in #1420.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -175,8 +175,8 @@ jobs:
         uses: pyodide/pyodide-actions/install-browser@v2
         with:
           runner: selenium
-          browser: node
-          browser-version: 20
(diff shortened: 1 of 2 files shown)


ariostas:
It turns out that the issue was that an update to Chrome broke something. I tried switching to using Node.js, but the http tests don't work there. For now I just pinned the Chrome version, and later I'll think about whether it's worth to continue having http tests, since they are pretty finicky and in a real deployment it's tricky to access files through http due to CORS protections.

ianna (approved):
@ariostas - thanks! It seems weird that we have to pin the browser version. It is ok for now, I think. 

@henryiii - could you please have a look? Thanks!

henryiii (approved):

PR #1426, opened 2025-04-22
feat: Optimized Reading of RNTuple Data from Storage into GPU Memory via CuFile GDS or POSIX

fstrug:
This PR adds support for reading RNTuple data from storage directly to GPU memory via RDMA on GPU Direct Storage (GDS) enabled systems. On systems without GDS, cufile runs in compatibility mode and reads are performed by the cpu via POSIX. Currently, there is only support for reading RNTuple datas compressed with the zstandard algorithm.

Changes are made to `uproot.behaviors.RNTuple.HasFields.arrays()` to control steering of arrays function with new optional arguments `backend` and `use_GDS`. 

There are known issues with nvcomp's implementation of the zstandard algorithm which crashes nvcomp (and uproot) when decompressing certain buffers. This behavior is non-determinant and the reason is unknown. Fixes are expected in the [next release of nvcomp]( https://github.com/rapidsai/kvikio/issues/673).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,18 @@ readme = "README.md"
 requires-python = ">=3.9"
 
 [project.optional-dependencies]
+GDS_cu11 = [
+  "kvikio-cu11>=25.02.01",
+  "dataclasses",
+  "functools",
+  "operator"
+]
+GDS_cu12 = [
+  "kvikio-cu12>=25.02.01",
+  "dataclasses",
+  "functools",
+  "operator"
+]
 dev = [
   "boost_histogram>=0.13",
   "dask-awkward>=2025.2.0",
@@ -87,18 +99,6 @@ test-pyodide = [
   "scikit-hep-testdata"
 ]
 xrootd = ["fsspec-xrootd>=0.5.0"]
-GDS_cu11 = [
-    "kvikio-cu11>=25.02.01",
-    "dataclasses",
-    "functools",
-    "operator",
-]
-GDS_cu12 = [
-    "kvikio-cu12>=25.02.01",
-    "dataclasses",
-    "functools",
-    "operator",
-]
 
 [project.urls]
 Download = "https://github.com/scikit-hep/uproot5/releases"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,6 +87,18 @@ test-pyodide = [
   "scikit-hep-testdata"
 ]
 xrootd = ["fsspec-xrootd>=0.5.0"]
+GDS_cu11 = [
+    "kvikio-cu11>=25.02.01",
+    "dataclasses",
+    "functools",
+    "operator",
+]
+GDS_cu12 = [
+    "kvikio-cu12>=25.02.01",
+    "dataclasses",
+    "functools",
+    "operator",
+]
 
 [project.urls]
 Download = "https://github.com/scikit-hep/uproot5/releases"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,15 +57,9 @@ requires-python = ">=3.9"
 [project.optional-dependencies]
 GDS_cu11 = [
   "kvikio-cu11>=25.02.01",
-  "dataclasses",
-  "functools",
-  "operator"
 ]
 GDS_cu12 = [
   "kvikio-cu12>=25.02.01",
-  "dataclasses",
-  "functools",
-  "operator"
 ]
 dev = [
   "boost_histogram>=0.13",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,42 @@ requires = [
   "hatch-vcs"
 ]
 
+[dependency-groups]
+dev = [
+  {include-group = "extra"},
+  {include-group = "test"}
+]
+extra = [
+  "boost_histogram>=0.13",
+  "dask-awkward>=2025.2.0",
+  "dask[array,distributed]",
+  "hist>=1.2",
+  "pandas",
+  "awkward-pandas"
+]
+test = [
+  "aiohttp",
+  "deflate",
+  "fsspec-xrootd>=0.5.0",
+  "isal",
+  "minio",
+  "paramiko",
+  "pytest-rerunfailures",
+  "rangehttpserver",
+  "requests",
+  "s3fs",
+  {include-group = "test-core"}
+]
+test-core = [
+  "pytest-timeout",
+  "pytest>=6",
+  "scikit-hep-testdata"
+]
+test-pyodide = [
+  "pytest-pyodide; python_version >='3.11'",
+  {include-group = "test-core"}
+]
+
 [lint.mccabe]
 max-complexity = 100
 
@@ -43,7 +79,7 @@ dependencies = [
   "numpy",
   "fsspec",
   "packaging",
-  "typing_extensions>=4.1.0; python_version < \"3.11\""
+  "typing_extensions>=4.1.0; python_version < '3.11'"
 ]
 description = "ROOT I/O in pure Python and NumPy."
 dynamic = [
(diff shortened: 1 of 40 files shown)


pyproject.toml
```diff
@@ -55,6 +55,18 @@ readme = "README.md"
 requires-python = ">=3.9"
 
 [project.optional-dependencies]
+GDS_cu11 = [
+  "kvikio-cu11>=25.02.01",
+  "dataclasses",
+  "functools",
+  "operator"
```
nsmith-, line 62: These 3 are modules in the python standard library. You don't have to list them as dependencies

src/uproot/behaviors/RNTuple.py
```diff
@@ -24,6 +24,12 @@
 import uproot.source.chunk
 from uproot._util import no_filter, unset
 
+# GDS Depdencies
+try:
+    import cupy as cp
+except ImportError:
+    pass
```
nsmith-, line 31: Move the import into the `_arrays_GDS` function. Then, you can catch the ImportError exception and raise a new error to the user instructing them to `python3 -m pip install uproot[GDS_cu11]` or _cu12.

src/uproot/behaviors/RNTuple.py
```diff
@@ -725,6 +871,107 @@ def arrays(
...
+
+        #####
+        # Read and decompress all columns' data
+        clusters_datas = self.GPU_read_clusters(
+            target_cols, start_cluster_idx, stop_cluster_idx
+        )
+        clusters_datas.decompress()
+        #####
+        # Deserialize decompressed datas
+        content_dict = self.Deserialize_decompressed_content(
```
nsmith-, line 923: Rename this function to use [snake case](https://en.wikipedia.org/wiki/Snake_case) to be consistent with the other functions in the library.

src/uproot/models/RNTuple.py
```diff
@@ -16,6 +16,17 @@
 import uproot.behaviors.RNTuple
 import uproot.const
 
+try:
+    import functools
+    import operator
+    from dataclasses import dataclass, field
```
nsmith-, line 22: These 3 don't need to be guarded by the try

src/uproot/models/RNTuple.py
```diff
@@ -16,6 +16,17 @@
...
+
+    import cupy as cp
+    from kvikio import CuFile, defaults
+    from kvikio.nvcomp_codec import NvCompBatchCodec
```
nsmith-, line 26: Each function that uses cupy or kvikio should call a function like `cupy = uproot.extras.cupy()` etc. to import the necessary module. This follows the pattern seen in https://github.com/scikit-hep/uproot5/blob/main/src/uproot/extras.py

nsmith- (changes requested):
Some initial comments. I will look more closely at the implementation later.

src/uproot/models/RNTuple.py
fstrug: We cannot avoid `from kvikio.nvcomp_codec import NvCompBatchCodec` since `nvcomp_codec` is not available in https://github.com/rapidsai/kvikio/blob/branch-25.06/python/kvikio/kvikio/__init__.py.

fstrug: Instead defining `uproot.extras.kvikio_nvcomp_codec()`.

fstrug:
Passing all RNTuple reading tests besides two.

`test_1411_rntuple_physlite_ATLAS.py::test_truth_muon_containers` - Bug found in awkward array fixed at https://github.com/scikit-hep/awkward/pull/3507. Had issues building `awkward-cpp` so I haven't verified with awkward `main` branch that tests passes as I got missing awkward kernel error at final `assert`.
`test_1250_rntuple_improvements.py::test_iterate` - failing at line 89. Iterate not yet implemented for GDS. Failing for `backend = "cpu" and `use_GDS = False`. Only changed function to directly call `HasFields._arrays()` instead of high level `HasFields.arrays(). This seems to be unrelated to this PR right now.

I am still unsure how we want to solve the following tests.
`test_0662_rntuple_stl_containers` - `Cupy` does not support `dtype = string`. Cudf supports string manipulations and support in `cupy` not anticipated soon. https://github.com/cupy/cupy/issues/8698#issuecomment-2439930580 Unsure how we might want to implement this.
`test_1223_empty_struct.py::test_invalid_variant()` - `Cupy` does not support `dtype = object`. It is possible to call `ak.to_backend(a, "cuda")` which works. I need to find a way to build `a.variant` with backend `cuda` via `cupy` arrays and `ak.from_buffers()`.

pyproject.toml
```diff
@@ -91,6 +91,20 @@ readme = "README.md"
 requires-python = ">=3.9"
 
 [project.optional-dependencies]
+GDS_cu11 = [
+  "kvikio-cu11>=25.02.01"
+]
+GDS_cu12 = [
+  "kvikio-cu12>=25.02.01"
+]
+dev = [
```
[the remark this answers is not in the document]
fstrug: Did we resolve this issue now by adding the GDS dependencies to `test` group?

ariostas (commented):
Like Nick said, seems like this is pretty close to being done. I just left a few minor comments.

tests-cuda/test_0630_rntuple_basics.py
```diff
@@ -0,0 +1,70 @@
...
+    ("backend", "GDS", "library"),
+    [
+        ("cuda", False, cupy),
+        pytest.param(
+            "cuda",
+            True,
+            cupy,
+            marks=pytest.mark.skipif(
+                cupy is None, reason="could not import 'cupy': No module named 'cupy'"
+            ),
```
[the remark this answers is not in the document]
ariostas: Ah okay, so in that case why not just skip the whole test file if `cupy` could not be imported?

src/uproot/models/RNTuple.py
```diff
@@ -737,6 +748,390 @@ def read_col_page(self, ncol, cluster_i):
...
+                content = library.copy(destination)
+                content = _extract_bits(content, nbits)
+            if dtype_str == "real32trunc":
+                content <<= 32 - nbits
+
+        # needed to chop off extra bits incase we used `unpackbits`
+        destination[:] = content[:num_elements]
+
+
+def _extract_bits(packed, nbits):
```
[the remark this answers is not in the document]
fstrug: I'm close to refactoring the CPU code to use `deserialize_page_decompressed_buffer()` entirely as said above in a reply to one of Nick's comment. I just need to find out what change is needed to make it work for truncated floats.

src/uproot/models/RNTuple.py
```diff
@@ -737,6 +748,390 @@ def read_col_page(self, ncol, cluster_i):
...
+        columns. Each ColBuffersCluster contains the page buffers, decompression
+        target buffers, and compression metadata for a column in a given cluster.
+
+        The ClusterRefs object contains all information needed for and performs
+        decompression in parallel over all compressed buffers.
+        """
+        cluster_range = range(start_cluster_idx, stop_cluster_idx)
+        clusters_datas = ClusterRefs()
+        # Open filehandle and read columns for clusters
+        filehandle = CuFileSource(self.file.source.file_path, "rb")
```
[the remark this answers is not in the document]
fstrug: CuFile provides GDS supports when reading from local and network attached storage via DMA and RDMA respectively given the storage has been configured for GDS at the system level. The `nvidia-fs` driver is the engine which handles memory mapping between storage and GPU memory for configured devices.

nsmith- (approved):
Looks good to me
We will probably have some more work to do once the GPU tests are set up. We should also look into making example files for testing all supported compression algorithms but that can be in a later PR.

ariostas (approved):
It also looks good to me. Thank you, @fstrug!

ianna (approved):
@fstrug - This looks good! Though I'm getting a future warning from `kvikio 25.06.00` about `nvcomp_codec` deprecation, I think, the PR should be merged. Please, squash and merge it if you are done with it. Thanks!

fstrug:
@ianna Yes, `kvikio` recently deprecated their nvcomp bindings since there is a separately maintained official `nvcomp` package. https://docs.nvidia.com/cuda/nvcomp/py_api.html

The reason I haven't moved to the official `nvcomp` package is because it doesn't currently support user-provided target buffers. I have raised this issue, and NVIDIA has it internally documented and targeted for a future release. https://forums.developer.nvidia.com/t/please-support-user-provided-output-buffers-in-python-api/332763

I am finished with this PR, so it can be squashed and merged.

ianna:
@eduardo-rodrigues - yes, we should make sure that this is well documented and presented to the community. Perhaps as a PyHEP topical meeting to start with, and eventually as a part of a Scikit-HEP tutorial. @fstrug has done a fantastic job! And this is just the beginning of the integration. It will be available in the next uproot release later this week.

PR #1427, opened 2025-04-24
chore: move to using dependency-groups

henryiii:
Using more uv (faster), switching to dependency groups. `uv run ...` now works.

Includes restriction on dask due to https://github.com/dask/dask/pull/11859.


ianna (approved):
@henryiii - Great - thank you! I'm merging it. Thanks.

PR #1431, opened 2025-05-02
refactor: RNTuple writing improvements

ariostas:
This is still a work in progress. I'll update the description with the highlights once I'm done.


ariostas:
I think this is all the refactoring that makes sense for now. Once we start integrating `rootfilespec` we can make things much nicer. I just have a question about `ak_add_doc`, but I asked it on Slack.

ianna (approved):
@ariostas - This looks great! Thank you. Please, merge it when you are done with it. Thanks!

PR #1433, opened 2025-05-07
chore(deps): unconstrain Dask dependency

ariostas:
The version of Dask was constrained in #1427 since we thought that they introduced a bug. However, the issue seems to be with the interface between `dask-awkward` and `dask`, so they are taking care of it on that side (see https://github.com/dask-contrib/dask-awkward/pull/588).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dev = [
 extra = [
   "boost_histogram>=0.13",
   "dask-awkward>=2025.2.0",
-  "dask[array,distributed] !=2025.4.0",
+  "dask[array,distributed]",
   "hist>=1.2",
   "pandas",
   "awkward-pandas"


ariostas:
Since it's a very minor change, I'll merge it without a review.

pyproject.toml
```diff
@@ -13,7 +13,7 @@ dev = [
 extra = [
   "boost_histogram>=0.13",
   "dask-awkward>=2025.2.0",
-  "dask[array,distributed] !=2025.4.0",
+  "dask[array,distributed]",
```
ianna, line 16: @ariostas - does it mean that it works with `2025.4.0`?

ianna: it looks like the CI on `main` is not happy with this change...

ikrommyd: 2025.4.1 has the same problem. I think it should be an upper pin for now like we did with dask-awkward `<2025.4.0`

ariostas: > it looks like the CI on main is not happy with this change...

The CI on main was fine. I'm not sure if I missed something.

> 2025.4.1 has the same problems. I think it should be an upper pin for now like we did with dask-awkward <2025.4.0

Since the incompatibility is with `dask-awkward`, it should be pinned there, not here. `uproot.dask(..., library="np")` works with `2025.4.*`, so nothing in Uproot itself is incompatible with `2025.4.*`

pfackeldey: instead you could also just pin `dask-awkward>=2025.5.0` ? (that should resolve it properly)

pfackeldey: dask-awkward v2025.5.0 does pin dask to be smaller than 2025.4.*

ariostas: > instead you could also just pin dask-awkward>=2025.5.0 ?

Yeah, I guess that's the way to go

Sorry for not waiting for a review. This turned out to be a lot more controversial than I expected. I just thought I should remove the constraint before we forgot about it (and since it was faulty now that 2025.4.1 was out)

PR #1436, opened 2025-05-14
fix: Only dispatch requests to Source if we actually will use them

nsmith-:
The notifications queue in _ranges_or_baskets_to_arrays cannot be aware of chunk requests in flight because the queue only gets pushed to when we receive the returned chunk data. If we have the necessary array in cache, then we exit this function and fire-and-forget chunk requests causing extra network traffic, and more to the point, potential race conditions between receiving chunks and closing the file on teardown.

Fixes https://github.com/scikit-hep/coffea/issues/1331


ikrommyd:
@ianna this is critical even for iris-hep training.

ianna (approved):
@nsmith- Looks great! Thanks!

PR #1437, opened 2025-05-15
fix: Remove unused file handle

nsmith-:
FSSpecSource uses an entirely stateless interface for the actual reading, so this file handle is now an unused remnant.


ianna (approved):
@nsmith- Thanks! As discussed at the meeting I'm merging it.

PR #1438, opened 2025-05-15
chore: update pin on dask-awkward

ianna:


ianna (commented):
as discussed at the meeting we need to move to 2025.5.0

ariostas (approved):

PR #1448, opened 2025-06-11
perf: faster reading of files with many branches

### LINKED ISSUE #1442: Poor performance reading file with many branches
We are using ROOT files with on the order of 10k branches that contain one value each with mixed data types. We have been using uproot version 3.14.4 where reading of these files into pandas dataframes takes only a couple of seconds. 
We are transitioning now to newer uproot versions and have found that reading these files takes significantly longer. 

Here are some examples: 
```python
import uproot3
import uproot 
import numpy as np
import pandas as pd

num_rows = 1
num_cols_df1 = 10000
num_cols_df2 = 10000

df1 = pd.DataFrame(
    np.random.randint(0, 100, size=(num_rows, num_cols_df1)),
    columns=[f"A{i}" for i in range(num_cols_df1)]
)

df2 = pd.DataFrame(
    np.random.choice(['foo', 'bar', 'baz'], size=(num_rows, num_cols_df2)),
    columns=[f"X{i}" for i in range(num_cols_df2)]
)
df = pd.concat([df1, df2], axis=1)

with uproot.recreate("test.root") as f:
    f["df"] = df
``` 

```python
%timeit uproot3.open("test.root")['df'].pandas.df()
13.7 s ± 453 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
```python
uproot.open("test.root:df").arrays(library="pd")
--> 49min 39s for one iteration  
```

Passing the branch names when calling `.arrays()` gives a small speed boost. Also, first storing the data using numpy and then converting to pandas is a tiny bit faster. 
```python
def testCase(file):
    c = uproot.open(file)
    branches = list(c['df'].keys())
    arrays = c['df'].arrays(branches, library="pd")
    return arrays

def testCaseNP(file):
    c = uproot.open(file)
    branches = list(c['df'].keys())
    arrays = c['df'].arrays(branches, library="np")
    d = pd.DataFrame(data=arrays, columns=branches)
    return d
```

```python
%timeit testCase("test.root")
1min 4s ± 1.71 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit testCaseNP("test.root")
1min ± 421 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Is there a way to restore the speed of uproot3 in this case?

Tested with uproot 3.14.4 and 5.6.1 on python 3.10.6

ariostas:
This PR fixes #1442 (at least partially). It's not clear to me why this getter is needed, but it has a huge performance cost, so I switched to using the branch names directly.

This change gives it a huge boost in performance for trees with many branches, although it is still significantly slower than Uproot3.

For comparison, I generated a file with 1000 branches, as done in #1442 and the output of
```python
%timeit uproot.open("test.root:df").arrays(library="pd")
```
went from
```
3.07 s ± 46.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
to
```
353 ms ± 7.82 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

@ianna do you know why this getter could have been needed in the first place? All the tests seem to pass, so I'm not sure why it was needed.

### DIFF
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -17,6 +17,7 @@ import re
 import sys
 import threading
 from collections.abc import Iterable, Mapping, MutableMapping
+from keyword import iskeyword
 
 import numpy
 
@@ -2923,7 +2924,7 @@ def _regularize_expressions(
             ):
                 branchname_expression = (
                     branchname
-                    if branchname.isalnum()
+                    if branchname.isidentifier() and not iskeyword(branchname)
                     else language.getter_of(branchname)
                 )
                 _regularize_expression(
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -2923,7 +2923,7 @@ def _regularize_expressions(
             ):
                 _regularize_expression(
                     hasbranches,
-                    language.getter_of(branchname),
+                    branchname,
                     keys,
                     aliases,
                     language,
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -1965,7 +1965,9 @@ class TBranch(HasBranches):
         """
         if self._cache_key is None:
             sep = ":" if isinstance(self._parent, uproot.behaviors.TTree.TTree) else "/"
-            self._cache_key = f"{self.parent.cache_key}{sep}{self.name}({self.index})"
+            self._cache_key = (
+                f"{self.parent.cache_key}{sep}{self.name}({self.member('fOffset')})"
+            )
         return self._cache_key
 
     @property
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -1978,6 +1978,7 @@ class TBranch(HasBranches):
         non-recursive index is always unique.
         """
         if not hasattr(self, "_index"):
+            # cache index of all branches of the parent to avoid repeating this loop for other branches
             for i, branch in enumerate(self.parent.branches):
                 branch._index = i
         return self._index
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -1965,9 +1965,7 @@ class TBranch(HasBranches):
         """
         if self._cache_key is None:
             sep = ":" if isinstance(self._parent, uproot.behaviors.TTree.TTree) else "/"
-            self._cache_key = (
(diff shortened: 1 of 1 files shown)


ikrommyd:
Ci is failing due to numpy 2.3 and awkward. Perhaps we need a quick awkward patch release.

ariostas:
I just found out that another big improvement could be made by changing the way `cache_key` works. Currently, it uses the index, which is very slow to find for TTrees with many branches, so we should think of a better alternative.

https://github.com/scikit-hep/uproot5/blob/61835f10456541965ec7fc6ba37ddb07ecb5cfe9/src/uproot/behaviors/TBranch.py#L1960-L1984

ariostas:
After the last commit it's actually back to the speed of Uproot3!

ianna (approved):
@ariostas - looks great! Thanks for fixing it!

src/uproot/behaviors/TBranch.py
```diff
@@ -1977,11 +1977,10 @@ def index(self):
         :ref:`uproot.behaviors.TBranch.TBranch.name` is not unique: the
         non-recursive index is always unique.
         """
-        for i, branch in enumerate(self.parent.branches):
-            if branch is self:
-                return i
-        else:
-            raise AssertionError
+        if not hasattr(self, "_index"):
+            for i, branch in enumerate(self.parent.branches):
```
nsmith-, line 1981: Isn't this missing `if branch is self`?

nsmith- (changes requested):

src/uproot/behaviors/TBranch.py
ariostas: No, the point was to add the indices to all branches in one go instead of just to self

nsmith-: oh sorry I see now, makes sense

nsmith-: might be worth a comment # cache all branch indices for performance reasons

src/uproot/behaviors/TBranch.py
```diff
@@ -2921,9 +2921,14 @@ def _regularize_expressions(
                     uproot.interpretation.grouped.AsGrouped,
                 ),
             ):
+                branchname_expression = (
+                    branchname
+                    if branchname.isalnum()
+                    else language.getter_of(branchname)
+                )
```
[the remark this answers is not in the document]
ariostas: Oh after a quick search, turns out we can just use `isidentifier()`

ariostas:
This is ready for review again, but there's a bunch of tests failing for an unrelated reason.

PR #1464, opened 2025-07-11
ci: updates and fixes to Pyodide tests

ariostas:
I made the following changes:

- Upinned Chrome version that is tested.
- Run tests on Firefox too.
- Instead of disabling CORS protection, I switched to using a CORS proxy to run the tests. I deployed the proxy to cors.scikit-hep.org (for now restricted it to only work with GH files). We can also use this for tutorials that use JupyterLite. I used https://github.com/Zibri/cloudflare-cors-anywhere
- I changed the awkward-forth test since #1426 made it so that awkward-forth is never needed when reading RNTuple.


ariostas:
I'll wait a bit on this so that we can use the latest version of Pyodide

PR #1465, opened 2025-07-11
fix: ignore case of file extension

### LINKED ISSUE #1462: Unable to open object in file ending with .ROOT
Attempting to directly open an object in a root file that ends in `.ROOT` results in a file not found error, *i.e*:

```
f = uproot.open("test.ROOT")
h = f["htest"]
```
works, but
```
h = uproot.open("test.ROOT:htest")
```
fails with a `FileNotFoundError: [Errno 2] No such file or directory` error. 

I am running version 5.5.1. I believe the issue is the regular expression here:
https://github.com/scikit-hep/uproot5/blob/main/src/uproot/_util.py#L316
which only matches files ending in `.root`, so the full string with the part after the colon is assumed to be the file name. I don't know if the filename is assumed like this anywhere else.

Obviously it's possible to rewrite the calling code to avoid this, but I think the correct behavior would be to not require a particular ending, or to at least accept any case version of it.

ariostas:
Fixes #1462

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -313,7 +313,7 @@ def file_object_path_split(urlpath: str) -> tuple[str, str | None]:
 
     separator = "::"
     parts = urlpath.split(separator)
-    object_regex = re.compile(r"(.+\.root):(.*$)")
+    object_regex = re.compile(r"(.+\.root):(.*$)", re.IGNORECASE)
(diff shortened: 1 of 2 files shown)


ianna (approved):
@ariostas - thanks for fixing it! Looks good to me.

PR #1472, opened 2025-07-21
Allowing ATLAS-style file names for ROOT files

zlmarshall:
The ATLAS production system produces ROOT files with names like "something.root.1" (or "something.root.2" in case the job is the second attempt at producing the file, and so on). The suffix is always "." followed by an integer. In order to ensure that these files are also correctly treated by uproot when they're opened using the file:object style of calling open(), we need either a more general regex or an alternative regex pattern. Here I'm adding an alternative to be as specific as possible about what the alternative should be. There are other ways to solve this problem, but I don't think naming a bajillion files in our production system, or changing the behavior of PanDA, is going to be a viable solution, so better to parse things this way.

Very fundamentally, I don't quite understand why this regex is necessary. One could simply split on the colon and then try/except based on the first part being a file name and the second part being an object in that file. Since you look for the magic in the file anyway, I'm not sure what this function provides for you except perhaps tidier error messages, at the cost of this sort of difficulty in case file naming is abnormal (to be clear, I recognize that this file naming of PanDA's is really not nice, and ROOT itself has broken opening files on occasion because of it).

### DIFF
--- a/src/uproot/_util.py
+++ b/src/uproot/_util.py
@@ -314,11 +314,18 @@ def file_object_path_split(urlpath: str) -> tuple[str, str | None]:
     separator = "::"
     parts = urlpath.split(separator)
     object_regex = re.compile(r"(.+\.root):(.*$)", re.IGNORECASE)
+    # Some files like "something.root.1" need to also be treated as ROOT files
+    object_regex_with_extra = re.compile(r"(.+\.root\.+[0-9]+):(.*$)", re.IGNORECASE)
     for i, part in enumerate(reversed(parts)):
         match = object_regex.match(part)
         if match:
             obj = re.sub(r"/+", "/", match.group(2).strip().lstrip("/")).rstrip("/")
             parts[-i - 1] = match.group(1)
+            break
+        match_with_extra = object_regex_with_extra.match(part)
+        if match_with_extra:
+            obj = re.sub(r"/+", "/", match_with_extra.group(2).strip().lstrip("/")).rstrip("/")
+            parts[-i - 1] = match_with_extra.group(1)
             break
 
     urlpath = separator.join(parts)


ariostas:
Thank you @zlmarshall! That didn't work, which now makes sense to me, since I see that things like `http://...` are failing to be parsed correctly.

src/uproot/_util.py
```diff
@@ -313,7 +313,7 @@ def file_object_path_split(urlpath: str) -> tuple[str, str | None]:
 
     separator = "::"
     parts = urlpath.split(separator)
-    object_regex = re.compile(r"(.+\.root):(.*$)", re.IGNORECASE)
+    object_regex = re.compile(r"(^.*):(.*$)", re.IGNORECASE)
```
ariostas, line 316:
```suggestion
    object_regex = re.compile(r"(.+\.root.*):(.*$)", re.IGNORECASE)
```
I think we can probably just use this regex to make it as generic as possible

tests/test_0976_path_object_split.py
```diff
@@ -202,7 +202,6 @@ def test_url_no_split(input_value):
         "local/file.root.zip://Events",
         "local/file.roo://Events",
         "local/file://Events",
-        "http://xcacheserver:8762//https://originserver:4212/path/file.root.1:CollectionTree",
```
ianna, line 205: @zlmarshall - why is it dropped from the test?

ianna (commented):
@zlmarshall - Thanks for fixing it! I think, it would be nice to have a test. Could we have a small file in scikit-hep data? Thanks.

ianna (approved):
@zlmarshall - Thanks for addressing the comments. It looks great. Please merge it if you are done with it. Thanks!

ariostas:
@all-contributors please add @zlmarshall for code

PR #1479, opened 2025-07-29
chore: officialy support Python 3.14

ariostas:
I added the Python 3.14 classifier, and changed the CI so that it also tests it.


ariostas:
Nevermind, this is still not easily testable

PR #1482, opened 2025-08-01
feat: better show method for RNTuples

ariostas:
This PR improves the show method on RNTuple so that it better shows the structure. This was inspired by UnROOT, which does a much better job of displaying the contents of an RNTuple.

This change breaks the api, but I think it's fine since `show` is something you would only use interactively and not in some existing script.

I still have to do some polishing and maybe add colors.

### DIFF
--- a/src/uproot/behaviors/RNTuple.py
+++ b/src/uproot/behaviors/RNTuple.py
@@ -1530,14 +1530,8 @@ class HasFields(Mapping):
         filter_typename=no_filter,
         filter_field=no_filter,
         recursive=True,
-        name_width=20,
-        typename_width=24,
-        path_width=30,
+        max_width=80,
         stream=sys.stdout,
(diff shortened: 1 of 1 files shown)


ariostas:
I added `**kwargs` just no not break the api, and I fixed the docstring. I think it's not worth adding color since that might complicate things and might not always work.

ariostas:
@jpivarski or @agoose77 could you please add Ianna and me as maintainers in ReadTheDocs? The build randomly fails, so it would be nice to be able to retry it easily.

jpivarski:
I added you, @ariostas, but I don't know @ianna's ReadTheDocs account (or the associated email address; I tried all the email addresses that I know).

image

ianna:
> That's an email address that I tried, among with your GMail and Princeton ones. Do you log into ReadTheDocs (.org) using your GitHub account? If so, your GitHub username should have worked, but that didn't either.

Could you please try again? Thanks

PR #1488, opened 2025-08-27
chore: warn about switching to writing RNTuples by default

ariostas:
This is still in draft since I want to confirm the timeline and make sure that it's not too noisy.


ariostas:
Apart from version number should we also mention a rough date on the timeline? Maybe just "early 2026"?

ianna (approved):
@ariostas - great! Thanks.

PR #1492, opened 2025-09-02
fix: hide some subfields for variant and atomic fields

ariostas:
This PR fixes a couple of things that I missed in #1469.

I made all subfields of variants hidden, since they cannot be accessed directly in a consistent way. Instead, the entire variant field needs to be read.

For example, with the example in the test
```python
data = ak.Array(
    {
        "union1": [
            {"a": 1, "b": 1},
            "two",
            {"a": 3, "b": 3},
            "four",
            "five",
        ],
        "union2": [
            {"a": 1, "b": 1},
            {"a": 2, "b": 2, "c": 2},
            {"a": 3, "b": 3},
            {"a": 4, "b": 4, "c": 4},
            "five",
        ],
    }
)
```
if one tried to do `rntuple["union1.a"]` one would get 2 elements, but it should be length 5 to be consistent. One could argue that in these cases we can just insert `None`s to make it the right shape, but I think the second example shows it's a bad idea in general. In that case, `rntuple["union2.a"]` would be have to include both types of structs, but they could have different types or meanings. So to avoid all these issues, we just forbid accessing them directly.

Additionally, I made the subfield of atomic fields hidden since they are equivalent to the parent field.

### DIFF
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -1539,7 +1539,7 @@ class RField(uproot.behaviors.RNTuple.HasFields):
                 or self.parent.record.flags & uproot.const.RNTupleFieldFlags.REPETITIVE
                 or (
                     self.parent.record.struct_role == uproot.const.RNTupleFieldRole.LEAF
-                    and "std::atomic" in self.parent.record.type_name
+                    and self.record.field_name == "_0"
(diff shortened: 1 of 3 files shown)


ianna (approved):
@ariostas - looks good to me! Thanks!

Issue #1498, opened 2025-09-19
ROOT file getting misinterpreted my uproot compared to ROOT


ikrommyd:
Here is the root file: https://www.dropbox.com/scl/fi/6q4tbou9gpxn66bkxiw9a/UL2016postVFP_Run2016G_el.root?rlkey=vl47r1z8uj4ml3syhusyyip1d&dl=0

```py
uproot.open("UL2016postVFP_Run2016G_el.root:Events").arrays()
# or
uproot.open("UL2016postVFP_Run2016G_el.root:Events")["Jet_dR_Ljet_Isoel"].array()
```
will give you 
```
ValueError: basket 0 in tree/branch /Events;1:Jet_dR_Ljet_Isoel has the wrong number of bytes (10412) for interpretation AsDtype("('>f4', (102,))")
in file /Users/iason/Downloads/UL2016postVFP_Run2016G_el.root
```
The file is readable by ROOT:
```
root [0] TFile *f = TFile::Open("UL2016postVFP_Run2016G_el.root");
Warning in <TClass::Init>: no dictionary for class edm::Hash<1> is available
Warning in <TClass::Init>: no dictionary for class edm::ParameterSetBlob is available
Warning in <TClass::Init>: no dictionary for class edm::ProcessHistory is available
Warning in <TClass::Init>: no dictionary for class edm::ProcessConfiguration is available
Warning in <TClass::Init>: no dictionary for class pair<edm::Hash<1>,edm::ParameterSetBlob> is available
root [1] TTree *t = (TTree*)f->Get("Events");
root [2] t->Scan("Jet_dR_Ljet_Isoel");
***********************************
*    Row   * Instance * Jet_dR_Lj *
***********************************
*        0 *        0 *           *
*        1 *        0 *           *
*        2 *        0 *           *
*        3 *        0 *           *
*        4 *        0 *           *
*        5 *        0 * 0.0040832 *
*        5 *        1 * 3.2452957 *
*        5 *        2 * 2.7313797 *
*        6 *        0 *           *
*        7 *        0 *           *
*        8 *        0 * 0.0178599 *
*        9 *        0 *           *
*       10 *        0 *           *
*       11 *        0 *           *
*       12 *        0 *           *
*       13 *        0 *           *
*       14 *        0 *           *
*       15 *        0 *           *
*       16 *        0 *           *
*       17 *        0 *           *
*       18 *        0 *           *
*       19 *        0 *           *
*       20 *        0 *           *
[...]

PR #1501, opened 2025-09-29
Fix: Handle malformed TLeaf titles to prevent crash with warning

dan-k-k:
Fixes issue #1498.

This PR resolves a ValueError that occurs when reading branches with a specific type of malformed metadata: a TLeaf that has a fixed dimension in its title (e.g., BranchName[102]) but is also designated as a jagged array via fLeafCount. The original logic in _from_leaves would incorrectly parse the fixed dimension, leading to a crash during data reshaping.

This fix adds a check at the top of the interpretation_of function to specifically detect this pattern. When found, it bypasses the faulty logic and returns a simple jagged array interpretation, allowing the file to be read correctly. This also resolves the issue for many other similarly malformed branches in the test file.


ariostas:
Thank you @dan-k-k! It seems like this is a common occurrence, so let's not even bother showing the warning (which is what's causing the tests to fail).

PR #1506, opened 2025-10-06
fix: cernbox file can't be opened over http

### LINKED ISSUE #1484: Cernbox file can't be opened over http
@alexander-held reported that the following code doesn't work.

```python
import uproot
uproot.open({"https://cernbox.cern.ch/remote.php/dav/public-files/8LgmhEZsKofVxSs/301333.root": "reco"})
```
However, it works locally.

The error message is the following.
```
uproot.deserialization.DeserializationError: attempting to get bytes -477079085:-477079084
outside expected range 24769167:24852719 for this Chunk
in file https://cernbox.cern.ch/remote.php/dav/public-files/8LgmhEZsKofVxSs/301333.root
```

Also, it's not the same issue as #1453.

ianna:
Fixes #1484

### DIFF
--- a/src/uproot/source/fsspec.py
+++ b/src/uproot/source/fsspec.py
@@ -145,7 +145,7 @@ class FSSpecSource(uproot.source.chunk.Source):
             # _cat_ranges is async while cat_ranges is not.
             coroutine = (
                 self._fs._cat_ranges(paths=paths, starts=starts, ends=ends)
-                if self._async_impl
+                if self._async_impl and not self._fs._cached
                 else async_wrapper_thread(
                     self._fs.cat_ranges, paths=paths, starts=starts, ends=ends
                 )


src/uproot/source/fsspec.py
```diff
@@ -33,6 +33,15 @@ def __init__(
     ):
         super().__init__()
         self._coalesce_config = coalesce_config
+
+        # Wrap CERNBox URLs automatically with simplecache::
+        def _maybe_wrap_cernbox(url: str) -> str:
```
nsmith-, line 38: Better to put this free function outside the init scope

ianna: thanks! done.

ianna:
> Hmm it seems like this is actually a CERN Box issue.
> 
> I'm looking at the headers with
> 
> ```shell
> curl -s -I -r 0-10 "https://cernbox.cern.ch/remote.php/dav/public-files/8LgmhEZsKofVx\
> Ss/301333.root"
> ```
> 
> which is requesting the 0-10 byte range. The response is this
> 
> ```
> HTTP/1.1 200 OK
> Accept-Ranges: bytes
> ```
> 
> which means that the server ignored the byte range (it should have been 206 instead of 200).
> 
> But still, we should check for this kind of failure modes. I think your implementation is good, but I'll think a bit more about it.

It looks like the heuristic is useful for real remote files, but for unit tests using local HTTP servers or tiny test chunks, it can backfire.

PR #1515, opened 2025-10-23
chore: improve mktree method and update docs to help transition to RNTuple

### LINKED ISSUE #1504: Documentation how to use mktree without assigning to WritableDirectory is hard to find
Most places in docs show how to write a TTree using assignment e.g.

```python
my_directory["tree"] = {"branch1": np.array(...), "branch2": ak.Array(...)}
```

But this is [soon writing RNtuples](https://github.com/scikit-hep/uproot5/pull/1488) (the warning works, it's what lead me to writing this issue 😉). Unfortunately there are still use cases that rely on writing TTrees - e.g. where uproot is used as a compatibility layer to code that expects to read in TTree data (which tbh i think will exist for still a quite long time).

One can find the way to do this in the docs with `mktree` when navigating to [`WritableTree`](https://uproot.readthedocs.io/en/latest/uproot.writing.writable.WritableTree.html) and then scrolling down which took me a while.

I think what could help a lot is to have a code snippet relatively at the top of the `mktree` docstring that shows how to write a full awkward array or pandas dataframe to a ROOT file - or even a quick note "migrating old code that relies on writing TTrees" showing how to replace old by new code, e.g. like

```python
df = pd.DataFrame(...)
# old
my_directory["tree"] = df
# new
my_directory.mktree("tree", {col: dtype for col, dtype in zip(df.columns, df.dtypes)})
my_directory.extend({col: df[col] for col in df.columns})
```

It would also be nice if `mktree` would support writing a full dataframe or array in one step, avoiding some of that boilerplate, but i can understand if you don't want to change this function at this point.

ariostas:
This PR improves the `mktree` method so that it accepts data instead of having to first us it with type data and then use `extend` to write the data. This makes it a much simpler replacement for the old syntax. I also updated the documentation and tests.

Closes #1504

### DIFF
--- a/tests-wasm/test_1272_basic_functionality.py
+++ b/tests-wasm/test_1272_basic_functionality.py
@@ -4,6 +4,11 @@ from __future__ import annotations
 import pytest
 from utils import run_test_in_pyodide
 
+# This comment below is needed for pytest_pyodide to work
+# The issue is that we use our own custom decorator
+# See https://github.com/pyodide/pytest-pyodide/pull/180
+# run_in_pyodide
+
 
 # Taken from test_0034_generic_objects_in_ttrees.py
 @run_test_in_pyodide(test_file="uproot-HZZ-objects.root", packages=["pytest", "xxhash"])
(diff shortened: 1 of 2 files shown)


ianna (approved):
@ariostas - Great! Thanks.

PR #1525, opened 2025-11-10
chore: pytest log_level is better than log_cli_level

henryiii:
Committed via https://github.com/asottile/all-repos

See https://github.com/scientific-python/cookie/issues/679


ianna (approved):
@henryiii - thanks! It looks great. Please merge it if you are done with it. Thanks!

PR #1543, opened 2025-12-21
feat: set divisions to unknown if we expect a read error report in `uproot.dask`

ikrommyd:
Following Martin's suggestion here: https://github.com/dask-contrib/dask-awkward/pull/592#issuecomment-3222290975, I am setting the divisions to unknown if we have an uproot read error report.
The reason is that when we have such a report object, we do not actually know if we'll be able to read all of our partitions. One may have 2 partitions [[0,20], [20,40]] and manage to read only one of them due to some data access error that the report is meant to catch and tell you about it later.
However, if we have known divisions, dask-awkward can do certain optimizations that will give back wrong results if reading a partition fails. it will tell you for example that `dak.num(events, axis=0)` is 40 just because it knows the divisions without calculating it during computation. If the user does an operation like `some average = some quantity / dak.num(events, axis=0)`, then that is wrong because the denominator will be 40 but the user actually managed to read 20 events, not 40 and the numerator is calculated from 20 events only.

Before:
```
In [7]: events, report = uproot.dask({"../coffea/tests/samples/nano_dy.root": {"object_path": "Events", "steps": [[0, 20], [20, 40]]}}, allow_read_errors_with_report=True)

In [8]: dak.num(events, axis=0)
Out[8]: dask.awkward<num, type=Scalar, dtype=int64, known_value=40>

In [9]: dak.num(events, axis=0).compute()
Out[9]: 40
```
After:
```
In [5]: events, report = uproot.dask({"../coffea/tests/samples/nano_dy.root": {"object_path": "Events", "steps": [[0, 20], [20, 40]]}}, allow_read_errors_with_report=True)

In [6]: dak.num(events, axis=0)
Out[6]: dask.awkward<numaxis0, type=Scalar, dtype=int64>

In [7]: dak.num(events, axis=0).compute()
Out[7]: np.int64(40)
```
If we don't want the report, nothing changes with this PR:
```
In [9]: events = uproot.dask({"../coffea/tests/samples/nano_dy.root": {"object_path": "Events", "steps": [[0, 20], [20, 40]]}}, allow_read_errors_with_report=False)

In [10]: dak.num(events, axis=0)
Out[10]: dask.awkward<num, type=Scalar, dtype=int64, known_value=40>

In [11]: dak.num(events, axis=0).compute()
Out[11]: 40
```

Let me know if this is not the right way to implement this in `uproot.dask` and there is a better way.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -38,7 +38,7 @@ jobs:
         run: ls -lha dist/
 
       - name: Generate artifact attestation for sdist and wheel
-        uses: actions/attest-build-provenance@977bb373ede98d70efdf65b84cb5f73e068dcc2a # v3.0.0
+        uses: actions/attest-build-provenance@00014ed6ed5efc5b1ab7f7f34a39eb55d41aa4f8 # v3.1.0
         with:
           subject-path: dist/uproot-*
 
--- a/.github/workflows/upload-nightly-wheels.yml
+++ b/.github/workflows/upload-nightly-wheels.yml
@@ -33,7 +33,7 @@ jobs:
         run: ls -lha dist/*.whl
 
       - name: Upload wheel to Anaconda Cloud as nightly
-        uses: scientific-python/upload-nightly-action@b36e8c0c10dbcfd2e05bf95f17ef8c14fd708dbf # 0.6.2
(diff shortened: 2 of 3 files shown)


martindurant:
Note: there are some methods to discover and set the divisions attribute again, so user beware. I don't think they would be called in normal uproot operations.

ikrommyd:
I assume this should be good to go?

PR #1547, opened 2026-01-05
chore: remove minio dependency

### LINKED ISSUE #1546: Move away from minio
[MinIO](https://github.com/minio/minio) changed to being in maintenance mode and all further development was stopped, so we should find an alternative for reading from S3 sources. There is very little code that uses minio, so it should be straightforward to move to a different library.

ariostas:
Closes #1546 

This was pretty straightforward since `fsspec` can use `s3fs`, so we didn't really need `minio`.


ianna (approved):
@ariostas - thanks for fixing it so quickly! It looks great. Indeed it is a better option. I think we should announce it to the community. Thanks!

PR #1550, opened 2026-01-06
chore: drop support for Python 3.9

ariostas:
Since Python 3.9 reached its EOL, we're dropping support for it in the upcoming Uproot v5.7.0 release. I took the chance to switch a couple of things to pattern matching, but later on we can clean up more things.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -659,6 +659,15 @@
       "contributions": [
         "code"
       ]
+    },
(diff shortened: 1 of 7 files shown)


ianna (approved):
@ariostas - This is great, thanks! It’s really nice to have a co-pilot working through this with me. Please, merge it if you are done with it. Thanks!

Issue #1374, opened 2025-02-04
uproot.iterate sets chunk size to 1 when unvalid pattern in filter_name


ArturU043:
### Definition

Entering a nonexistent branch name / pattern in the filter_name argument causes the function to open and load the file data many times with a step_size automatically set to 1.
The number of steps to load the file is equal to the number of rows in the file. This leads to a very long execution time for large files.

This situation can happen frequently when a typo is inserted. 

## Code example 

[uproot.iterate-bug.ipynb](https://gist.github.com/ArturU043/332ef582254f14d9d5425a24a2cfbffe)

PR #1554, opened 2026-01-17
fix: exit early from uproot.iterate when no branches/fields match the filters

Rachit931:
This PR fixes an issue where providing a non-existent "filter_name" to "urpoot.iterate" did not raise an error. 

When "filter_name" does not matches any provided branch in the ROOT file, the resolved branch set empty which leads to a fallback iteration path with "chinks=None", causing "steps_per_file" to effectively become 1. This resulted in empty outputs to and the file being processed entry-by-entry.

This changes makes "uproot.iterate" fail fast by raising a "ValueError" when "filter_name" does not match any existing branches. A regression test is also included for cross checking. 

Address : #1374

### DIFF
--- a/tests/test_0043_iterate_function.py
+++ b/tests/test_0043_iterate_function.py
@@ -218,4 +218,3 @@ def test_function_iterate_pandas_2():
     for arrays, report in uproot.iterate(files, "Muon_Px", report=True, library="pd"):
         assert arrays["Muon_Px"].index.values[0] == expect
         expect += report.tree.num_entries
-


ariostas (changes requested):
@Rachit931 thanks for working on this!

However, I don't think we should raise an error in this case. The current implementation allows for the flexibility of using files with inhomogeneous branches. For example, you could do something like this:

```python
import uproot
import numpy as np

tree1_data = {
    "branch1": np.ones(1000),
    "branch2": np.zeros(1000)
}
tree2_data = {
    "branch1": np.ones(1000),
    "branch2": np.zeros(1000),
    "branch3": np.zeros(1000)
}

with uproot.recreate("test_file1.root") as file:
    file["Tree"] = tree1_data
with uproot.recreate("test_file2.root") as file:
    file["Tree"] = tree2_data

arrays = list(uproot.iterate("test_file*.root:Tree", filter_name="branch3"))
print(arrays)
```

So it would be good to keep that flexibility. You correctly identified that the issue is that it ends up setting the step size to 1. So a simple solution could be to set the step size to a big number if it will just return an empty array.

Rachit931:
Thanks for the clarification ! 
I've removed the error and updated iterate to skip files where filter_name matches no branches, which avoids the step_size issue preserving heterogeneous file behavior
I've also updated the test to reflect the intended behavior ( an empty array )

ariostas:
I'll need to think about this a bit more.

- I think it probably makes sense to skip when there are no matches, but I'll think some more about it. I'm not sure if there was a reason to do it this way. Maybe to keep then number of entries consistent?
- It would make more sense to do the same for `filter_typename` and `filter_branch`, to be consistent.

ariostas:
@Rachit931 Okay, yeah, I don't think there's any good reason to return a bunch of empty arrays. Pleast go ahead and implement the skipping for the other two filters. Thank you!

Also, I think it would be worth considering another approach since this still doesn't address the issue of the step size being set to 1. So if someone does `file["tree"].iterate(filter_name="non-existent-name")` they would still get a bunch of empty arrays. A better option might be to change this function so that if the interpretations are empty then it just sets the step size to a big number.

https://github.com/scikit-hep/uproot5/blob/d3152ef9fcf1e07afc2be00d25012f73010562f9/src/uproot/behaviors/TBranch.py#L3555-L3576

Issue #1429, opened 2025-04-28
Usage of options like max_num_elements with uproot.open()


dudarboh:
Maybe it is just me, but I don't understand how to use the `max_num_elements` argument with `uproot.open()`

I have tried a few options on my ROOT file and none of them work:
```python
import uproot

with uproot.open("filename.root:treename", options={"max_num_elements":1000}) as events:
    data = events.arrays(["mcTotal"])
    print(len(data["mcTotal"]))

with uproot.open("filename.root:treename", max_num_elements=1000) as events:
    data = events.arrays(["mcTotal"])
    print(len(data["mcTotal"]))
```

I assume it is not a bug, but rather I am missing something very simple.
If that is the case, I think a little example in the docs wouldn't hurt.
Or is it a bug?

ariostas:
Hi @dudarboh, thanks for bringing this up! This is definitely something that could use some better documentation. `max_num_elements` is used when opening a file with `xrootd` to figure out how to split things up (as far as I understand).

What you want is the `entry_stop` argument of the `arrays` function.
```python
with uproot.open("filename.root:treename") as events:
    data = events.arrays(["mcTotal"], entry_stop=1000)
    print(len(data["mcTotal"]))
```

PR #1555, opened 2026-01-19
docs: clarify max_num_elements option behavior

Rachit931:
This PR adds a short clarification to the docstring for `max_num_elements` explaining that it can be provided either directly as a keyword argument or via the `options` dictionary.

This addresses the potential confusion for users and aligns the documentation with existing behavior.

Addresses #1429

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -177,6 +177,7 @@ def dask(
     * handler (:doc:`uproot.source.chunk.Source` class; None)
     * timeout (float for HTTP, int for XRootD; 30)
     * max_num_elements (None or int; None)
+        The maximum number of elements to be requested in a single vector read, when using XRootD.
     * num_workers (int; 1)
     * use_threads (bool; False on the emscripten platform (i.e. in a web browser), else True)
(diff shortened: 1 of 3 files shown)


src/uproot/reading.py
```diff
@@ -83,6 +83,8 @@ def open(
     * handler (:doc:`uproot.source.chunk.Source` class; None)
     * timeout (float for HTTP, int for XRootD; 30)
     * max_num_elements (None or int; None)
+        Limits the total number of elements read from jagged arrays to prevent excessive memory use.
```
ariostas:
```suggestion
        The maximum number of elements to be requested in a single vector read, when using XRootD.
```

src/uproot/reading.py
```diff
@@ -83,6 +83,8 @@ def open(
...
+        This can be provided either directly as a keyword argument or inside the `options` dictionary.
```
ariostas:
```suggestion
```

The options dictionary holds the extra kwargs, so this line is not needed.

Rachit931:
Thanks for clarifying ! I have updated it accordingly.

Issue #1514, opened 2025-10-20
GPU RNTuple reading doesn't work with new versions of kvikio


ariostas:
The `kvikio.nvcomp_codec` module was deprecated and removed a while ago, so now RNTuple reading on GPU doesn't work.

@fstrug Do you know if we can just switch to adding the `nvidia-nvcomp-cu12` dependency and make the following substitutions? Or are there some subtleties?

`import kvikio.nvcomp_codec as kvikio_nvcomp_codec` -> `from nvidia import nvcomp`
`kvikio_nvcomp_codec.NvCompBatchCodec` -> `nvcomp.Codec`
`codec.decode_batch` -> `codec.decode`

Issue #1497, opened 2025-09-11
Move RNTuple GPU Decompression from KvikIO to nvCOMP


fstrug:
# Problem
The kvikio library is removing its python bindings to nvcomp - https://github.com/rapidsai/kvikio/pull/798. The [GPU decompression in RNTuple reading](https://github.com/scikit-hep/uproot5/blob/main/src/uproot/models/RNTuple.py#L1792) will need to be updated instead to use the official nvCOMP python library - https://github.com/NVIDIA/nvcomp.

Initially, I stayed with the kvikio library for nvCOMP decompression bindings because its API supports user-provided output buffers. The idea was that we could reduce memory allocations by doing so, but it turns out the kvikio implementation [creates unnecessary buffers for the decompressed data which is later copied into the user-provided buffers](https://github.com/rapidsai/kvikio/blob/8e11ed2edc0c392ec409d2309793c11af8e91ee2/python/kvikio/kvikio/nvcomp_codec.py#L216). 

The nvCOMP team is considering updating the python implementation to allow for user-provided output buffers. https://forums.developer.nvidia.com/t/please-support-user-provided-output-buffers-in-python-api/332763

# Solution
Ultimately, transitioning from the kvikio bindings to the official nvCOMP python API should not significantly impact performance. The code however will need some adjustments. First, the official nvCOMP API [does not take cupy.ndarray as inputs to its codecs](https://docs.nvidia.com/cuda/nvcomp/samples/nvcomp.html#Zero-copy-import-device-array). Uproot will have to handle this conversion to and from `nvcomp.Array`.
```
data_gpu = cp.array(ascending)
nvarr_d = nvcomp.as_array(data_gpu)
lz4_comp_arr = lz4_codec.encode(nvarr_d)
```
Second until user-provided buffers are supported, uproot will have to handle the copy into the full output buffer.

# Motivation for using user-provided output buffers
Through RNTuple metadata we know the total decompressed size for all pages in a given cluster.
https://github.com/scikit-hep/uproot5/blob/9798afd274ba4c4f4d94ce1853a7f062f1ddeb4e/src/uproot/models/RNTuple.py#L764-L765
[...]

PR #1557, opened 2026-01-23
fix: GPU interpretation of RNTuple data working and implement nvCOMP decompression with nvCOMP library

fstrug:
This PR brings GPU interpretation of RNTuple data back in working order and transitions GPU decompression to using the nvCOMP python package. This addresses issues https://github.com/scikit-hep/uproot5/issues/1514 and https://github.com/scikit-hep/uproot5/issues/1497.

After some updates to post-processing and deserialization of RNTuple data, GPU interpretation no longer worked. A `cupy.ndarray` cannot be post processed in place as was done. Therefore `post_process` was renamed to `post_process_in_place` for use on `numpy.ndarray`, and a new `post_process` function which can perform a similar function on `cupy.ndarray` is added. Further minor tweaks to the GPU interpretation made to match behavior of CPU.

Kvikio deprecated their bindings to nvCOMP in this PR https://github.com/rapidsai/kvikio/pull/714. GPU decompression is now handled with the official Nvidia [nvCOMP python package](https://docs.nvidia.com/cuda/nvcomp/py_api.html).

### DIFF
--- a/tests-cuda/test_0962_rntuple_update.py
+++ b/tests-cuda/test_0962_rntuple_update.py
@@ -53,9 +53,7 @@ def test_new_support_RNTuple_split_int16_reading(backend, interpreter, library):
         assert df.one_integers[0] == 2
         assert df.one_integers[-1] == 1
         assert ak.all(
-            library.unique(
-                library.array(df.one_integers[: len(df.one_integers) // 2])
-            )
+            library.unique(library.array(df.one_integers[: len(df.one_integers) // 2]))
             == library.array([2])
         )
         assert ak.all(


ariostas:
@fstrug I refactored the CPU code a bit to make the dtypes less confusing and so that hopefully both CPU and GPU can use the same `post_process` method. I broke the GPU side, but I'll work on fixing that.

ariostas:
Seems like it was only a matter of switching to the in_place method. Let me know what you think.

fstrug:
Looks good to me, even managed to reduce the lines of code just a smidge. :)

PR #1564, opened 2026-01-30
fix: raise clear error for malformed rntuple offset arrays 

Rachit931:
WHAT THIS PR DOES : 

This PR adds a validation step when constructing arrays from RNtuple data.

For jagged fields, RNtuple offset buffers are assumed and expected to have
cluster_num_entries + 1 elements. When this expectation is violated,
the current code can fail later with confusing errors from Awkward.

This change checks the offset length immediately after buffers are
assembled and raises a clear ValueError when the offsets are malformed.

WHY IS IT NEEDED : 

Malformed offset arrays can currently generate into ak.from_buffers,
leading to hard-to-debug failures that doesn't clearly point to corrupted
RNtuple data.

Tests

A focused test was added that directly simulates the internal
container dictionary produced during arrays() execution and verifies
that invalid offset lengths raise an error.

The test does not rely on ROOT file I/O and targets only the offset
length assumption.

The validation logic is implemented as a small helper function to keep
arrays() readable and make the assumption easy to test.
I’m happy to inline this logic into arrays() if preferred .

### DIFF
--- a/src/uproot/behaviors/RNTuple.py
+++ b/src/uproot/behaviors/RNTuple.py
@@ -841,16 +841,15 @@ class HasFields(Mapping):
         )
         entry_start -= cluster_offset
         entry_stop -= cluster_offset
-
-        if entry_start == 0 and entry_stop == cluster_num_entries:
-            _validate_rntuple_offsets(container_dict, cluster_num_entries)
-
         arrays = uproot.extras.awkward().from_buffers(
             form,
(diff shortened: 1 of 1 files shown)


Rachit931:
I added the offset validation as a small helper function to make the assumption explicit and easy to test in isolation. 
If preferred for the logic to live directly inside arrays(), happy to inline it and remove the helper.

PR #1573, opened 2026-02-07
fix: respect entry bounds for RNTuple.iterate and add of regressions tests for out of bounds

KaranSinghDev:
This PR adds a regression test to verify the behavior of TBranch.array() and TTree.iterate() when provided with entry range that are our of bounds.

While uproot handles these cases by returning empty arrays or clamped ranges, I believe there is no test in the repo for this specific case. I made this test to work as a guardrail for future changes that might cause Index Error,  data corruption during refactoring 

The tests consists of handling 5 edge cases:
1.  **Hard Out-of-Bounds:** When a range is requested entirely past the end of file. (returns empty array).
2.  **Clamped Read:** When a range overlaps the end of the file. ( returns a truncated/clamped array).
3.  **Exact Boundary:** When a read starts exactly at the file's last entry. ( returns empty array).
4.  **Negative Indexing:** When a large negative start index is provided. ( returns the whole file, matching Python's convention).
5.  **Iterative OOB:**  TTree.iterate yields zero batches for an invalid range.

I have validated the test with pytest (all 5), verifed with pre-commit and also have avoided external network dependencies .


KaranSinghDev:
@ariostas Thank you for pointing out, I have changed the name to 1573. About the RNTuples tests, I need some time, but I will share update .

Issue #1568, opened 2026-02-04
Consider using `numcodecs` over `cramjam`


ariostas:
Nick Smith pointed out that we should consider switching from `cramjam` to `numcodecs` since it supports the same compression algorithms, but additionally supports pre-compression filters that are useful for RNTuple IO.

ariostas:
In #1574, @Rachit931 found that `numcodecs` doesn't support LZ4 compression in a standard way. Along with the lack of active development of `numcodecs`, I think it's better to stick with `cramjam`.

Issue #1570, opened 2026-02-05
Clean up `uproot.extras`


ariostas:
The `extras` submodule contains a bunch of handles for dependencies that used to be optional, but are required now. It would be good to remove the required dependencies from there and switch to using `import dependency` instead of `dependency = uproot.extras.dependency()` throughout the codebase. And we should also update the readme to reflect the changes in dependencies.

Rachit931:
I'm definitely interested, will update after comparing.

Rachit931:
I’ve completed the cleanup of uproot.extras.awkward usage and have a PR ready.
All tests pass and the skipped tests are due to missing optional dependencies.

PR #1574, opened 2026-02-08
feat: use numcodecs as an optional backend for LZMA, ZSTD, ZLIB 

Rachit931:
This PR is towards evaluating numcodecs as an alternative backend for compression in uproot, as discussed in #1568. 

It introduces numcodecs as an optional backend for LZMA and ZSTD compression and decompression within compression.py, while preserving existing interfaces, error handling and fallback behavior. The existing cramjam (and stdlib, where applicable) implementations remains unchanged and are used as fallbacks if numcodecs is unavailable or unsuitable.

As of now, I have intentionally limited the scope of this PR to LZMA and ZSTD. While reading through the existng code, it seemed to me that for zlib and LZ4 paths rely on mroe specific behavior and error handling, and I wasn't confident that a straightforward swap to numcodecs would preserve semantics there. To avoid guesing and to keep this change minimal, those codecs are left unchanged for now.

### DIFF
--- a/src/uproot/compression.py
+++ b/src/uproot/compression.py
@@ -273,7 +273,7 @@ class LZMA(Compression, _DecompressLZMA):
             out = codec.encode(data)
             out = b"".join(out) if isinstance(out, list) else bytes(out)
             return out
-        
+
         except ModuleNotFoundError:
             # Failure due to numcodecs not installed = fall back to cramjam/stdlib
             pass
--- a/src/uproot/compression.py
+++ b/src/uproot/compression.py
@@ -196,10 +196,9 @@ class _DecompressLZMA:
             codec = numcodecs.LZMA()
             decoded = codec.decode(data)
 
-            if isinstance(decoded, list):
-                decoded = b"".join(decoded)
-            else:
-                decoded = bytes(decoded)
+            decoded = (
+                b"".join(decoded) if isinstance(decoded, list) else bytes(decoded)
+            )
 
             # numcodecs does not gaurentee outpute size ( must validate )
             if uncompressed_bytes is not None and len(decoded) != uncompressed_bytes:
@@ -363,10 +362,9 @@ class _DecompressZSTD:
             codec = numcodecs.Zstd()
             decoded = codec.decode(data)
 
-            if isinstance(decoded, list):
-                decoded = b"".join(decoded)
-            else:
-                decoded = bytes(decoded)
+            decoded = ( 
+                b"".join(decoded) if isinstance(decoded, list) else bytes(decoded)
+            )
 
             # numcodecs does NOT guarantee outpute size (must validate)
             if len(decoded) != uncompressed_bytes:
--- a/src/uproot/compression.py
+++ b/src/uproot/compression.py
@@ -189,18 +189,47 @@ class _DecompressLZMA:
     _method = b"\x00"
 
     def decompress(self, data: bytes, uncompressed_bytes=None) -> bytes:
+        # Try numcodecs 
+        try : 
+            import numcodecs
+
+            codec = numcodecs.LZMA()
+            decoded = codec.decode(data)
+
+            # numcodecs does not gaurentee outpute size ( must validate )
+            if uncompressed_bytes is not None and len(decoded)!=uncompressed_bytes : 
+                raise ValueError ( 
+                    "numcodecs LZMA produced incorrect output size"
+                )
+            
+            return decoded 
+        
+        except ImportError: 
+            # Failure due to numcodecs not being installed = fall back to cramjam/stdlib 
+            pass 
+        
+        except ValueError : 
+            # Failure due to output-size validation failed = fall back to cramjam/stdlib
+            pass
+
+        # Fallback : Try cramjam(preferred) or stdlib 
         cramjam = uproot.extras.cramjam()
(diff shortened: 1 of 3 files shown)


Rachit931:
CI showed a `list` was still reaching the fsspec write path in `test_fsspec_writing_ssh_simple`.
Even after adding a small normalization before block splitting in `def compress()` so the internal logic always works on bytes-like data.
I still can't figure out where the leak is coming from.

ariostas:
@Rachit931 the fsspec issue is unrelated to what you're doing. They broke something with their latest release. I'm just waiting a couple of days to see if they patch it or if we have to pin to an older version.

Rachit931:
Thanks for the clarification! 
In that case, I am happy to drop the additional normalization I added while chasing the fsspec failure, if that's preferred. 
I'll wait for the fsspec situation to settle before making any further updates. 

Separate question : Once this settles, I wanted to ask if the extending the numcodecs backend to other ready codecs (eg LZ4 and zlib) or keeping this PR limited to LZMA/ZSTD ?

ariostas (commented):
Thank you, @Rachit931. I left a couple of comments.

> I wanted to ask if the extending the numcodecs backend to other ready codecs (eg LZ4 and zlib) or keeping this PR limited to LZMA/ZSTD ?

It would be good to extend them to the full set of codecs.

To make sure that this is being tested by the CI, we should add the `numcodecs` dependency to the `test` group in `pyproject.toml`. Then, in the future when we're ready to fully switch we can make it a strict dependency and remove cramjam.

Lasty, it would be better if you can use commit messages where you describe what changed in that commit instead of using the same message every time. This makes it easier to go back and see how things evolved.

src/uproot/compression.py
```diff
@@ -342,8 +408,27 @@ def level(self, value):
...
+        except ModuleNotFoundError:
+            # Failure due to numcodecs not installed = Fall back
+            pass
+
+        # Fallback : cramjam
+        cramjam = uproot.extras.cramjam()
+        zstd = getattr(cramjam, "zstd", None)
+
+        if zstd is None:
+            raise RuntimeError("ZSTD compression requires ramjam or numcodecs")
```
ariostas: cramjam always has zstd, so you don't have to do this. For lzma that was done because only recent versions have it.

src/uproot/compression.py
```diff
@@ -539,8 +624,14 @@ def compress(data: bytes, compression: Compression) -> bytes:
...
     """
+
+    def _normalize_bytes(x):
+        if isinstance(x, list):
+            return b"".join(x)
+        return x
+
     if compression is None or compression.level == 0:
-        return data
+        return _normalize_bytes(data)
```
ariostas: I don't think `data` will ever be a list.

ariostas (commented):
> When I added numcodecs as a backend for ZLIB, LZ4, multiple tests kept failing with the error : local variable 'out' referenced before assignment and RuntimeError : LZ4 decompression error.

The first error, it sounds like there was something wrong with your code. You should look for why it might be unassigned. For the second one, try to find an example where that happens. If possible, try to find a minimal reproducer of a byte-string that works with `cramjam` but not with `numcodecs`. If that's the case, we should report that to them so they can fix it.


> To the extent I could find out : For ZLIB and LZ4,uproot already knows how large the data should be after compression and expects the decompression library to check this while decoding.
stdlib and cramjam both allow uproot to pass in this expected size and fails immediately if the decoded data does not match it which helps in catching corrupted or mismatched data early.
Whereas, numcodecs does not provide a way to pass in or check the expected output size during decoding.
Maybe there's a modification which I can't this of, would love to know your opinion.

You are already doing the right approach of just explicitly verifying that the length matches the expected one.

Rachit931:
Appreciate the feedback ! I’m currently looking into a minimal reproducer for a byte-string for the `numcodecs` case . In the meantime, I’ve simplified the code per the feedback and pushed those changes.

And I also have a PR ready with the cleanup related to issue #1570, will include that once the changes are settled here.

Rachit931:
@ariostas  This is the reproducer I could find : 
```
import cramjam
import numcodecs
import os

raw = os.urandom(1024)

compressed = cramjam.lz4.compress(raw)

expected_size = len(raw)

# Test cramjam
out1 = cramjam.lz4.decompress(compressed, expected_size)
print("cramjam is functional:", len(out1))

# Test numcodecs
codec = numcodecs.LZ4()
try:
    out2 = codec.decode(compressed)
    print("numcodecs is functional", len(out2))
except Exception as e:
    print("numcodecs failed:", e)
``` 

Output : 
```
cramjam  is functional: 1024
numcodecs failed: LZ4 decompression error: -7
```

This proves that the same valid LZ4 byte-string can be decompressed by `cramjam` but fails with `numcodecs`. 
Please let me know if I am missing something or misusing the API.

ariostas:
Thank you, @Rachit931! I'd say we should wait a bit before moving on with this. I'm getting a bit worried because they still haven't commented on the issue you opened in the `numcodecs` repo. So I'm not sure how actively maintained it is.

ariostas:
It seems like the `numcodecs` devs are not interested in solving this. Considering that `cramjam` is very actively supported, I no longer think it's worth making the switch. If this changes later on, we can reconsider. Thanks for your help, @Rachit931!

PR #1579, opened 2026-02-16
refactor: Removes optional handling for required dependencies.  

Rachit931:
This PR removes the use of uproot.extras for dependencies that are now required: awkward, cramjam, and xxhash.

Since these libraries are listed in install_requires, they are no longer treated as optional. Their imports have been simplified accordingly.

The README has also been updated to reflect the current set of required dependencies and to remove outdated language suggesting that Awkward is optional.

There are no functional changes in this PR. The goal is to align the implementation and documentation with the current packaging configuration.

Tests were left unchanged to keep this PR focused on the dependency cleanup.
Addresses #1570 .

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -668,6 +668,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "wdconinc",
+      "name": "Wouter Deconinck",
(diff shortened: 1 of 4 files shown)


Rachit931:
Thanks ! Happy to help.

Issue #1510, opened 2025-10-15
RNTuple version 1.0.1.0 can't be read


ariostas:
They added some extra stuff to the footer in this version and uproot tries to interpret this part as a checksum, so it fails.

Additionally, we need to add proper support for RNTuple attributes.

ariostas:
I'll close this issue since these RNTuples work now, but I'll open a new open for proper RNTuple attributes support.

PR #1580, opened 2026-02-16
chore: support RNTuple v1.0.1.0 format (attribute set record frame)

wdconinc:
This PR adds support for RNTuple v1.0.1.0 files.

In the new version a new attribute set record frame was added; [root-project/root@b6e7318](https://github.com/root-project/root/blob/b6e7318cdfb657455188a47ea4dcc1da5dceb2dd/tree/ntuple/doc/BinaryFormatSpecification.md#linked-attribute-set-record-frame)). This is the format in used by ROOT v6.38. The new format requires version-dependent parsing since it is not just an empty frame, but there is no frame at all.

The tests (written with help of AI, so somewhat wordy and with more asserts than maybe necessary) will require the test data added in https://github.com/scikit-hep/scikit-hep-testdata/pull/208. Both tests succeed in local tests:
```console
>>> import sys
>>> sys.path.insert(0, 'tests')
>>> import test_1510_rntuple_v1010
>>> test_1510_rntuple_v1010.test_rntuple_v1010_footer_parsing()
>>> test_1510_rntuple_v1010.test_rntuple_v1000_backward_compatibility()
>>> 
```

This only partly addresses https://github.com/scikit-hep/uproot5/issues/1510. No change to uproot support for the RNTuple attributes.

### DIFF
--- a/src/uproot/models/RNTuple.py
+++ b/src/uproot/models/RNTuple.py
@@ -1583,9 +1583,11 @@ class ClusterGroupRecordReader:
 class LinkedAttributeSetRecordReader:
     def read(self, chunk, cursor, context):
         out = MetaData("LinkedAttributeSetRecord")
-        out.schema_version_major, out.schema_version_minor, out.anchor_uncompressed_size = cursor.fields(
-            chunk, _rntuple_linked_attribute_set_format, context
-        )
+        (
+            out.schema_version_major,
+            out.schema_version_minor,
+            out.anchor_uncompressed_size,
+        ) = cursor.fields(chunk, _rntuple_linked_attribute_set_format, context)
         out.locator = LocatorReader().read(chunk, cursor, context)
         out.name = cursor.rntuple_string(chunk, context)
         return out


wdconinc:
FYI @ariostas

ariostas (approved):
Thank you, @wdconinc! After the url fix, and moving the format definitions to the top, it looks good to me.

My plan is to do some refactoring at some point to make the parsing more robust against version changes. Instead of checking versions, what we should be doing is checking the position of the cursor. This is why RNTuple frames contain size information. If we see that the cursor is at the end, we know that extra stuff like attributes are not present. But on the other hand, if after reading all the known pieces, we find that the cursor is not at the end, we'll know there's extra stuff and that we should skip over it. This way, it won't break again if they add new things to the spec. (This applies to more than just the footer)

However, this is good for now. Uproot will start delegating metadata parsing to `rootfilespec`, so I will work on implementing that logic there.

By the way, would you want to have a patch release soon for your `eicrecon` work?

ariostas:
@all-contributors please add @wdconinc for code

PR #1599, opened 2026-03-16
fix: resolve issue mktree/mkrntuple creating subdirectories

ariostas:
This PR fixes #1598, the issue was that a newly-created `WritableDirectory` was not being tracked properly, so it was being recreated during subsequent calls. Now, it is added to `_subdirs`, which fixes the issue.

AI disclosure: I used Claude Code to debug this.

### DIFF
--- a/src/uproot/writing/writable.py
+++ b/src/uproot/writing/writable.py
@@ -1237,6 +1237,7 @@ class WritableDirectory(MutableMapping):
                     self._file.uuid_function(),
                 ),
             )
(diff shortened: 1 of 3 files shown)


ariostas:
It should be fine since the issue is not specific to `mktree`, but yeah, let me add a test for it.

ikrommyd (approved):
I tested this with the actual issue we had in HiggsDNA and it looks good.

PR #1601, opened 2026-03-16
fix: fix rare issue that causes corrupted free_num_bytes

### LINKED ISSUE #1578: File Corruption after Repeatedly Writing Hist Histograms
Uproot version 5.7.1

For my analysis I have many intermediate histogram products I am constructing in python and wish to save into a temporary file so I don't have to re-run the time-consuming pre-processing stage. For some reason though, partway through my processing the root file becomes corrupted and `uproot.update("./intermediate.root")` fails on the file due to the assert [on this line](https://github.com/scikit-hep/uproot5/blob/6995bd1a0c0d689cd35d1ed4cd073910c7bf9c64/src/uproot/writing/_cascade.py#L577). At that point uproot is still able to read the histograms in the file and ROOT can still open the file without reporting issues.

Here is a download link (~360MB) to the corrupted root file: [Google Drive - intermediate.root](https://drive.google.com/file/d/1wYUeliP_qsxf5OjOSBjiJY6rVhtOBkEz/view?usp=sharing)

Basically the preprocessing consists of:
```
# We will store intermediate results in "intermediate.root"
with uproot.create("intermediate.root") as file:
    pass

for raw_data_file in raw_data_files:
    # Read in data...
    # Construct hist Histograms...
    directory = 'rawroot_' + str(raw_data_file_number) + '/'
    with uproot.update("intermediate.root") as file:
        output_hist: hist.Hist
        for output_hist in output_hist_list:
            file[directory + output_hist.name] = output_hist
```
I don't see any obvious issues with this approach, but anyways the pre-processing fails partway through due to the corruption issue I observe blocking further writes to the intermediate result file.

Here are some bug-testing steps I have taken:

1. Maybe the issue is because of the large number of histograms I am writing to the file?
It didn't seem likely to me, but anyways I ran a toy example where I wrote 100,000 histograms to a file successfully using the same directory structure I have in my sample file. No issues here.

2. Maybe the last histograms I wrote into the file during pre-processing are corrupt for some reason?
I checked the possibly suspicious histograms in python before they get written and they seem fine.

3. What if I try writing each rawroot raw data file's intermediate output into a separate new intermediate ROOT file instead of using directories like in my sample file?
This works! Pre-processing completes and I can continue with analysis. So, for now I am using this workaround

4. Since I can read the `intermediate.root` file just fine, I tried re-reading all written histograms back in and immediately copying them to another root file.
So long as I perform this all inside a single `with` block, there is no issue to keep writing more histograms:
```
# This works
with uproot.open('intermediate.root') as in_file:
    with uproot.create('intermediate_copy.root') as out_file:
        # Copy all histograms from in to out
        # Now write more histograms to out_file, this succeeds
        pass
    pass

# Now if I try opening with uproot.update('intermediate_copy.root') I have no issues
```
However, just making a copy and re-opening the copy does not
```
# This doesn't work
with uproot.open('intermediate.root') as in_file:
    with uproot.create('intermediate_copy.root') as out_file:
        # Copy all histograms from in to out
        pass
    pass

# Fails with same assert error
with uproot.update('intermediate_copy.root') as file:
    pass
```

So, I'm kind of at a loss what the problem might be. My best (only?) guess at this point is that repeatedly re-opening, adding a few histograms, and then closing a ROOT file is somehow causing this corruption. Then again, performing a deep copy of all histograms in my file using just a single `while` block also results in a corrupted file. Anyways, I have a workaround solution for now (using multiple intermediate files - one for each raw data file), but I'm curious what's causing this.

ariostas:
This PR solved the issue that was reported in #1578. I used Claude Code to help with the debugging. This is the summary of what it found:

> I've found the bug. Here's what happens:
>
>  1. When allocate() consumes a free slice (line 641-643 or 649-652), it modifies _data.slices but does not reset _data._allocation
>  2. So _data.allocation returns a stale, larger cached value
>  3. In FreeSegments.write() (lines 753 and 757), this stale allocation is written to the FileHeader as free_num_bytes
>  4. But FreeSegmentsData.serialize() writes only num_bytes of actual data (smaller)
>  5. On the next file open, update_existing reads free_num_bytes bytes, passes free_num_bytes - key_size as num_bytes to FreeSegmentsData.deserialize, then assert position == num_bytes fails because only the actual smaller slice
  entries were parsed
>
>  The fix is to use self._data.num_bytes (always freshly computed from current slices) instead of self._data.allocation (stale cache) in FreeSegments.write()

I went through the logic, and this reasoning makes sense and it indeed solves the issue.

This is the test file that I was using for this.

```python
import numpy as np
import hist
import uproot

n_histograms = 1_000
n_events = 10_000
n_tries = 10

for n in range(n_tries):
    print(f"Try {n+1}/{n_tries}")
    uproot.recreate("histograms.root")

    for i in range(n_histograms):
        #print(f"Creating histogram {i+1}/{n_histograms}")
        h = hist.Hist(
            hist.axis.Regular(50, 0, 5, name=f"x{i}", label=f"Variable {i}")
        )
        
        data = np.random.normal(loc=2.5, scale=0.7, size=n_events)
        h.fill(**{f"x{i}": data})
        
        with uproot.update("histograms.root") as f:
            f[f"hist_{i}"] = h
```
I won't add it to the test files since it takes a while to run, but I'll leave it here for future reference.


ianna (approved):
@ariostas - great catch! Thanks for fixing it.

PR #1603, opened 2026-03-17
fix: bug in num_entries to properly use regularize_files

nickwp:
There is a bug in `num_entries`, for example in the case where the path provided includes a glob returning multiple files.
The current code already uses `uproot._util.regularize_files` to handle this, and iterates over the resulting file paths and object paths in `paths2`, which should nicely handle all cases. But then it yields the contents of the original `paths` (after some extra manual processing to `paths` that should have already been handled by `regularize_files` anyway).

After removing that manual processing of `paths` and just using the result of `regularize_files` to provide the file_path and object_path, it should work in general.

This now works correctly in the case that broke before, i.e. providing a file path with a glob that results in multiple file paths from `regularize_files` but had just one original file path in `paths`. It should also work for any case that `uproot.iterate`, `uproot.concatenate` and `uproot.dask` work, since they're all relying on `regularize_files` in the same way.


ariostas (approved):
Thank you, @nickwp! This looks great. I just made a minor change renaming the test file to match the PR number.

ariostas:
@all-contributors please add @nickwp for code

PR #1604, opened 2026-03-17
fix: allow tuples and slightly more generic NumPy dtypes in mktree and mkrntuple

ariostas:
This PR fixed the regression introduced in #1515 where `mktree` stopped accepting tuples to specify a type specification. `mkrntuple` now also accepts these tuples, and in both cases the error messages for wrong types should be better.

Closes #1597


ianna (approved):
@ariostas - looks great! Thanks!

Issue #1528, opened 2025-11-28
(Potential) memory leak with uproot.iterate


pfackeldey:
(this is tested in uproot v5.6.5, likely present in other versions as well).

### The issue & reproducer

When running the following snippet and benchmarking this with memray:
```python
import uproot

def loop_iterate(rootfile):
    with uproot.open(rootfile, array_cache=None, object_cache=None) as f:
        tree = f["tree"]
        for batch in tree.iterate(step_size="200 MB", library="ak"):
            print(repr(batch))


if __name__ == "__main__":
    loop_iterate("~/Downloads/zlib9-jagged0.root")
```
I'm getting a memory consumption of the physical RAM usage (RSS) of up to 1.6GB (even though the step size is 200 MB): 

<img width="2238" height="1022" alt="Image" src="https://github.com/user-attachments/assets/21e36a12-2b34-4091-b9e6-6942741f8968" />

This is already surprising, however another indication that there's something up is that an explicit `gc.collect` at the end of each iteration improves the RSS situation by ~2x, i.e.:
```diff
import uproot
+import gc

def loop_iterate(rootfile):
    with uproot.open(rootfile, array_cache=None, object_cache=None) as f:
        tree = f["tree"]
        for batch in tree.iterate(step_size="200 MB", library="ak"):
            print(repr(batch))
+           del batch
+           gc.collect()


if __name__ == "__main__":
    loop_iterate("~/Downloads/zlib9-jagged0.root")
```
which gives up to 800MB RSS consumption:

<img width="2226" height="1002" alt="Image" src="https://github.com/user-attachments/assets/fe7eb61e-8dfb-4a6e-9231-81a3d084781f" />

### Why is this bad?

RSS is the physical RAM usage by this process, which dask monitors to decide if a worker should be killed due to OOM or not. 

### What I've found so far...

The memory usage grows by the following function: https://github.com/scikit-hep/uproot5/blob/main/src/uproot/behaviors/TBranch.py#L1440-L1452 and to be more explicit by this part of it: https://github.com/scikit-hep/uproot5/blob/main/src/uproot/behaviors/TBranch.py#L3421-L3428

What does work correctly is that the file `arrays` dictionary by the above function is ~200 MB, that's good! However, this `_ranges_or_baskets_to_arrays` still uses ~800 MB to fill the ~200 MB `arrays` dict and _does not free it again_.

[...]

PR #1606, opened 2026-03-17
fix: make step sizes of iterate more accurate

### LINKED ISSUE #1572: `step_size` of `iterate` method is not accurate
When you pass memory sizes into the various `iterate` methods you end up getting arrays with sizes that are quite different to the size that you requested. It happens for both TTrees and RNTuples, but it is much worse for RNTuples.

This was first reported by @pfackeldey when studying memory usage and fragmentation.

I'm looking into this first, and then follow up with a more careful look into the usage and fragmentation.

ariostas:
This PR fixes the accuracy of the step size of the `iterate` methods. This is in part what led to the discussion in #1528.

Here is a script that I used to compare the before and after.
```python
from pathlib import Path
import numpy as np
import awkward as ak
import uproot

filename = Path("tmp.root")
if not filename.exists():
    print("Creating test file...")
    with uproot.recreate(filename) as f:
        for i in range(200):
            print(f"Filling file with random data, iteration {i+1}/200", end="\r")
            data_x = []
            data_y = []
            for n in range(1000):
                data_x.append(np.random.rand(np.random.randint(1,100)))
                data_y.append(np.random.randint(1000, size=np.random.randint(1,100)))
            if i == 0:
                f.mktree("tree", {"x": data_x, "y": ak.Array(data_y)})
                f.mkrntuple("rntuple", {"x": data_x, "y": ak.Array(data_y)})
            else:
                f["tree"].extend({"x": data_x, "y": ak.Array(data_y)})
                f["rntuple"].extend({"x": data_x, "y": ak.Array(data_y)})
    print("\nTest file created.")

for step_size in ["1 MiB", "2 MiB", "5 MiB", "50 MiB", "100 MiB"]:
    with uproot.open(filename) as f:
        tree = f["tree"]
        rntuple = f["rntuple"]

        tree_array = next(tree.iterate(step_size=step_size))
        rntuple_array = next(rntuple.iterate(step_size=step_size))

        print(f"Requested step size:          {step_size}")
        print(f"Actual step size for TTree:   {tree_array.nbytes / 1_024**2:.2f} MiB")
        print(f"Actual step size for RNTuple: {rntuple_array.nbytes / 1_024**2:.2f} MiB")
        print("")
```
Note that you need 12e9a06e0286aa3623b0d439bd312b28d123485d for it to work since I prefer using MiB over MB.

Here is the output from before:
```
Requested step size:          1 MiB
Actual step size for TTree:   1.63 MiB
Actual step size for RNTuple: 2.31 MiB

Requested step size:          2 MiB
Actual step size for TTree:   3.23 MiB
Actual step size for RNTuple: 3.85 MiB

Requested step size:          5 MiB
Actual step size for TTree:   8.11 MiB
Actual step size for RNTuple: 8.53 MiB

Requested step size:          50 MiB
Actual step size for TTree:   81.41 MiB
Actual step size for RNTuple: 83.34 MiB

Requested step size:          100 MiB
Actual step size for TTree:   157.24 MiB
Actual step size for RNTuple: 155.71 MiB
```
And here's the output with the fix.
```
Requested step size:          1 MiB
Actual step size for TTree:   1.01 MiB
Actual step size for RNTuple: 1.55 MiB

Requested step size:          2 MiB
Actual step size for TTree:   2.02 MiB
Actual step size for RNTuple: 2.32 MiB

Requested step size:          5 MiB
Actual step size for TTree:   5.02 MiB
Actual step size for RNTuple: 5.42 MiB

Requested step size:          50 MiB
Actual step size for TTree:   50.43 MiB
Actual step size for RNTuple: 50.56 MiB

Requested step size:          100 MiB
Actual step size for TTree:   100.96 MiB
Actual step size for RNTuple: 100.45 MiB
```


ianna (approved):
@ariostas - As we've discussed at the meeting - this is great. Thanks!

Issue #1533, opened 2025-12-05
Writing profiles with storage=WeightedMean()


wiso:
This is similar to #1531 about the problem writing profiles. The following code is run on top of uproot 5.6.8 patched with the present proposed fix in #1532 and hist 2.9.0

```python
import uproot
import hist

h = hist.Hist(
    hist.axis.IntCategory([10, 20, 30]),
    storage=hist.storage.WeightedMean())

with uproot.recreate("test.root") as fout:
    fout['h'] = h
```

```
/home/turra/micromamba/envs/mamba-python3.12/lib/python3.12/site-packages/uproot/writing/identify.py:272: FutureWarning: .metadata was not set, returning None instead of Attribute error, boost-histogram 1.7+ will error.
  if obj.metadata is not None and "fSumw2" in obj.metadata.keys():
Traceback (most recent call last):
  File "/home/turra/test-histo.py", line 9, in <module>
    fout['h'] = h
    ~~~~^^^^^
  File "/home/turra/micromamba/envs/mamba-python3.12/lib/python3.12/site-packages/uproot/writing/writable.py", line 986, in __setitem__
    self.update({where: what})
  File "/home/turra/micromamba/envs/mamba-python3.12/lib/python3.12/site-packages/uproot/writing/writable.py", line 1598, in update
    uproot.writing.identify.add_to_directory(v, name, directory, streamers)
  File "/home/turra/micromamba/envs/mamba-python3.12/lib/python3.12/site-packages/uproot/writing/identify.py", line 79, in add_to_directory
    writable = to_writable(obj)
               ^^^^^^^^^^^^^^^^
  File "/home/turra/micromamba/envs/mamba-python3.12/lib/python3.12/site-packages/uproot/writing/identify.py", line 283, in to_writable
    fSumw2 = obj.view()["count"]
             ~~~~~~~~~~^^^^^^^^^
  File "/home/turra/micromamba/envs/mamba-python3.12/lib/python3.12/site-packages/boost_histogram/view.py", line 20, in __getitem__
    sliced = super().__getitem__(ind)
             ^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: no field of name count
```

ariostas:
Thank you, @wiso! This is now also fixed in #1532

wiso:
I think this should be reopen, see my comment in https://github.com/scikit-hep/uproot5/pull/1532

PR #1532, opened 2025-12-03
fix: writing TProfiles from hist

ariostas:
This PR adds some flexibility to the `hist` profiles that can be written to file. The old code was assuming ROOT histograms that were converted to `hist` with `to_hist`, which adds some metadata, but if the metadata was not present it just didn't work. Someone should double-check that I'm inputting the right data into `to_TProfile`. Closes #1531.

Issue #1531, opened 2025-12-03
Writing profiles


wiso:
From https://github.com/scikit-hep/uproot5/pull/1000 I undestand writing profile is supported.

```python
import uproot
import hist

f = uproot.recreate("test.root")
h1 = hist.Hist.new.Reg(100, 0, 10).Mean()
f['h1'] = h1
```

gives

```
File ~/micromamba/envs/mamba-python3.11/lib/python3.11/site-packages/uproot/writing/writable.py:984, in WritableDirectory.__setitem__(self, where, what)
    982 if self._file.sink.closed:
    983     raise ValueError("cannot write data to a closed file")
--> 984 self.update({where: what})

File ~/micromamba/envs/mamba-python3.11/lib/python3.11/site-packages/uproot/writing/writable.py:1576, in WritableDirectory.update(self, pairs, **more_pairs)
   1573     for item in path:
   1574         directory = directory[item]
-> 1576     uproot.writing.identify.add_to_directory(v, name, directory, streamers)
   1578 self._file._cascading.streamers.update_streamers(self._file.sink, streamers)

File ~/micromamba/envs/mamba-python3.11/lib/python3.11/site-packages/uproot/writing/identify.py:169, in add_to_directory(obj, name, directory, streamers)
    166     tree.extend(data)
    168 else:
--> 169     writable = to_writable(obj)
    171     for rawstreamer in writable.class_rawstreamers:
    172         if isinstance(rawstreamer, tuple):

File ~/micromamba/envs/mamba-python3.11/lib/python3.11/site-packages/uproot/writing/identify.py:362, in to_writable(obj)
    360 if obj.kind == "MEAN":
    361     if hasattr(obj, "storage_type"):
--> 362         if "fSumw2" in obj.metadata.keys():
    363             fSumw2 = obj.metadata["fSumw2"]
    364         else:

AttributeError: 'NoneType' object has no attribute 'keys'
```

```python
uproot.__version__ 5.6.6'
hist.__version__ 2.9.0
```

ariostas:
Thanks for the report! I'll look into it.

wiso:
@ianna it seems not. I replied on https://github.com/scikit-hep/uproot5/pull/1532

PR #1609, opened 2026-03-26
fix: TProfile reading and writing logic

ariostas:
This PR aims to fix the IO logic for TProfiles. These issues were brought up in #1531 and #1533.

The fixes here were made primarily made with Claude, with some manual tweaking afterwards. I did extensive questioning and I went through the changes carefully, and I ended up convincing myself that they're right. But I'd appreciate if someone else who is more familiar with TProfiles can also take a look.

@wiso, would you be able to have a look? I'll add the example you posted in #1532 as a test.


ariostas:
@wiso, would you be able to check if this is working correctly now?

ianna (approved):
@ariostas - great! Code coverage is better and though it is not at the required (high-ish) level yet I'll merge it. Thanks!

PR #1615, opened 2026-04-09
chore(ci): resolve issues with GPU CI

ariostas:
`cccl-python` and `kvikio` recently got updated and broke the CI, so I'm checking how to keep the CI running while we figure out how to fix it.


ariostas:
Turns out that we just needed to drop the nvidia conda channel. I forgot to drop it when we did it for Awkward. I'll merge this PR without a review since it's very minor, so that the CI works for other PRs.

PR #1619, opened 2026-04-13
fix: issues with freethreading and include Python 3.14t in CI

### LINKED ISSUE #1617: Add python 3.14t testing
Uproot is not currently being tested in python 3.14t. And there are failures that one can see locally (or see https://github.com/scikit-hep/integration-tests/pull/19).
We should add testing for free threaded python.

ariostas:
This PR adds 3.14t to the CI matrix. I also removed intermediate versions to keep the number of jobs now, since they shouldn't really be necessary.

AI disclosure: I used assistance from Gemini for this PR.

### DIFF
--- a/src/uproot/source/fsspec.py
+++ b/src/uproot/source/fsspec.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import concurrent.futures
+import contextlib
 import queue
 import re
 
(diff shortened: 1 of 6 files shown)


src/uproot/source/fsspec.py
```diff
@@ -90,6 +94,19 @@ def __enter__(self):
...
+            self._open_file.__exit__(exception_type, exception_value, traceback)
+
+            # Workaround for leaking fsspec filesystems (like ZipFileSystem/TarFileSystem)
+            # that don't correctly close their underlying file objects.
+            if hasattr(self._fs, "close"):
+                with contextlib.suppress(Exception):
+                    self._fs.close()
+            if hasattr(self._fs, "of") and hasattr(self._fs.of, "__exit__"):
+                with contextlib.suppress(Exception):
+                    self._fs.of.__exit__(None, None, None)
```
ariostas, line 107: There are some issues with `fsspec` with freethreading, but this workaround works for now. I'll open a PR to `fsspec` to fix this upstream.

ikrommyd:
@ariostas how do you want to handle this given the fsspec PR you did?

ariostas:
I was thinking of merging this with the workaround and after some time put a lower bound for fsspec. What do you think?

ikrommyd:
yeah I agree, this should be good to go once you fix the merge conflict.

ikrommyd:
I'll add uproot to the 3.14t integration tests.

ianna (approved):
@ariostas - great! Agree, it’s a good solution. We cannot rely on a file handle being closed when it goes out of scope anymore. Thanks!

ikrommyd (approved):

ikrommyd:
@ariostas you probably want to make codecov not required in uproot too (like awkward)

ikrommyd (approved):

PR #1622, opened 2026-04-21
fix: FreeSegments end calculation when crossing 2 GB boundary

### LINKED ISSUE #1620: Extending tree beyond 1.86 GB causes exception since v5.7.3
I am trying to write NumPy arrays into a ROOT file with `extend()`, but every time the total size exceeds 1.86 GB, I get the following expectation:
```python
  File "[...]\uproot\writing\writable.py", line 1899, in extend
    self._cascading.extend(self._file, self._file.sink, data)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[...]\uproot\writing\_cascadetree.py", line 489, in extend
    self.write_anew(sink)
    ~~~~~~~~~~~~~~~^^^^^^
  File "[...]\uproot\writing\_cascadetree.py", line 1199, in write_anew
    self._key = self._directory.add_object(
                ~~~~~~~~~~~~~~~~~~~~~~~~~~^
        sink,
        ^^^^^
    ...<6 lines>...
        big=True,
        ^^^^^^^^^
    )
    ^
  File "[...]\uproot\writing\_cascade.py", line 1557, in add_object
    location = self._freesegments.allocate(requested_bytes, dry_run=False)
  File "[...]\uproot\writing\_cascade.py", line 657, in allocate
    if self.at_end:
       ^^^^^^^^^^^
  File "[...]\uproot\writing\_cascade.py", line 633, in at_end
    assert end_of_record <= self._data.end
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

This behavior only occurs in version 5.7.3 and not in 5.7.2 or earlier with the same dependency versions. I can also reproduce it with Python 3.14.4 on Windows 11 and Python 3.13.12 on Fedora 42. It is also independent of the number of branches or number of events written per `extend()`.

Here is a minimal example for reproduction:
```python
import numpy as np
import uproot


data = np.random.default_rng().uniform(size=300_000_000)

with uproot.recreate("test.root") as file:

    file.mktree("tree", {"branch": np.float64})

    for start in range(0, data.shape[0], 30_000_000):
        file["tree"].extend({"branch": data[start : start + 30_000_000]})
```

ariostas:
This PR fixes #1620. It turned out that #1601 caused an issue where `self._data.end` was set expecting the file to stay below 2GB, but then when writing it would realize it needed to switch to using a big file pointer, so there was a mismatch. The `required_end` takes into account whether it will have to be extended into a big file.

AI disclosure: I used assistance from Gemini for this PR, but I fully verified the changes and the logic.

### DIFF
--- a/src/uproot/writing/_cascade.py
+++ b/src/uproot/writing/_cascade.py
@@ -521,6 +521,20 @@ class FreeSegmentsData(CascadeLeaf):
 
         return total
 
+    def required_end(self, data_location):
+        total = 0
+        for _, stop in self._slices:
+            if stop - 1 >= uproot.const.kStartBigFile:
+                total += _free_format_big.size
+            else:
+                total += _free_format_small.size
+
+        tentative_end = data_location + total + _free_format_small.size
+        if tentative_end < uproot.const.kStartBigFile:
+            return tentative_end
+        else:
+            return data_location + total + _free_format_big.size
+
     def serialize(self):
         pairs = []
         for start, stop in self._slices:
@@ -660,8 +674,8 @@ class FreeSegments(CascadeNode):
             out = self._key.location
             if not dry_run:
                 self._key.location = self._key.location + num_bytes
-                self._data.end = (
-                    self._key.location + self._key.allocation + self._data.allocation
+                self._data.end = self._data.required_end(
+                    self._key.location + self._key.num_bytes
                 )
             return out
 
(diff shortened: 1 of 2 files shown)


ianna (approved):
@ariostas - Great! Thanks for fixing it. The coverage target is high - which is the only thing that is not green. I'll merge it. Thanks.

PR #1624, opened 2026-04-27
docs: update to reflect that RNTuple is the default writing format

ariostas:
We switched to writing RNTuples by default back in v5.7.0, but some parts of the documentation still mentioned this as being a future change. I updated them to reflect that the change already happened.

AI disclosure: I used Gemini to find the places in the docs that needed to be updated.


ianna (approved):
@ariostas - thanks for updating it!

PR #1633, opened 2026-05-19
ci: pin micromamba version and bump Pyodide version

ariostas:
We'll have to pin the micromamba version that the CI uses because the latest release is broken on Windows. (See https://github.com/mamba-org/setup-micromamba/issues/306 and https://github.com/mamba-org/micromamba-releases/issues/97)

Also, the Pyodide build started failing, but simply bumping the version fixed it.


ianna (approved):
@ariostas - thanks for following up on this!

PR #1641, opened 2026-06-04
feat: support path objects with storage_options in uproot.open

kratsg:
## Summary

- Allows path objects with a `storage_options` attribute (e.g. `UPath` from `universal-pathlib`) to be passed directly to `uproot.open()`
- `storage_options` are merged into the kwargs forwarded to `FSSpecSource`/fsspec, so auth headers, Bearer tokens, and macaroons can be bundled into the path object rather than passed as separate kwargs every call
- Explicit kwargs to `uproot.open()` still take precedence over path-embedded options

```python
from upath import UPath

f = UPath("https://storage.example.com/file.root", headers={"Authorization": f"Bearer {token}"})
uproot.open(f)  # no separate headers= needed
```

/cc @ponyisi

### DIFF
--- a/src/uproot/reading.py
+++ b/src/uproot/reading.py
@@ -130,16 +130,17 @@ def open(
         file_path = path
         object_path = None
 
-    file_path = uproot._util.regularize_path(file_path)
-
     # Support path objects that carry fsspec storage_options (e.g. UPath from
-    # universal-pathlib). Caller-provided kwargs take precedence over the
-    # path-embedded options so that one-off overrides still work.
+    # universal-pathlib). Extract before regularize_path so that if the object
+    # ever gains __fspath__ the options aren't lost when it is converted to str.
+    # Caller-provided kwargs take precedence over path-embedded options.
     if hasattr(file_path, "storage_options") and not isinstance(
         file_path, (str, bytes)
     ):
         options = {**file_path.storage_options, **options}
         file_path = str(file_path)
+
+    file_path = uproot._util.regularize_path(file_path)
 
     if not isinstance(file_path, str) and not (
         hasattr(file_path, "read") and hasattr(file_path, "seek")
--- a/src/uproot/reading.py
+++ b/src/uproot/reading.py
@@ -132,6 +132,15 @@ def open(
 
     file_path = uproot._util.regularize_path(file_path)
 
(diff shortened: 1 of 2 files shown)


kratsg:
@ariostas @ianna this came up in the `servicex` channel in the [IRIS-HEP slack](https://iris-hep.slack.com/archives/CJH870SR2/p1780513701084959?thread_ts=1779593639.482559&cid=CJH870SR2). I think there will need to be a larger refactoring of how uproot handles paths to make this a little nicer -- rather than adding this sort-of duck-typing one-off *but* for now, this is the smallest change I can make to support the `upath` case that @ponyisi brought up.

ianna (approved):
@kratsg - This is great! Thanks for implementing it.

kratsg:
> @kratsg - This is great! Thanks for implementing it.

no rpoblem. Long term, I think I agree with @ponyisi in that we should implement `upath` support in `uproot` to remove some of the additional overhead that is incurred at the moment.

src/uproot/reading.py
```diff
@@ -132,6 +132,15 @@ def open(
...
 
+    # Support path objects that carry fsspec storage_options (e.g. UPath from
+    # universal-pathlib). Caller-provided kwargs take precedence over the
+    # path-embedded options so that one-off overrides still work.
+    if hasattr(file_path, "storage_options") and not isinstance(
+        file_path, (str, bytes)
+    ):
+        options = {**file_path.storage_options, **options}
+        file_path = str(file_path)
+
```
ariostas: Thank you @kratsg and sorry for the delay in reviewing this. I think it would be safer if we move this new block above the `regularize_path` above so that if `UPath` (or something else) ends up implementing `__fspath__` it doesn't end up discarding the storage options,

kratsg: `UPath` does define `__fspath__` which actually is just what breaks things in the first place...

ariostas: `UPath` defines `__vfspath__`, not `__fspath__`. If it defined, `__fspath__` then the `regularize_path` function would convert it into a string already. So that's why I think it's safer to do the `regularize_path` afterwards.

ariostas (approved):
Thank you @kratsg!

Issue #1502, opened 2025-10-01
`uproot.dask` cannot handle grouped `TObject` properly


mrzimu:
Hi,
I'm trying to read a simple object derived from `TObject` through `uproot.dask`.  It seems that `uproot.dask` cannot handle branch relationship between  `TObject` and `TObject.fUniqueID`, `TObject.fBits`.

When I read the whole tree:

```python
import uproot
dak_arr = uproot.dask({"demo-data.root" : "/my_tree/basic_types"})
dak_arr.compute()
```

The exception is:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mrli/work/uproot5/.venv/lib/python3.12/site-packages/dask/base.py", line 370, in compute
    (result,) = compute(self, traverse=False, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mrli/work/uproot5/.venv/lib/python3.12/site-packages/dask/base.py", line 656, in compute
    results = schedule(dsk, keys, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mrli/work/uproot5/src/uproot/_dask.py", line 1261, in __call__
    result, _ = self._call_impl(i, start, stop)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mrli/work/uproot5/src/uproot/_dask.py", line 1265, in _call_impl
    return self.read_tree(
           ^^^^^^^^^^^^^^^
  File "/home/mrli/work/uproot5/src/uproot/_dask.py", line 1016, in read_tree
    mapping = self.form_mapping_info.load_buffers(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mrli/work/uproot5/src/uproot/_dask.py", line 969, in load_buffers
    assert src_dtype == dst_dtype  # Sanity check!
           ^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

I simply checked the loop this assertion in:

https://github.com/scikit-hep/uproot5/blob/65b4af60f506d2d47432aa31621761ef76e1ecea/src/uproot/_dask.py#L957-L970

[...]

PR #1642, opened 2026-06-07
fix: `uproot.dask` cannot handle grouped `TObject` properly

mrzimu:
Hi @ariostas @ikrommyd ,

I tried to have an LLM fix issue #1502 , and it made the following changes. It seems like all the tests pass and the issue is resolved, but the changes are quite extensive, so I don’t feel very confident reviewing them.

I’ve written down the LLM’s understanding of the bug and the fix, which might be helpful:
- [dask.md](https://github.com/user-attachments/files/28687095/dask.md) which fixes the but in `uproot.dask`
- [virtual.md](https://github.com/user-attachments/files/28687097/virtual.md) which fixes the similar bug when using awkward virtual array (`.arrays(virtual=True)`)

I’d like to know what you think of the LLM’s fix — is it something we could consider improving the tests for and then merge, or should we set it aside for now and conduct a more detailed review later?

### DIFF
--- a/tests/test_1642_read_AsGrouped_with_dask_or_virtual_array.py
+++ b/tests/test_1642_read_AsGrouped_with_dask_or_virtual_array.py
@@ -1,0 +1,27 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot5/blob/main/LICENSE
+
+import pytest
+import skhep_testdata
+import awkward
+
+import uproot
+
+dask = pytest.importorskip("dask")
+da = pytest.importorskip("dask.array")
+
+
+def test_read_AsGrouped_with_dask_or_virtual_array():
+    """
+    ``RNTuple.iterate(report=True)`` must yield ``(arrays, Report)`` pairs,
+    matching the contract documented in its docstring and already provided
+    by ``TTree.iterate``.
+    """
+    path = skhep_testdata.data_path("uproot-issue-1502.root")
+    with uproot.open(path) as f:
+        arr = f["tree"].arrays()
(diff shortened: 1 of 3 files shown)


ariostas:
Thank you, @mrzimu! The changes look good to me. Could you open a PR in [scikit-hep-testdata](https://github.com/scikit-hep/scikit-hep-testdata) to add your test file from #1502 file, and then add a couple of tests here?

mrzimu:
> Thank you, @mrzimu! The changes look good to me. Could you open a PR in [scikit-hep-testdata](https://github.com/scikit-hep/scikit-hep-testdata) to add your test file from #1502 file, and then add a couple of tests here?

Sure, I'll add them

tests/test_1642_read_AsGrouped_with_dask_or_virtual_array.py
```diff
@@ -0,0 +1,27 @@
...
+    ``RNTuple.iterate(report=True)`` must yield ``(arrays, Report)`` pairs,
+    matching the contract documented in its docstring and already provided
+    by ``TTree.iterate``.
+    """
+    path = skhep_testdata.data_path("uproot-issue-1502.root")
+    with uproot.open(path) as f:
+        arr = f["tree"].arrays()
+        virtual_arr = f["tree"].arrays(virtual=True)
+
+    assert awkward.array_equal(arr, awkward.materialize(virtual_arr))
```
ariostas:
```suggestion
        assert awkward.array_equal(arr, awkward.materialize(virtual_arr))
```
To make sure that the file is still open

ariostas (approved):
Thank you, @mrzimu!

PR #1644, opened 2026-06-08
chore: improve streamer reading

ariostas:
There were a couple of places where things could be done a bit better.


ianna (approved):
@ariostas - this is great! Thanks for fixing it.

PR #1647, opened 2026-06-10
chore: add an AGENTS file

henryiii:
If anyone has anything else they are using, feel free to add it to this. This is a default output from Claude Fable 5.

Assisted-by: ClaudeCode:claude-fable-5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,11 @@ skhep_testdata/
 # Pyodide
 .pyodide*
 dist-pyodide/
+
+# Symlink to AGENTS.md
+CLAUDE.md
+.claude/
+
+# Lockfiles
+uv.lock
+*pylock*
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,48 @@
+# AGENTS.md
+
+This file provides guidance to agents when working with code in this repository.
+
+## What this is
+
+Uproot is a library for reading and writing ROOT files (the CERN HEP data format) in pure Python and NumPy — an I/O library only, with no C++ ROOT dependency.
+
+## Commands
+
+```bash
+uv pip install -e . --group=dev      # editable install with dev dependencies
+
+uv run pytest tests                  # full test suite
+uv run pytest tests/test_1603_num_entries.py            # one test file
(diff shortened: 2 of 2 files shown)


henryiii:

The CI failure is **unrelated to your PR's changes** — it's an environmental/dependency error on one runner.

## What failed

Exactly one job failed: **"Test on windows-latest with Python 3.10"**. Every other job (20+ across Linux/macOS/Windows, GPU, Pyodide, free-threading) passed. The "Lint PR" workflow passed too.

## Why

pytest never ran a single test — it crashed during **collection**, while *importing* three test modules:

```
ERROR tests/test_0916_read_from_s3.py    - ssl.SSLError: [ASN1: NOT_ENOUGH_DATA] not enough data (_ssl.c:4040)
ERROR tests/test_1063_dask_distributed.py - ssl.SSLError: [ASN1: NOT_ENOUGH_DATA] not enough data (_ssl.c:4040)
ERROR tests/test_1085_dask_write.py      - ssl.SSLError: [ASN1: NOT_ENOUGH_DATA] not enough data (_ssl.c:4040)
Interrupted: 3 errors during collection
Process completed with exit code 2
```

The `SSLError [ASN1: NOT_ENOUGH_DATA]` is raised deep inside an import chain (`s3fs` → `aiobotocore` → `botocore` → `urllib3`/`ssl`). It comes from OpenSSL failing to parse certificate/ASN.1 data while a transitive dependency loads at import time. This is a **broken/corrupt OpenSSL or cert bundle in that one Windows micromamba environment**, not a code error.

## This is not caused by your changes

Your PR (#1647) only adds an `AGENTS.md` file — a docs-only change with zero impact on imports, SSL, or dependencies. The "Check required tests" job then went red simply because it depends on that one failing job.

## What to do

This is a flaky environment failure. Just **re-run the failed job**:

```bash
gh run rerun 27296346887 --failed
```

It should pass on a fresh runner. If it reproduces consistently, the underlying issue is in the windows-py3.10 dependency stack (`s3fs`/`aiobotocore`/OpenSSL), which is worth filing separately — but it's not something to fix in this PR.

Issue #1646, opened 2026-06-10
Claude Fable AI review on uproot


henryiii:
:robot: _AI text below_ :robot:

Review complete. Seven parallel agents covered all of `src/uproot/` (~56k lines); the high-severity findings below were verified by actually running the code (by the agents, and I re-confirmed a sample myself), not just by reading it. Here's the synthesis, ordered by impact.

# Critical — silent wrong data or file corruption

These return wrong results or write bad files for valid user input, with no error:

1. **`behaviors/TBranch.py:3053-3061`** — `_regularize_interpretation` builds the big-endian dtype and then throws it away: `return AsDtype(interpretation)` uses the original argument instead of `dtype`. `tree["px1"].array(interpretation="f8")` returns byte-swapped garbage (confirmed: `1.27e+239` instead of `-41.2`). Same path via `arrays([("name", "f8")])`.
2. **`interpretation/numerical.py:111`** — `output.view(dtype.newbyteorder("="))` *reinterprets* bytes instead of converting; any non-native `to_dtype` (e.g. `AsDtype(">f8", ">f8")`) returns garbage. Confirmed end-to-end.
3. **`writing/_cascadetree.py:646-685`** — writing a **string branch from a sliced awkward array** (`offsets[0] != 0`) produces an unreadable basket; the string path never normalizes offsets the way the jagged path does at line 728. Reproduced: file written, then fails to read back.
4. **`writing/interpret.py:141-151`** — TGraph `fMinimum`/`fMaximum` computed from **x** instead of y (`np.min(x)`/`np.max(x)`). Written graphs get a wildly wrong Y-axis range in ROOT.
5. **`compression.py:560-561`** — when a 16 MB block is incompressible, the compressed size can exceed `2**24-1` and the 3-byte header silently drops the high byte → corrupt file. ROOT stores such blocks uncompressed; uproot should too (or raise).
6. **`behaviors/TBranch.py:431-438`** (and the same code in `behaviors/RNTuple.py:401-408`) — `concatenate(..., allow_missing=True)`: the `continue` path skips `global_start = global_stop`, so every file after a skipped one is read at shifted entry offsets. Confirmed: returns entries outside the requested range.
7. **`behaviors/RNTuple.py:1018-1030`** — `iterate` doesn't clamp the per-step stop to the requested `entry_stop`; `iterate(step_size=30000, entry_stop=40000)` on 50k entries yields 50k. Confirmed.
[...]

PR #1648, opened 2026-06-10
fix: interpretation dtype handling returned byte-swapped data

henryiii:

## Summary

Three related bugs caused user-supplied `interpretation=` specs to return byte-swapped garbage, and `!=` between interpretations to crash.

### What was broken

Raw `TBasket` data is big-endian. Several code paths mishandled byte order:

1. **`_regularize_interpretation` (`behaviors/TBranch.py`)** — the else branch computed a big-endian `dtype` but then returned `AsDtype(interpretation)` using the **original** argument, discarding the conversion. The separate `numpy.dtype` branch had the same problem.
2. **`AsDtype.final_array` (`interpretation/numerical.py`)** — `output = output.view(output.dtype.newbyteorder("="))` **reinterpreted** bytes instead of converting values. A no-op for native `to_dtype`, but corrupted values for a user-supplied big-endian `to_dtype`.
3. **`Interpretation.__ne__` (`interpretation/__init__.py`)** — `raise not self == other` raised `TypeError: exceptions must derive from BaseException` for any `!=`.

### Minimal reproducer

```python
import uproot, skhep_testdata, numpy
import uproot.interpretation.numerical as N

t = uproot.open(skhep_testdata.data_path("uproot-Zmumu.root"))["events"]

# before: [1.27e+239, ...]   after: [-41.195, 35.118, 35.118]
t["px1"].array(interpretation="f8", library="np")[:3]
t["px1"].array(interpretation=numpy.dtype("f8"), library="np")[:3]
t["px1"].array(interpretation=N.AsDtype(">f8", ">f8"), library="np")[:3]

N.AsDtype("f8") != N.AsDtype("f8")   # before: TypeError   after: False
```

### What changed

1. `_regularize_interpretation` now returns `AsDtype(dtype)` with the big-endian-converted dtype, and both the `numpy.dtype` and string-spec cases go through the same branch for consistency.
2. `final_array` converts a non-native `to_dtype` output to native byte order by value (`astype`) rather than reinterpreting bytes. For native `to_dtype` (the default) it leaves `output` untouched, which preserves the in-place buffer used by `AsDtypeInPlace`.
3. Deleted the broken `__ne__`; Python 3 negates `__eq__` automatically.

### Testing

Relevant existing suites pass (`test_0017`, `test_0018`, `test_0034`, `test_0487`, `test_1477`, etc.). Nothing skipped.

Part of #1646.

### DIFF
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -3053,12 +3053,10 @@ def _regularize_aliases(hasbranches, aliases):
 def _regularize_interpretation(interpretation):
     if isinstance(interpretation, uproot.interpretation.Interpretation):
         return interpretation
-    elif isinstance(interpretation, numpy.dtype):
-        return uproot.interpretation.numerical.AsDtype(interpretation)
     else:
         dtype = numpy.dtype(interpretation)
         dtype = dtype.newbyteorder(">")
-        return uproot.interpretation.numerical.AsDtype(interpretation)
+        return uproot.interpretation.numerical.AsDtype(dtype)
 
 
 def _regularize_branchname(
--- a/src/uproot/interpretation/__init__.py
+++ b/src/uproot/interpretation/__init__.py
@@ -176,9 +176,6 @@ class Interpretation:
     def __eq__(self, other):
         raise AssertionError
 
-    def __ne__(self, other):
-        raise not self == other
-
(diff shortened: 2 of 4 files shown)


ariostas (approved):
Looks good! I just simplified the tests a bit since multiple of them were pretty similar.

ianna (approved):
@henryiii and @ariostas -- Thanks!

PR #1653, opened 2026-06-10
fix: HTTP source multipart batching, fallback leak, and close behavior

henryiii:

Part of #1646.

## What was broken

All in `src/uproot/source/http.py` (plus one mirror in `xrootd.py`), found by code-reading + simulation:

1. **`HTTPSource.chunks` range-batching** (lines ~672-701). When the `Range` header overflowed more than once, the loop submitted the wrong slice: with `j` the start of the current batch and `i` the running index, it used `ranges[j:j+i]` instead of `ranges[j:i]`. The futures/results were built for a *superset* of the ranges actually in the header, so `handle_multipart` ran out of parts with futures still pending, then `handle_no_multipart` re-ran completed futures (double-posting some chunks to the notifications queue while others were lost). When the overflow fell on the **last** range, `last_batch_appended=True` made the final guard skip the pending batch entirely, so those chunks were never fetched and the caller blocked forever on `notifications.get()`.

2. **`_set_fallback`** unconditionally created a new `MultithreadedHTTPSource` every call. Against a server without multipart support, each batch leaked the previous fallback's daemon worker threads (blocked on `queue.get` forever).

3. **`HTTPSource.__exit__` / `XRootDSource.__exit__`** called `self._executor.shutdown()` instead of `self._executor.__exit__(...)`. `ResourceThreadPoolExecutor.shutdown` (inherited from `ThreadPoolExecutor`) only drains workers; only `__exit__` sets `_closed = True`. So `source.closed` stayed `False` after close, and a subsequent `chunk()` enqueued onto a dead pool and hung instead of raising `OSError`. `MultithreadedSource.__exit__` already does it the right way.

4. **`next_header`** assumed `Content-Range` was the last header before the blank line (it read exactly one more line after matching). A server emitting another header after `Content-Range` silently misaligned the part payload.

5. **`handle_multipart`** had an unused `num_found` counter and a garbled `HTTPException` message (stray quote chars: `... {range_string.decode()!r} "` / `"in HTTP multipart`).

## Evidence

- A standalone simulation of the original batching loop on 8 small ranges with a tiny header limit produced batches `[(0,5),(10,15)]`, `[(20,25),(30,35),(40,45),(50,55)]` (4 ranges) with header `bytes= 20-24, 30-34` (only 2 ranges) — superset mismatch — and a final empty batch `[]` while its header listed `60-64, 70-74` (lost ranges).
- A unit test driving `HTTPSource.chunks` with a recording fake executor/resource asserts that every batch's `Range` header lists exactly its batch ranges and that the batches together cover every input range exactly once. It **fails** on the old code and passes on the fix.

## What changed

- Rewrote the batching loop to track `batch_start`; when adding range `i` would overflow, submit `ranges[batch_start:i]` and start a new batch at `i`; after the loop, submit `ranges[batch_start:]` unconditionally. Removed the `last_batch_appended` machinery. Also fixed the new-batch header to drop the leading `", "` cleanly (was `bytes= 20-...`).
- Guarded `_set_fallback` with a lock so the fallback is created at most once.
- `HTTPSource.__exit__` and `XRootDSource.__exit__` now call `self._executor.__exit__(...)`.
- `next_header` now reads all part headers up to (and including) the blank separator, skipping leading blank/boundary lines.
- Removed `num_found`; fixed the exception message text.

Regression tests are added in a follow-up commit (`tests/test_<PR>_...py`): a pure-unit batching/coverage test with a mocked submit, a close-then-chunk test that asserts `closed` becomes `True` and a later `chunk()` raises `OSError`, and a `next_header` test with an extra header after `Content-Range`.

## Skipped

Nothing skipped — all five findings held up and are fixed.

### DIFF
--- a/src/uproot/source/http.py
+++ b/src/uproot/source/http.py
@@ -22,6 +22,7 @@ import queue
 import re
 import socket
 import sys
+import threading
 import urllib.parse
 from urllib.parse import urlparse
 
@@ -433,10 +434,8 @@ for URL {source.file_path}"""
 
         original_futures = dict(futures)
 
-        num_found = 0
         while len(futures) > 0:
             range_string, _size = self.next_header(response_buffer)
-            num_found += 1
             if range_string is None:
                 self.handle_no_multipart(source, ranges, original_futures, results)
                 return
@@ -450,8 +449,7 @@ for URL {source.file_path}"""
 
             if len(data) != length:
                 raise http.client.HTTPException(
-                    f"""wrong chunk length {len(data)} (expected {length}) for byte range {range_string.decode()!r} "
-                    "in HTTP multipart
+                    f"""wrong chunk length {len(data)} (expected {length}) for byte range {range_string.decode()!r} in HTTP multipart
 for URL {self._file_path}"""
                 )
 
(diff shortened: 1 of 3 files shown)


ariostas (approved):
I had to revert the xrootd change because it broke things (and added a comment explaining why). The rest looks good.

PR #1655, opened 2026-06-10
fix: model member-name and AwkwardForth-API bugs in TTree v18, TTime, TAtt

henryiii:

Part of #1646.

## Summary

Five model-level bugs fixed:

- **TTree v18** (`src/uproot/models/TTree.py`): `Model_TTree_v18` defined a property called `member_values` — a name that exists nowhere else and is never looked up. Every sibling version (v16, v17, v19, v20) defines `member_names`. The result was that `member_names` silently returned the base-class empty list `[]` for v18 trees. Renamed to `member_names`.

- **TTime** (`src/uproot/models/TTime.py`): `Model_TTime_v2.read_members` used a long-removed ForthGenerator API (`get_gen_obj()`, `get_keys()`, `add_to_header()`, `should_add_form()`, `add_form_key()`, `add_to_pre()`, `add_form()`, old `add_node(...)` signature). None of these exist on the current `ForthGenerator` / `Node` API in `_awkwardforth.py`. Any Forth-path read of a TTime-containing branch would raise `AttributeError`. Ported to the current `Node`-based API, mirroring `TDatime.py` (same structure: one big-endian fixed-width integer field, no versioning header).

- **TAtt** (`src/uproot/models/TAtt.py`): Typo `"fMarkserSize"` in `Model_TAttMarker_v2.member_names` — corrected to `"fMarkerSize"` (consistent with the field name used in `read_members`, `strided_interpretation`, `awkward_form`, and `_serialize`).

- **TTable** (`src/uproot/models/TTable.py`): Removed dead module-level `format = {...}` dict that (a) was never used anywhere — only the adjacent `_dtype` dict is used — and (b) shadowed the Python builtin `format`. Also updated the module docstring, which incorrectly referred to `TTree` instead of `TTable`.

- **TMatrixT** (`src/uproot/models/TMatrixT.py`): Module docstring referred to `TLeaf and its subclasses`; corrected to `TMatrixT`.

## Test plan

- [x] `uv run pytest tests/test_1647_model_fixes.py -v` — 5 new regression tests, all pass
- [x] `uv run pytest tests/test_0018_array_fetching_interface.py tests/test_0033_more_interpretations_2.py tests/test_0014_all_ttree_versions.py tests/test_0023_ttree_versions.py tests/test_0418_read_TTable.py tests/test_1610_read_TMatrixTSym_from_ttree.py tests/test_0011_generate_classes_from_streamers.py -q` — all pass
- [x] `prek -a --quiet` — clean

## Skipped / not applicable

- TGraph.py and TH.py docstrings were explicitly excluded (generated files).
- No ROOT-file-level Forth-path test for TTime: no TTime-containing ROOT test file is available in `skhep_testdata`; the fix is verified via source inspection and API-level assertions.


ariostas (approved):
Removed the tests since they were not particularly useful

PR #1656, opened 2026-06-10
fix: entry-range arithmetic in iterate/concatenate and basket selection

henryiii:

This PR fixes three coupled entry-range arithmetic bugs in `iterate`, `concatenate`, and basket selection.

## 1. Off-by-one extra basket on exact `entry_stop` boundaries

When `entry_stop` landed exactly on a basket boundary, the basket collector read, decompressed, and interpreted one extra basket contributing zero entries.

Reproducer (3 baskets of 10 entries, offsets `[0,10,20,30]`):
```python
b.entries_to_ranges_or_baskets(0, 10)  # selected baskets [0, 1], should be [0]
```

The condition `start <= entry_stop` in `entries_to_ranges_or_baskets` and `_hasbranches_num_entries_for` (`src/uproot/behaviors/TBranch.py`) selected a basket whose `start == entry_stop`. The matching per-basket trim loops in the interpretations (`numerical.py`, `jagged.py`, `strings.py`, `objects.py`) had the same `start <= entry_stop <= stop` / `start <= entry_stop` boundary conditions.

Because the interpretation expects exactly the baskets the collector supplied, both sides are fixed together (`start < entry_stop`), and the degenerate empty-range case (`entry_start == entry_stop`) is handled explicitly so empty reads still work. `custom.py` uses a `numpy.where`-based selection that does not have this off-by-one, so it was left unchanged.

After the fix, ranges are minimal and results are byte-identical for non-boundary ranges.

## 2. `concatenate(..., allow_missing=True)` shifted offsets after a skipped file

When `hasbranches.arrays` raised `KeyInFileError` (a requested branch missing from one file) and the loop `continue`d, it skipped advancing `global_start = global_stop`, so every subsequent file was read at offsets shifted by the skipped file's entry count.

Reproducer (3 files x 10 entries, middle file missing branch `x`, `entry_start=5`, `entry_stop=25`):
- Before: 15 entries returned (values 5-9, 20-29, including out-of-range 25-29)
- After: 10 entries (values 5-9, 20-24), correct

Fixed by advancing `global_start = global_stop` in the `allow_missing` continue path, in both `TBranch.concatenate` and `RNTuple.concatenate`.

## 3. `RNTuple.HasFields.iterate` did not clamp per-step `entry_stop`

`iterate` passed `entry_stop=start + step_size` per step without clamping to the user's `entry_stop` (only clamped to `num_entries` inside `arrays()`).

Reproducer (50000-entry RNTuple):
```python
nt.iterate(step_size=30000, entry_stop=40000)  # yielded 50000 entries, should be 40000
```

Fixed with `entry_stop=min(start + step_size, entry_stop)`. With `report=True` the ranges now come out as `[(0, 30000), (30000, 40000)]`.

## Testing

- `prek -a --quiet`: clean
- Existing suites pass: `tests/test_0034*`, `tests/test_0017*`, `test_0016_interpretations`, `test_0043_iterate_function`, `test_0044_concatenate_function`, `test_0067_common_entry_offsets`, `test_0194_fix_lost_cuts_in_iterate`, `test_0335_empty_ttree_division_by_zero`, `test_1573_out_of_bounds_slicing`, `test_1630_rntuple_iterate_report`, `test_1406_improved_rntuple_methods`, `test_1250_rntuple_improvements`, plus a wide keyword sweep (`iterate/concatenate/basket/interpret/string/jagged/entry`): all pass.

Regression tests will be added in a follow-up commit on this branch.

Part of #1646.

### DIFF
--- a/src/uproot/behaviors/RNTuple.py
+++ b/src/uproot/behaviors/RNTuple.py
@@ -400,6 +400,7 @@ def concatenate(
                 arrays = library.global_index(arrays, global_start)
             except uproot.exceptions.KeyInFileError:
                 if allow_missing:
+                    global_start = global_stop
                     continue
                 else:
                     raise
@@ -1021,7 +1022,7 @@ class HasFields(Mapping):
                 filter_typename=filter_typename,
                 filter_field=filter_field,
                 entry_start=start,
-                entry_stop=start + step_size,
+                entry_stop=min(start + step_size, entry_stop),
                 library=library,
                 backend=backend,
                 interpreter=interpreter,
--- a/src/uproot/behaviors/TBranch.py
+++ b/src/uproot/behaviors/TBranch.py
@@ -430,6 +430,7 @@ def concatenate(
                 arrays = library.global_index(arrays, global_start)
             except uproot.exceptions.KeyInFileError:
                 if allow_missing:
+                    global_start = global_stop
                     continue
                 else:
                     raise
@@ -2777,7 +2778,9 @@ in file {self._file.file_path}"""
         out = []
         start = entry_offsets[0]
         for basket_num, stop in enumerate(entry_offsets[1:]):
-            if entry_start < stop and start <= entry_stop:
+            if entry_start < stop and (
+                start < entry_stop or entry_start == entry_stop == start
+            ):
                 if 0 <= basket_num < self._num_normal_baskets:
                     byte_start = self.member("fBasketSeek")[basket_num]
                     byte_stop = byte_start + self.basket_compressed_bytes(basket_num)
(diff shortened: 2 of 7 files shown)


henryiii:

Regression tests added in `tests/test_1656_entry_range_arithmetic.py` (commit 142b01f). They cover: no extra basket on exact `entry_stop` boundaries (numerical/jagged/strings interpretations), correct boundary and empty-range reads, `concatenate(allow_missing=True)` keeping global offsets aligned when a file is skipped, and `RNTuple.iterate` clamping the per-step `entry_stop`. Three of the five tests fail on the pre-fix code and pass after.

Note on the RNTuple `concatenate` change: it is applied for symmetry with the TTree fix and matches the reported issue, but in current code an RNTuple `arrays()` call with a missing field returns empty records rather than raising `KeyInFileError`, so the `allow_missing` continue path there is not reachable via a black-box reproducer and is not covered by a dedicated test.

ariostas (approved):

PR #1661, opened 2026-06-10
fix: utility-layer bugs (steps leakage, dask error paths, LRU cache) and Py2-era cleanup

henryiii:

Part of #1646.

## Summary

- **Finding 1** (`_util.py`): Reset `maybe_steps = None` at the top of each dict iteration in `_regularize_files_inner` to prevent steps from one entry leaking into subsequent plain-string entries.
- **Finding 2** (`_util.py`): Fix the Damerau-Levenshtein transposition condition — the second comparison used `b[j-2]` instead of `b[j-1]`, causing all adjacent transpositions to score 2 instead of 1.
- **Finding 3** (`_util.py`): Remove the broken `ratio` parameter from `damerau_levenshtein` (bad math, never used by callers).
- **Finding 4** (`extras.py`): Catch `packaging.version.InvalidVersion` (in addition to `TypeError`) in `older_xrootd`, so self-built XRootD with non-PEP440 version strings correctly return `False`.
- **Finding 5** (`_dask.py`): Move the `count == 0` check above the `step_size` computation in both `_get_dask_array` and `_get_dak_array` to prevent `ZeroDivisionError` masking the intended `ValueError` when `allow_missing=True` and zero TTrees are found.
- **Finding 6** (`_dask.py`): Fix error-message tuple unpacking to use `for f, o, *_ in files` so 3-tuple entries (with explicit steps) don't raise `ValueError` while building the error message.
- **Finding 7** (`cache.py`): Replace the O(n) `_order` list with ordered-dict move-to-end semantics (`self._data[k] = self._data.pop(k)`) for O(1) LRU operations; fix docstring typo in `LRUArrayCache` ("array_cache" → "object_cache").
- **Finding 8** (`_dask.py`, `pyproject.toml`): Replace the `try/except typing_extensions` import dance with a plain `from typing import ...` (all names available since Python 3.8); remove `typing_extensions>=4.1.0` from project dependencies (minimum is already Python 3.10).
- **Finding 9** (`_util.py`): Simplify `tobytes()` helper body to always call `.tobytes()` (NumPy ≥1.9).
- **Finding 10** (`_util.py`, `source/file.py`): Remove Python-2-era `__builtins__` dict-vs-module dance; use `FileNotFoundError` builtin directly; keep `_FileNotFoundError` as a backward-compatible alias.
- **Finding 11** (`_util.py`): Remove unreachable `__fspath__`/`pathlib` branches from `regularize_path` — `isinstance(path, os.PathLike)` already covers any `__fspath__` implementer including `pathlib.Path`.
- **Finding 12** (`_util.py`): Hoist per-call `re.compile()` calls in `file_object_path_split` and `memory_size` to module-level constants.
- **Finding 13** (`behavior.py`): Replace `exec(compile(...))/eval(...)` string execution with `importlib.import_module(...)`.
- **Finding 14** (merge `_get_ttree_form`): **Skipped.** The dask-private `_get_ttree_form` has `HasFields`/RNTuple support and a different `(awkward, ttree, ...)` signature; merging into `_util.get_ttree_form` would require importing `HasFields` there (circular-import risk) and changing the TBranch call site. Benefit is marginal vs. risk.

## Test plan

- [x] All 898 existing tests pass (excluding distributed/network/ROOT-dependent)
- [x] 9 new regression tests added in `tests/test_1660_fix_util_dask_cache.py`:
  - `test_regularize_files_steps_no_leakage` — dict steps not inherited by next plain-string entry
  - `test_damerau_levenshtein_transposition` — adjacent transpositions score 1
  - `test_damerau_levenshtein_no_ratio_param` — confirm `ratio` param removed
  - `test_lru_cache_eviction_order` — LRU evicts correct entry
  - `test_lru_cache_setitem_update_moves_to_end` — update moves to MRU
  - `test_lru_cache_keys_order` — keys() is ordered LRU→MRU
  - `test_lru_cache_thread_safety` — concurrent get/set doesn't error
  - `test_lru_cache_no_limit` — None limit never evicts
  - `test_dask_zero_files_error_message_with_explicit_steps` — 3-tuple entries don't crash error-message generation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,8 +86,7 @@ dependencies = [
   "xxhash",
   "numpy",
   "fsspec!=2026.2.0",
-  "packaging",
-  "typing_extensions>=4.1.0; python_version < '3.11'"
+  "packaging"
 ]
 description = "ROOT I/O in pure Python and NumPy."
 dynamic = [
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -6,18 +6,9 @@ import socket
 import time
 from collections.abc import Callable, Iterable, Mapping
 from concurrent.futures import Executor
+from typing import TYPE_CHECKING, Any, Final, NamedTuple, Protocol, TypeVar
 
 import awkward
-
-from uproot.source.chunk import SourcePerformanceCounters
-
-try:
-    from typing import TYPE_CHECKING, Final, NamedTuple
-
-    from typing_extensions import Any, Protocol, TypeVar
-except ImportError:
-    from typing import TYPE_CHECKING, Any, Final, Protocol, TypeVar
-
 import numpy
 
 import uproot
@@ -27,6 +18,7 @@ from uproot.behaviors.RNTuple import (
     _regularize_step_size as _RNTuple_regularize_step_size,
 )
 from uproot.behaviors.TBranch import HasBranches, TBranch, _regularize_step_size
+from uproot.source.chunk import SourcePerformanceCounters
 
 if TYPE_CHECKING:
     from awkward._nplikes.typetracer import TypeTracerReport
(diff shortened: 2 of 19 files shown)


ariostas (approved):
I did a little bit of extra cleanup.

PR #1664, opened 2026-06-23
fix: Add missing class mapping for ROOT::TString

Paramveersingh-S:
## Summary

Fixes issue #8 by adding the missing model mapping for `ROOT.TString` to support nested reading within a `TTree`.

## Problem

The model mappings at the bottom of `uproot/models/TString.py` were missing entries for scoped C++ `TString` aliases (e.g. `ROOT::TString`). This led to `uproot` failing to map and read these nested objects properly.

## Solution

Registered the missing C++ scope mappings at the end of `uproot/models/TString.py`:
- `uproot.classes["ROOT::TString"] = Model_TString`
- `uproot.classes["ROOT::ROOT::TString"] = Model_TString`

### DIFF
--- a/src/uproot/models/TString.py
+++ b/src/uproot/models/TString.py
@@ -105,3 +105,5 @@ in file {self.file.file_path}"""
 
 
 uproot.classes["TString"] = Model_TString
+uproot.classes["ROOT::TString"] = Model_TString
+uproot.classes["ROOT::ROOT::TString"] = Model_TString


ariostas:
Thank you for your interest in helping out, @Paramveersingh-S.

However, there is no issue #8, and `TString` is not aliased under the `ROOT` namespace. If you have an example where this is an issue, let me know and we can reopen this.

Issue #1627, opened 2026-05-07
`from_pyroot` fails with ROOT 6.32 (LCG_109) — `TPython::Eval` removed


wiso:
### Description

`uproot.from_pyroot()` raises an `AttributeError` when used with ROOT 6.32, which ships with LCG_109. The root cause is that `pyroot.py` internally calls `TPython::Eval` via `gInterpreter.Declare`, but `TPython::Eval` has been removed in ROOT 6.32.

### Steps to reproduce

```python
import ROOT
import uproot

rdf = ROOT.RDataFrame("CollectionTree", "myfile.root")
rdf = rdf.Define("m_lly_GeV", "HGamEventInfoAuxDyn.m_lly / 1E3")
h_root = rdf.Histo1D(("m_lly", "", 100, 0, 200), "m_lly_GeV").GetValue()

h = uproot.from_pyroot(h_root)  # fails here
```

### Error

```
AttributeError: Failed to get attribute _Uproot_buffer_sizer from ROOT
input_line_N:11:25: error: no member named 'Eval' in 'TPython'
    (void*)TPython::Eval("__import__('uproot').pyroot.pyroot_to_buffer.sizer")
```

The failure originates in `uproot/pyroot.py` at the `gInterpreter.Declare(...)` call that defines `_Uproot_buffer_sizer`, which internally uses `TPython::Eval`

### Environment

- **uproot version**: 5.6.0
- **ROOT version**: 6.32 (via LCG_109, `x86_64-el9-gcc14-opt`)
- **Python version**: 3.13
- **Platform**: EL9 / CVMFS (`/cvmfs/sft.cern.ch/lcg/views/LCG_109/`)

### Expected behaviour

`uproot.from_pyroot(th1)` converts a PyROOT TH1 object to an uproot object, which can then be converted to a `hist.Hist` via `.to_hist()`.

ariostas:
Hi @wiso, this was already addressed a while ago. Could you try with Uproot v5.7.2 or newer to see if it works?

wiso:
Actually it is not working. I am using 5.7.4 now, and ROOT 6.32.02

```
import ROOT
import uproot
h = ROOT.TH1F("H", "H", 100, 0, 100)
uproot.from_pyroot(h)
```

```
input_line_38:11:92: error: too many arguments to function call, expected single argument 'cmd', have 2 arguments
[...]

PR #1569, opened 2026-02-05
fix: resolve issues with pyroot buffer reallocation

ariostas:
This PR fixes the code used for buffer relocation that I changed in #1463. Somehow that code passed all the tests we had, but it wasn't the right way to use `TPython::Exec`. I looked more into how it's supposed to be used, and now it works well. I extended the tests to make sure we are testing more thoroughly.

Closes #1559

PR #1667, opened 2026-06-25
fix: avoid std::any in pyroot_to_buffer to support older ROOT versions

Paramveersingh-S:
### Description
This PR resolves issue #1627, where `uproot.from_pyroot()` fails on ROOT 6.32+ due to the removal of `TPython::Eval`. 

While a previous fix on `main` attempted to use `TPython::Exec` combined with `std::any`, that approach relies on C++17 and passing pointers back through PyROOT bindings, which can be fragile across different compiler setups and ROOT versions. 

This PR refactors `pyroot_to_buffer` to completely bypass the need to return complex values from `TPython`. Instead, it uses a far more robust and universally compatible approach: **C++ Global Variables**.

### Technical Details
- Declared two simple C++ global variables (`_uproot_sizer_ptr` and `_uproot_out_buffer_ptr`) via `ROOT.gInterpreter.Declare`.
- In the C++ `_uproot_TMessage_reallocate` callback, we now simply read from and write to these global pointers.
- In the Python `reallocate()` function, we directly update the global variable: `ROOT._uproot_out_buffer_ptr = pyroot_to_buffer.buffer.ctypes.data`.
- This entirely eliminates all usage of `std::any` and complex `TPython::Exec` return mechanics, resulting in extremely clean standard C++.
- Thread-safety is inherently maintained because `pyroot_to_buffer` is protected by `pyroot_to_buffer.lock`.

This guarantees compatibility with ROOT 6.32+ while maintaining full backwards compatibility with older ROOT versions.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -15,12 +15,20 @@ permissions:
 
 jobs:
   build:
-    name: Test on ${{ matrix.platform }} with Python ${{ matrix.python-version }}
+    name: Test on ${{ matrix.platform }} with Python ${{ matrix.python-version }}${{ matrix.root-version != '' && format(' and ROOT {0}', matrix.root-version) || '' }}
     strategy:
       fail-fast: false
       matrix:
         platform: [windows-latest, macos-latest, ubuntu-latest]
         python-version: ['3.10', '3.14', 3.14 python-freethreading]
+        root-version: ['']
+        include:
+          - platform: ubuntu-latest
+            python-version: '3.10'
+            root-version: latest
+          - platform: ubuntu-latest
+            python-version: '3.10'
+            root-version: '6.32'
 
     runs-on: ${{ matrix.platform }}
     timeout-minutes: 30
@@ -44,6 +52,7 @@ jobs:
             --strict-channel-priority
             conda-forge/label/python_rc::_python_rc
             python=${{ matrix.python-version }}
+            ${{ matrix.root-version == 'latest' && 'root' || matrix.root-version == '6.32' && 'root=6.32.*' || '' }}
           micromamba-version: 2.6.0-0   # unpin when there is a release newer than 2.6.1-0
 
       - name: Set PYTHON_GIL=0 for free-threaded builds
@@ -56,13 +65,6 @@ jobs:
         env:
           EXPECTED_PYTHON_VERSION: ${{ matrix.python-version }}
         run: python -c "import os, sys; actual = '.'.join(str(s) for s in sys.version_info[:2]); expected = os.environ['EXPECTED_PYTHON_VERSION'].split()[0]; assert actual == expected, f'{actual} incorrect!'"
-
-      - name: Install ROOT
-        if: matrix.python-version == 3.10  &&  runner.os == 'Linux'
-        run: |
-          micromamba env list
-          micromamba install root
-          micromamba list
 
       - name: Install sshd for fsspec ssh tests
         if: runner.os == 'Linux'
--- a/src/uproot/pyroot.py
+++ b/src/uproot/pyroot.py
@@ -97,19 +97,17 @@ public:
   size_t oldsize;
 };
 
+_Uproot_buffer_sizer* _uproot_sizer_ptr = nullptr;
+size_t _uproot_out_buffer_ptr = 0;
+
 char* _uproot_TMessage_reallocate(char* buffer, size_t newsize, size_t oldsize) {
-    std::any sizer_any;
-    TPython::Exec("_anyresult = __import__('uproot').pyroot.pyroot_to_buffer.sizer_asany", &sizer_any);
-    _Uproot_buffer_sizer& sizer = std::any_cast<_Uproot_buffer_sizer&>(sizer_any);
-    sizer.buffer = reinterpret_cast<size_t>(buffer);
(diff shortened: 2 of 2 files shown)


ariostas:
Hi @Paramveersingh-S, do you have an example where this new change works, but the old one doesn't? As far as I can tell, the issue in #1627 should have been fixed in v5.7.2 by #1569.

It seems to me like your changes are equivalent to what is currently in main. I do like that it's a bit shorter, but the `TPython::Exec`/`std::any` approach is what is recommended in the ROOT documentation (see https://root.cern/doc/master/classTPython.html and https://root.cern/doc/v634/release-notes.html)

Paramveersingh-S:
Hi @ariostas, thanks for taking the time to review this! 
You are completely correct that `#1569` successfully resolves the immediate issue for ROOT 6.32+ by migrating away from the removed `TPython::Eval` method. For users running the latest ROOT versions on modern environments, the current `main` branch works fine.
However, the `TPython::Exec` + `std::any` approach inadvertently introduced a severe backwards-compatibility break for older ROOT versions. **The old change fails entirely on ROOT `<6.30` (e.g., 6.26 or 6.28), which is still heavily used in many enterprise and computing-cluster environments.**
Here is the deep-dive into the two fatal flaws of the `std::any` approach, and the exact scenario where it fails: #1510 The Reproducer: Missing Overload in Older ROOT (API Break)
The primary issue lies in the function signature of `TPython::Exec`. The two-argument overload (`Bool_t TPython::Exec(const char* cmd, std::any* ret)`) was introduced very recently in ROOT to compensate for the removal of `TPython::Eval`. 
If a user running **ROOT 6.26** executes `uproot.from_pyroot()`, the Cling JIT compiler attempts to compile the `_uproot_TMessage_reallocate` callback from `#1569` and will instantly fail with:

error: no matching function for call to 'TPython::Exec'
Because in older ROOT versions, TPython::Exec strictly takes a single argument (const char* cmd). My PR avoids this completely by abandoning TPython for the return value and instead utilizing PyROOT's native ability to directly bind and manipulate C++ global pointers. This guarantees API compatibility across all ROOT versions, old and new.

2. The C++14 Standard Dependency (ABI Break)
std::any is exclusively a C++17 feature. While ROOT 6.30+ enforces C++17, legacy ROOT versions default to compiling Cling in C++14 mode. If a user is locked to an older environment where Cling operates in C++14, the JIT-compiled C++ block in main will throw a compilation error the moment it tries to resolve <any>. By reverting to standard C++ pointers (which are entirely standard-agnostic), this PR restores compatibility for C++11/C++14 environments.

3. Execution Overhead & Complexity
From a purely architectural standpoint, using TPython::Exec specifically to pass a pointer back and forth is computationally heavier. It requires C++ to ask the Python interpreter to parse and evaluate a string (_anyresult = __import__...). Using PyROOT to expose a C++ pointer to Python—where Python simply sets ROOT._uproot_out_buffer_ptr = new_ptr—is direct memory access. It bypasses the interpreter's string-evaluation layer, resulting in lower overhead during buffer reallocation.

Thread Safety (A Note on Globals)
I completely agree that global variables are typically an anti-pattern due to race conditions. However, in this highly specific context, the pyroot_to_buffer reallocation flow is strictly serialized by the existing pyroot_to_buffer.lock. The global pointers are 100% thread-safe under this lock.

Summary: While #1569 fixed ROOT 6.32+, it silently broke from_pyroot for anyone on older ROOT stacks by injecting C++17 requirements and calling an API overload that didn't exist yet. This PR achieves the same fix for 6.32+ but uses vanilla, standard-agnostic C++ that guarantees flawless backwards compatibility.

ariostas:
Thank you, @Paramveersingh-S. Yeah, that's a good point.

I'm testing it out, and it seems to work fine with new versions of Uproot (and the CI also passes), but I'm getting a segfault with ROOT 6.32. I'll look into it further.

ariostas:
@Paramveersingh-S I made some changes (with assistance from Claude) so that it works on ROOT 6.32. I also changed the CI job so that it tests both the latest version of ROOT as will as 6.32 to make sure older releases work fine. Let me know what you think.

Paramveersingh-S:
@ariostas This final iteration is fantastic.

By completely abandoning the attempt to hijack the TMessage reallocation hook, we bypass all the fragility of PyROOT interpreter callbacks, GIL state management, and TPython API changes. Letting TMessage manage its own memory and simply copying the resulting payload via a standard memcpy is definitively the most robust way to handle this across all ROOT versions. It cleanly resolves both the ROOT 6.32 segfaults and the C++14 legacy compatibility issues I highlighted.

One minor suggestion for future maintainability: The message.Buffer() + 8 logic in _uproot_copy_tmessage is highly precise—it perfectly replicates our old behavior where SetBuffer discarded the initial 8-byte TMessage header, ensuring Uproot only receives the raw object payload. It might be worth adding a tiny inline C++ comment (// Skip 8-byte TMessage header to match legacy payload-only extraction) just so future maintainers understand exactly where that magic offset comes from!

Also, bypassing PyROOT's buffer bindings by passing buffer.ctypes.data directly to _uproot_memfile_copyto is a great, bulletproof workaround for any strict type-casting changes in newer PyROOT versions. The cleanup utilizing context managers (with uproot.open(...) as file:) is a great touch too.

ariostas (approved):

PR #1668, opened 2026-06-25
fix: add typenames to form so Coffea can use them

ariostas:
#1642 removed grouped branches from appearing in the form in dask mode (which makes sense and matches the eager mode). However, Coffea relied on those branches to extract typenames.

This PR adds the typenames as a parameter of the form, so that Coffea can use them. There will be an accompanying PR in Coffea.


mrzimu (commented):
It's good to me 👍

ianna (approved):
@ariostas - looks good to me! Thanks.

PR #1670, opened 2026-06-30
fix: accidental dropping of AsGrouped

ariostas:
Seems like the filtering introduced in #1642 was a little too strong. This popped up as there was a test in Coffea that still failed, and the fix turned out to be more complicated than I thought.

In summary, what was happening is that if you did something like `tree.keys(filter_name="grouped_branch")` it would return an empty list (and similarly for `arrays`). So one needs to do some extra checks to see when they can be dropped.

@mrzimu it would be great if you can have a look at this and let me know what you think.

AI disclosure: I used assistance from Claude Code for this PR.

### DIFF
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -935,7 +935,10 @@ class TrivialFormMappingInfo(ImplementsFormMappingInfo):
         interpretation_executor,
         options: Any,
     ) -> Mapping[str, AwkArray]:
-        # First, let's read the arrays as a tuple (to associate with each key)
+        # Read the arrays as a top-level awkward RecordArray. Omitting how= (the
+        # default) ensures that AsGrouped branches are returned as proper awkward
+        # RecordArrays rather than Python tuples of sub-arrays (which how=tuple
+        # would produce), allowing awkward.to_buffers() to work correctly below.
         arrays = tree.arrays(
             keys,
             entry_start=start,
@@ -943,7 +946,6 @@ class TrivialFormMappingInfo(ImplementsFormMappingInfo):
             ak_add_doc=options["ak_add_doc"],
             decompression_executor=decompression_executor,
             interpretation_executor=interpretation_executor,
-            how=tuple,
         )
 
         if isinstance(tree, HasFields):
@@ -959,9 +961,9 @@ class TrivialFormMappingInfo(ImplementsFormMappingInfo):
         # subform, as they're derived from `branch.interpretation.awkward_form`
         # Therefore, we can correlate the subform keys using `expected_from_buffers`
         container = {}
-        for key, array in zip(keys, arrays, strict=True):
+        for key in keys:
             # First, convert the sub-array into buffers
-            ttree_subform, _length, ttree_container = awkward.to_buffers(array)
+            ttree_subform, _length, ttree_container = awkward.to_buffers(arrays[key])
 
             # Load the associated projection subform
             projection_subform = self._form.content(key)
--- a/src/uproot/_dask.py
+++ b/src/uproot/_dask.py
@@ -1649,15 +1649,23 @@ def _get_dak_array(
             )
         )
 
-    # Filter out AsGrouped branches: they are grouping containers without their own
-    # data buffers. This matches tree.arrays(expressions=None) which also skips them.
+    # Filter out AsGrouped branches whose children are already present separately in
+    # common_keys. Such branches are pure grouping containers with no data buffers of
+    # their own, and keeping them would produce a redundant layer in the form.
+    # AsGrouped branches whose children are NOT in common_keys are kept because they
+    # carry structural information (e.g. ElementLink records) that would otherwise be lost.
     if ttrees and not isinstance(ttrees[0], HasFields):
+        common_keys_set = set(common_keys)
         common_keys = [
             k
             for k in common_keys
             if not isinstance(
                 ttrees[0][k].interpretation,
                 uproot.interpretation.grouped.AsGrouped,
+            )
+            or not any(
+                child in common_keys_set
+                for child in ttrees[0][k].keys(recursive=True, full_paths=True)
             )
         ]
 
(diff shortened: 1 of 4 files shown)


mrzimu:
> Seems like the filtering introduced in #1642 was a little too strong. This popped up as there was a test in Coffea that still failed, and the fix turned out to be more complicated than I thought.
> 
> In summary, what was happening is that if you did something like `tree.keys(filter_name="grouped_branch")` it would return an empty list (and similarly for `arrays`). So one needs to do some extra checks to see when they can be dropped.
> 
> @mrzimu it would be great if you can have a look at this and let me know what you think.
> 
> AI disclosure: I used assistance from Claude Code for this PR.

Ahhh, I'll have a look at this ASAP

ariostas:
> Ahhh, I'll have a look at this ASAP

No rush! I'll revisit this tomorrow to clean it up a bit.

mrzimu (commented):
Yes, the original filter behavior is too aggressive. I feel the main problem is that the behavior of `filter_name` is still a bit ambiguous, so may I summarize it. 

Suppose that we have branches like:
```
Jet (AsGrouped)
- Jet.type
- Jet.energy
```

When "filter_name":
1. Only matches AsGrouped branch: return its subbranches with a top-level layout, e.g. `filter_name="Jet" -> {Jet: [Jet.type, Jet.energy]}`
2. Only matches non-AsGrouped branches: group them without a top-level layout, e.g. `filter_name="Jet.*" -> [Jet.type, Jet.energy]`
3. Matches both AsGrouped branch and its subbranches:
    - all subbranches are matched: **without** top-level layout, e.g. `filter_name="Jet*" -> [Jet.type, Jet.energy]`
    - some of the subbranches are matched: seems not possible to happen?

Maybe the test can be written following this logic?

ariostas:
> some of the subbranches are matched: seems not possible to happen?

I think it can happen with some regex, or it could be something like this:
```python
filter_branch=lambda b: b.name in ("Jet", "Jet.type")
```
In that case we would have to decide what makes sense for it to return. It would make more sense to me to return the grouped `Jet`, but currently, and even before #1642, it returns just `Jet.type`. So I don't know if this would technically be a breaking change. What do you think, @mrzimu? When I was implementing the equivalent functionality for RNTuples, I decided that if a parent matches, then the children are all included.

I think we agree that `tree.arrays(filter_name="Jet")` returning an empty array should be considered a "bug". Then by the same logic maybe it makes sense to also accept that a filter that matches the asgrouped and a partial subset should end up returning the whole asgrouped?

mrzimu:
Yeah I totally agree that `tree.arrays(filter_name="Jet")` returning an empty array should be considered a "bug".

As for the case `filter_branch=lambda b: b.name in ("Jet", "Jet.type")`, it is pretty hard to deduce what user aims to do (does the user want both (perhaps mistakenly), or do they specifically want the `type` sub-branch of `Jet`?), so I think returning the full grouped structure is more predictable and user-friendly. Also, it's consistent with what you've implemented for RNTuple.

But if this would be a breaking change, keeping the existing behavior with better documentation could also be reasonable.

ariostas:
Thank you, @mrzimu. I'll bring this up tomorrow during the meeting to see what other people think.

ikrommyd:
> Yeah I totally agree that `tree.arrays(filter_name="Jet")` returning an empty array should be considered a "bug".

What is the scenario here? That "Jet" does not exist?  My intuition says that if Jet does not exist, an error should be raised but I don't know what effects this would have in coffea.
As for statements like `filter_branch=lambda b: b.name in ("Jet", "Jet.type")` I think the user is not requesting both here. They are requesting "any" of them. So if no match is found, you error, but if at least one is found, you return that.

That's my intuition of how I would have implemented this myself. Now there may be scenarios here where this is bad inside coffea and I am not aware of.

mrzimu:
> > Yeah I totally agree that `tree.arrays(filter_name="Jet")` returning an empty array should be considered a "bug".
> 
> What is the scenario here? That "Jet" does not exist? My intuition says that if Jet does not exist, an error should be raised but I don't know what effects this would have in coffea. As for statements like `filter_branch=lambda b: b.name in ("Jet", "Jet.type")` I think the user is not requesting both here. They are requesting "any" of them. So if no match is found, you error, but if at least one is found, you return that.
> 
> That's my intuition of how I would have implemented this myself. Now there may be scenarios here where this is bad inside coffea and I am not aware of.

Following this logic, when both the "Jet" branch and its sub-branch "Jet.type" are matched by the filter, returning the whole "Jet" group seems more reasonable to me(at present it just returns "Jet.type").

ariostas:
> What is the scenario here?

The scenario is that `Jet` is a `AsGrouped` branch that doesn't contain any data by itself, but instead it just groups other branches (`Jet.type`, `Jet.energy`, ... in the example above).

So in the example `filter_branch=lambda b: b.name in ("Jet", "Jet.type")`, `Jet.type` definitely should be returned, but what about other branches that are grouped by `Jet`? I think that the output should be the same as with `filter_branch=lambda b: b.name  == "Jet"`.

Although another question is what should happen with `filter_branch=lambda b: b.name.startswith("Jet")`. It might seem contradictory, but I think in this case it makes more sense for it to behave like `filter_name="Jet.*"` (i.e. ignoring the grouped branch).

So my proposal for the different cases would be: (by non-trivial subset I mean non-empty and not containing everything).

- Only AsGrouped matched -> return all subbranches grouped (as Record array).
- Only non-trivial subset of sub-branches matches -> return those ungrouped.
- AsGrouped and all subbranches matched -> return all subbranches ungrouped.
- AsGrouped and non-trivial subset of sub-branches matched -> return all subbranches grouped (as Record array).

ikrommyd:
Yeah if Jet matches an AsGrouped branch I would return them all.

ariostas:
I went through some iterations with Claude and I think this implementation makes sense. There was actually still one slight difference between the eager and dask case, but they fully match now. It also added tests to make sure everything matches up.

@mrzimu let me know what you think

ianna (approved):
@ariostas — looks good to me — as discussed yesterday. Please go ahead and merge it if you are finished with it. Thanks

ariostas:
I'll merge this so that we can make a release. If it turns out we missed some subtlety well revisit again

PR #1675, opened 2026-07-03
fix: concurrency settings from build workflow

ianna:
Rename concurrency settings from the build workflow.

### DIFF
--- a/.github/workflows/build-distributions.yml
+++ b/.github/workflows/build-distributions.yml
@@ -9,6 +9,9 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-build-distributions
+  cancel-in-progress: true
 jobs:
   dist:
     name: Build distribution artifacts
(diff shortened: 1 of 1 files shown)


ianna (commented):
Deploy to PyPI was canceled since deadlock was detected for concurrency group between the top level workflow and this one.

.github/workflows/build-distributions.yml
```diff
@@ -9,10 +9,6 @@ on:
 permissions:
   contents: read
```
ariostas, line 11:
```suggestion

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}-build-distributions
  cancel-in-progress: true
```

Thanks for pointing this out. Sorry I missed it earlier.

The issue was that `${{ github.workflow }}` resolves to the caller workflow, so it ends up being the same as the one below.
https://github.com/scikit-hep/uproot5/blob/eac7850c6dfcc643ff731668890c6f42575f9685/.github/workflows/deploy.yml#L10

But I think it would be better to keep the concurrency settings. With this change, they won't conflict.

ianna: Thanks @ariostas ! Yes that’s another option. I agree.

ariostas (approved):

PR #1681, opened 2026-07-11
Support for `to_boost` conversion for TProfile2D (analogously to the 1D) [WIP]

pietroGru:
**Description**
The PR implements the `to_boost()` method for TProfile2D.

_Status_: at present, the code compiles and I am able to retrieve to call the `to_hist()` method for some TProfile2D object from uproot. As shown from the following picture the view seems correctly filled.

| root | uproot |
| --- | --- |
| Screenshot 2026-07-11 182819 | Screenshot 2026-07-11 182910 |

The PR is in draft as it is a work in progress: it is open to feedbacks and there are open points to clarify.
1. the test `test_0405_write_a_histogram.py` fails. Understand whether is PR-related or not.
2. are the quantities attached to the TProfile correctly matching in root/uproot?
3.check what happens in cases where the hist has a mean storage.

### DIFF
--- a/src/uproot/behaviors/TProfile2D.py
+++ b/src/uproot/behaviors/TProfile2D.py
@@ -103,4 +103,95 @@ class TProfile2D(uproot.behaviors.TProfile.Profile):
             return values[1:-1, 1:-1], errors[1:-1, 1:-1]
 
     def to_boost(self, metadata=boost_metadata, axis_metadata=boost_axis_metadata):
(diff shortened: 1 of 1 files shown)


ariostas:
@pietroGru could you add a test for this? You can use this file from skhep-testdata: `uproot-issue-227a.root`. Also, since you marked this as ready for review, could you also remove the "[WIP]" from the title?

PR #1687, opened 2026-07-21
feat: Add RNTuple row extension and field addition via update mode

Yokubas:
## Summary
Implements in-place modification of existing RNTuples:

- `f["name"].extend({"x": array1, "y": array2})` — append new rows to existing RNTuple
- `f["name"].add_fields({"z": np.int32, "w": np.float32, ...})` — add one or more new fields back-filled with zeros
- `f["name"].add_fields({"particle.phi": np.float32, "particle.eta": np.float64, ...})` — add subfields to existing untyped structs
- `f["name"].extend({"x": array1, "z": array2}, accept_new_fields=True)` — auto-add new fields and extend

## How it works
- Reads existing RNTuple metadata (anchor, header, footer, page lists)
- Reconstructs a writable NTuple object from the existing file
- For row extension: writes new pages and adds new cluster group to footer
- For field addition: uses RNTuple's deferred column mechanism — new fields are added to the footer's extension records with `first_element_index = num_entries`, marking where new data starts. Old cluster groups are left completely untouched. The reader automatically zero-pads entries before `first_element_index`. Subsequent `extend` calls write new cluster groups that include the new column.
- Only footer and new data are written — existing data never touched
- Anchor updated in-place since it's always the same size
- `add_fields`/`extend` on an RNTuple in a subdirectory now resolve their key against that subdirectory, not always the file root
- `add_fields` on an RNTuple created earlier in the same session (via `mkrntuple` or directory assignment, not reopened via `uproot.update`) now lazily loads the state it needs instead of assuming it was already loaded

## Tests
42 tests in `tests/test_1687_rntuple_update.py` covering:
- Basic extend and add_fields
- Multiple extends in separate sessions
- Multiple fields added in a single call
- Sequential add_fields calls holding the same object
- add_fields + extend in same session with same object
- Variable length arrays
- Mixed types (scalar + jagged)
- Empty ntuples
- Multiple ntuples in same file
- Subfields and deeply nested subfields with correct parent resolution
- Validation: nonexistent parent, typed parent, collection parent, wrong field types
- accept_new_fields kwarg behavior
- ROOT verification reading actual column values via RNTupleReader
- add_fields/extend on an RNTuple in a subdirectory (regression test)
- add_fields on a same-session (not reopened) RNTuple (regression test)
- Repeated add_fields calls in one session don't duplicate field records (regression test)
- extend() after add_fields() doesn't scramble which data lands in which column, verified against real ROOT's own RNTupleReader as well as uproot's reader (regression test for the corruption below)
- add_fields dotted-path parent resolution rejects an ambiguous bare name instead of silently picking the wrong parent (regression test)
- A field's real (non-root) nesting depth is no longer misidentified as root during multi-level parent resolution (regression test)
- add_fields's duplicate-field-name check no longer false-positives against an unrelated nested field's name (regression test)
- add_fields checks column-encoding compatibility before mutating shared state, and that check's error message is no longer masked by an unrelated dotted-path error (regression test)
- `WritableNTuple.extend` has a docstring again (regression test — see below)
- add_fields on a genuine ROOT-written fixture (`ntpl001_staff`) raises cleanly (regression test)

## Blocking issues addressed (from review)
- Silent data corruption on ROOT-written files → column record comparison raises clear `ValueError` with encoding mismatch instead of writing garbage data
- Wrong element_offset for jagged fields → per-column element counts recovered from existing page lists instead of assuming all columns have `num_entries` elements
- Multi-cluster groups in add_fields → now fully supported — writes one page per new field per cluster sized to that cluster's entry span
- Stale in-memory state after add_fields → field records and column counts updated directly in memory; footer/page lists/akform reloaded from file
- Subfield parent resolved by bare name → walks full dotted path using parent-id chain, correctly handles both uproot and ROOT parent-id conventions
- Extension field records were double-counted on the second and every subsequent `add_fields` call in one session (footer's cumulative extension list was re-added on top of an already-updated field list each time) — this could corrupt the file badly enough that a third `add_fields` call raised `IndexError` on reopen. Fixed by recombining the header's fixed field list with the footer's cumulative list, instead of the previous call's already-merged result.
- `extend()` after `add_fields()` could silently write each field's data under the wrong column: `extend()`'s dict-to-awkward conversion always sorts fields alphabetically, but a header reloaded from disk reflects the true on-disk (insertion) field order — awkward's Form equality doesn't check field order, so the mismatch went undetected and columns were written under each other's keys. Confirmed independently against ROOT's own `RNTupleReader`, not just uproot's reader. Fixed by reordering the data to match the header's actual field order before writing.
- The dotted-path parent-matching used for single-level `"parent.field"` paths matched by bare name only, with no check that the match was unique — if two fields (e.g. a top-level record and an unrelated nested record) shared the same bare name, `add_fields` could silently attach the new field under the wrong one. Now raises a clear "ambiguous" error and suggests a fully-qualified path.
- Fixed a related bug found while fixing the above: multi-level path resolution's "is this a root field" check (`parent_field_id == 0 or parent_field_id == i`) wasn't a valid signal on its own — it also matched any non-root field whose real parent happened to sit at index 0 — which could make a legitimate 3+-level dotted path fail to resolve. A field is root iff it's its own parent; fixed to check only that.
- `add_fields`'s duplicate-field-name check compared a new (always top-level) field's name against every existing field's bare name, including nested ones — a legitimate new top-level field could be rejected as "already exists" if any unrelated nested field happened to share its name. Now only compares against actual top-level fields.
- `add_fields` checked column-encoding compatibility only partway through the per-field loop, after already mutating the shared, persistent footer object and after dotted-path resolution could raise a more specific but less fundamental error first — unlike `extend()`, which checks it immediately. Moved to the top of `add_fields`, matching `extend()`.
- Found and fixed a related but currently-unreachable bug in the same area: the column-count recovery used when reopening a file only read the first cluster of each cluster group, undercounting for any group with more than one cluster (not possible via uproot's own writer today, but would silently corrupt element offsets the moment that changes).
- `WritableNTuple.extend`'s docstring had been accidentally placed after a statement instead of as the function's first statement, silently turning it into a dangling string and dropping it from generated docs — restored, and both `extend`/`add_fields` now explicitly document that only uproot-written RNTuples are supported for these operations after reopening a file.

## Known limitations
- ROOT-written RNTuples cannot be extended (split encoding not yet supported) — this restriction is now stated explicitly in the `extend`/`add_fields` docstrings, and is covered by a test against a real ROOT-written fixture rather than only a synthetic one
- Only scalar numeric types supported in add_fields
- Multi-cluster RNTuples not yet supported for add_fields (a latent bug in the read-back accounting for this case was also fixed defensively — see above — even though it isn't reachable via the current public API)
- File-like objects not supported (requires file path for re-reading metadata)


src/uproot/writing/writable.py
```diff
@@ -1054,6 +1049,142 @@ def get_chunk(start, stop):
...
+        (anchor, header, footer, page lists) and sets up the in-memory
+        state needed for subsequent ``extend`` or ``add_fields`` calls.
+
+        Args:
+            key: The ROOT key object pointing to the RNTuple in the file.
+
+        Returns:
+            :doc:`uproot.writing.writable.WritableNTuple`
+        """
+        import numpy
```
ariostas: This import should probably go at the top. The two below are fine (to avoid circular imports).

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1049,142 @@ def get_chunk(start, stop):
...
+
+        import uproot.writing._cascade as casc
+        import uproot.writing._cascadentuple as cnt
+
+        name = key.name.string
+        existing_file = uproot.open(self.file_path, minimal_ttree_metadata=False)
```
ariostas: It seems a little weird to me to use `uproot.open` when you already have a file open. I'm guessing this is just to open it in read mode, but I feel like there should be some other way. I'll get back to you on this.

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1049,142 @@ def get_chunk(start, stop):
...
+            )
+            footer.extension_field_record_frames.append(new_field)
+        for cr in existing_footer.extension_links.column_records:
+            new_col = cnt.NTuple_Column_Description(
+                cr.type, cr.nbits, cr.field_id, cr.flags, cr.repr_idx
+            )
+            footer.extension_column_record_frames.append(new_col)
+        for cg in existing_footer.cluster_group_records:
+            loc = cg.page_list_link.locator
+            start = loc.offset - 56
```
ariostas: What's the 56?

src/uproot/writing/writable.py
```diff
@@ -2173,17 +2308,200 @@ def extend(self, data):
...
+        For example,
+
+        .. code-block:: python
+
+            with uproot.update("file.root") as f:
+                f["mytuple"].add_fields({"z": np.int32})
+        """
+        import numpy
+
+        import uproot.compression
```
ariostas: These two imports should be at the top of the file

src/uproot/writing/writable.py
```diff
@@ -2173,17 +2308,200 @@ def extend(self, data):
...
+                )
+                key = self._file.root_directory._cascading.data.get_key(
+                    self._path[-1], 1
+                )
+                reloaded = self._file.root_directory._load_existing_ntuple(key)
+                self._cascading = reloaded._cascading
+
         self._cascading.extend(self._file, self._file.sink, data)
 
+    def add_fields(self, new_fields):
```
ariostas, line 3146: We should normalize the input with `uproot.writing._type_specification_to_awkward_form` so that it can take things that are more generic than a dict. (And so it matches what `mkrntuple` accepts)

tests/test_1687_rntuple_update.py
ariostas, line 1: The convention for test filenames is `test_XXXX_[short description].py`, where XXXX is a PR or issue number. So now that you opened this PR, you can rename it to `test_1687_rntuple_update.py`.

tests/test_rntuple_update.py
```diff
@@ -0,0 +1,402 @@
+import uproot
+import os
+import pytest
+
+ROOT = pytest.importorskip("ROOT")
```
ariostas: It would be better to only skip the piece that needs ROOT instead of the entire file.

tests/test_1687_rntuple_update.py
ariostas, line 1: The tests look great! The only things that I see missing for now are:
- Try adding a subfield "parent.x", where parent doesn't exist, and make sure it fails.
- Try adding a subfield "parent.x", where parent has some non-empty typename. This should fail with a message saying that only untyped records can be extended.
- Try adding a subfield to an an untyped collection (an array of untyped records). This should fail, if I remember correctly, but we need to double-check.

ariostas (commented):
Thank you @Yokubas, this is fantastic progress! I left a few comments.

src/uproot/writing/writable.py
Yokubas: The 56 was a magic number that turned out to be wrong — the actual RBlob key header size is `_key_format_big.size + 8 = 42`. It worked by accident because of the freesegments protection that is applied afterwards. I've replaced it with `_rblob_key_size = uproot.reading._key_format_big.size + 8` computed at the top of the method.

src/uproot/writing/writable.py
Yokubas: Yes, I open it in read mode to get the existing metadata. Happy to change it if you find a better way.

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1049,143 @@ def get_chunk(start, stop):
...
+        existing = existing_file[name]
+        _ = existing.keys()
+        full_akform, _ = existing.to_akform()
+        am = existing._ntuple.all_members
+        existing_key = existing_file.key(name + ";1")
+        anchor_location = existing_key.fSeekKey + existing_key.fKeylen
+        num_entries = existing.num_entries
+        existing_footer = existing._footer
+        existing_page_list_envelopes = existing.page_list_envelopes
+        existing_field_records = existing._header.field_records
```
ariostas: Here you should use `existing._ntuple.field_records` so that you also consider fields defined in the footer.

Along with this, you should add a test that adds a new field, and then adds another one with the same name.

src/uproot/writing/writable.py
```diff
@@ -2173,17 +2309,204 @@ def extend(self, data):
...
+                )
+            new_field_names = {k for k in data.keys() if k not in existing_keys}
+            if new_field_names and not accept_new_fields:
+                raise ValueError(
+                    f"Data contains fields not in this RNTuple: {sorted(new_field_names)}. "
+                    f"Call add_fields() first, or pass accept_new_fields=True to add them automatically."
+                )
+            elif new_field_names and accept_new_fields:
+                self.add_fields(
+                    {k: numpy.array(list(data[k])).dtype for k in new_field_names}
```
ariostas: Let's avoid Python lists. Maybe something like this is better.
```suggestion
                    {k: numpy.asarray(ak.flatten(ak.Array(data[k]), axis=None)).dtype for k in new_field_names}
```

tests/test_1687_rntuple_update.py
```diff
@@ -401,3 +401,29 @@ def test_ntuple_add_nested_subfield(tmp_path):
...
+def test_ntuple_add_subfield_typed_parent(tmp_path):
+    with uproot.recreate(os.path.join(tmp_path, "test.root")) as f:
+        f["mytuple"] = ak.Array([{"pt": 1.0}, {"pt": 2.0}])
+
+    with uproot.update(os.path.join(tmp_path, "test.root")) as f:
+        with pytest.raises((ValueError, TypeError)):
+            f["mytuple"].add_fields({"pt.x": np.float32})
+
+
+def test_ntuple_add_subfield_to_collection(tmp_path):
```
ariostas, line 732: For this one, I meant something like this. Let's say you have an array of jets like this:

```python
a = ak.Array({"jets": [{"pt": 1., "eta": 0.}, {"pt": 2., "eta": 3.}]})
```
and then you try to add `jets.phi`. It should also fail.


Same thing when you have a variant. This one is harder to come up with something reasonable, but let's say you have something like
```python
b = ak.Array({"variant": [{"jet": {"pt": 1., "eta": 2.}}, 2]})
```
and then you try to add `variant.jet.eta`. That should also fail.

tests/test_1687_rntuple_update.py
```diff
@@ -401,3 +401,29 @@ def test_ntuple_add_nested_subfield(tmp_path):
...
+def test_ntuple_add_subfield_nonexistent_parent(tmp_path):
+    with uproot.recreate(os.path.join(tmp_path, "test.root")) as f:
+        f["mytuple"] = {"x": np.array([1, 2, 3], dtype=np.float32)}
+
+    with uproot.update(os.path.join(tmp_path, "test.root")) as f:
+        with pytest.raises(ValueError, match="not found"):
+            f["mytuple"].add_fields({"nonexistent.phi": np.float32})
+
+
+def test_ntuple_add_subfield_typed_parent(tmp_path):
```
ariostas, line 687: For this test, I meant something different. When an RNTuple is written with ROOT, structs are usually written with their C++ type. For example:

```pycon
>>> import skhep_testdata
>>> import uproot
>>> f = uproot.open(skhep_testdata.data_path("test_nested_structs_rntuple_v1-0-0-0.root"))
>>> rnt = f["ntuple"]
>>> rnt.show()
ntuple (ROOT::RNTuple)
└─ my_struct (TopStruct)
   ├─ i (std::int32_t)
   └─ sub_struct (SubStruct)
      ├─ i (std::int32_t)
      └─ sub_sub_struct (SubSubSruct)
         ├─ i (std::int32_t)
         └─ v (std::vector<std::int32_t>)
```
The structs have C++ typenames (`TopStruct`, `SubStruct`, `SubSubStruct`), so trying to extend any of these should give an error.

src/uproot/writing/writable.py
ariostas: This one is trickier than I expected. I was looking into it with Claude and it seems it's not so easy. Maybe for now we can just leave it like this and leave a comment saying we should look at this again later.

tests/test_1687_rntuple_update.py
ariostas: Btw, you can check the typenames of fields with `.typename`

tests/test_1687_rntuple_update.py
Yokubas: Kept the original `test_ntuple_add_subfield_to_collection` and added two new tests as you suggested — `test_ntuple_add_subfield_to_collection_of_records` for the array of jets case, and `test_ntuple_add_subfield_to_variant` for the variant case. Both fail as expected.

tests/test_1687_rntuple_update.py
Yokubas: The `test_nested_structs_rntuple_v1-0-0-0.root` file can't be opened with uproot.update — it fails with `AssertionError: uuid_version == 1` in `_cascade.py`. I changed the test to use an uproot-created file where` pt` has type_name `double`, which tests the same behavior. Let me know if you'd like me to fix the UUID version issue separately or if there's another file I can use.

ariostas: The test should specifically be so that it can differentiate between a typed and untyped record. In the case where `pt` is a double, it's not a record so it's not enough of a test.

> The test_nested_structs_rntuple_v1-0-0-0.root file can't be opened with uproot.update

That seems like a big problem! Maybe ROOT files written with newer versions of ROOT can't be opened with `update`. We'll definitely have to look into this.

Yokubas: Updated the test to use `test_nested_structs_rntuple_v1-0-0-0.root` now that the UUID fix has been merged. The test now checks all three typed structs — `my_struct` (TopStruct), `sub_struct` (SubStruct), and `sub_sub_struct` (SubSubSruct) — and verifies each raises a `ValueError` with the typename in the message.

src/uproot/writing/writable.py
```diff
@@ -2502,6 +2502,12 @@ def add_fields(self, new_fields):
         footer.cluster_group_record_frames = []
         for cg_idx, cg in enumerate(existing_footer.cluster_group_records):
             ple = existing_page_list_envelopes[cg_idx]
+            if len(ple.pagelinklist) > 1:
+                raise ValueError(
+                    f"add_fields does not yet support RNTuples with multiple clusters per cluster group "
+                    f"(cluster group {cg_idx} has {len(ple.pagelinklist)} clusters). "
+                    f"This is a known limitation that will be fixed in a future version."
+                )
```
ariostas: We should implement this since I think it should be pretty straightforward. But let me know if it's not so easy and we can leave it for later.

src/uproot/writing/writable.py
```diff
@@ -2371,6 +2371,15 @@ def extend(self, data, accept_new_fields=False):
 
             **As a word of warning,** be sure that each call to :ref:`uproot.writing.writable.WritableNTuple.extend` includes at least 100 kB per branch/array. (NumPy and Awkward Arrays have an `nbytes <https://numpy.org/doc/stable/reference/generated/numpy.ndarray.nbytes.html>`__ property; you want at least ``100000`` per array.) If you ask Uproot to write very small TBaskets, it will spend more time working on TBasket overhead than actually writing data. The absolute worst case is one-entry-per-:ref:`uproot.writing.writable.WritableTree.extend`. See `#428 (comment) <https://github.com/scikit-hep/uproot5/pull/428#issuecomment-908703486>`__.
         """
+        # if _column_counts has more columns than _column_keys, reload cascading
+        # this happens after add_fields adds extension columns
+        if len(self._cascading._column_counts) > len(self._cascading._header._column_keys):
+            key = self._file.root_directory._cascading.data.get_key(
+                self._path[-1], 1
+            )
+            reloaded = self._file.root_directory._load_existing_ntuple(key)
```
ariostas: Maybe it's better for `add_fields` to update the necessary things instead of reloading the entire ntuple

src/uproot/writing/writable.py
Yokubas: I implemented multi-cluster support. Instead of raising an error, `add_fields` now loops over all clusters in each cluster group and writes one zero-filled page per new field per cluster, sized to that cluster's entry span. Tested with 3 cluster groups and it works correctly. Note that multiple clusters per cluster group (ROOT behavior) can't be easily tested since ROOT-written files have split encoding which we reject — but the loop handles that case correctly too.

src/uproot/writing/writable.py
Yokubas: Reduced the file reload — `_existing_field_records` and `_column_counts` are now updated directly in memory. Still need to read `_existing_footer`, `_existing_page_list_envelopes`, and `_header._akform` from file because the writable footer uses `cluster_group_record_frames` while the next `add_fields` call needs `cluster_group_records` from the read-only footer. Could avoid this by adding a `cluster_group_records` property to the writable footer that mirrors `cluster_group_record_frames` — would that be the preferred approach?

src/uproot/writing/writable.py
```diff
@@ -2173,17 +2354,316 @@ def extend(self, data):
...
+                actual_field_name,
+                type_name,
+            )
+            footer.extension_field_record_frames.append(new_field)
+            new_col = cnt.NTuple_Column_Description(
+                type_num, type_size, next_field_id, 0, 0
+            )
+            footer.extension_column_record_frames.append(new_col)
+
+            new_data = numpy.zeros(num_entries, dtype=numpy.dtype(ak_primitive))
```
ariostas: Sorry, I missed this part on my first review. We actually, don't want to explicitly backfill the new fields with zeros. This is taken care by attaching a deferred column to the new field. So the existing cluster groups stay as they were, with no new columns for the new fields, and new cluster groups will start to contain the new column. So you'll have to modify `NTuple_Column_Description` so that it can also take a first-element index (see https://github.com/root-project/root/blob/master/tree/ntuple/doc/BinaryFormatSpecification.md#column-description).

So now it makes sense why there were some issues with cluster groups with multiple clusters. This way, since you're leaving old cluster groups intact and Uproot (currently) only writes cluster groups with a single cluster then you don't have to worry about multiple clusters in a cluster group.

Yokubas: Implemented the deferred column approach — `NTuple_Column_Description` now accepts `first_element_index`, and when it's > 0 the DEFERRED flag is set and the value is serialized. `add_fields` no longer backfills zeros — old cluster groups stay completely untouched, and the reader pads zeros automatically via the deferred column mechanism. Also fixed the reload path to preserve `first_element_index` when copying extension column records. 32 tests still pass.

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1049,190 @@ def get_chunk(start, stop):
...
+        name = key.name.string
+        if self.file_path is None:
+            raise TypeError(
+                "uproot.update() on a file-like object does not support accessing "
+                "existing RNTuples; use uproot.update() with a file path instead."
+            )
+        # TODO: opening the file again in read mode to access existing metadata is
+        # a bit awkward since the file is already open in write mode. We should
+        # look into a better way to do this in the future.
+        existing_file = uproot.open(self.file_path, minimal_ttree_metadata=False)
```
ariostas: Maybe we can also use `_ReadForUpdate` here.

src/uproot/writing/_cascadentuple.py
```diff
@@ -242,21 +242,33 @@ def serialize(self):
...
             self.bits_on_disk,
             self.field_id,
-            self.flags,
+            int(flags),
             self.repr_index,
         )
+        if self.first_element_index > 0:
+            import struct
+
+            header_bytes += struct.pack("<Q", int(self.first_element_index))
```
ariostas: Let's reuse `_rntuple_first_element_index_format`

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1049,213 @@ def get_chunk(start, stop):
...
+            )
+
+        _readforupdate_rn = uproot.writing._cascade._ReadForUpdate(
+            self._file.file_path,
+            self._file.uuid,
+            _get_chunk_rn,
+            self._file._cascading.tlist_of_streamers,
+        )
+        _readforupdate_rn.options = dict(uproot.reading.open.defaults)
+        _readforupdate_rn.options["minimal_ttree_metadata"] = False
```
ariostas: Is this needed?

tests/test_1687_rntuple_update.py
```diff
@@ -0,0 +1,1063 @@
...
+
+def test_ntuple_add_subfield_to_variant(tmp_path):
+    with uproot.recreate(os.path.join(tmp_path, "test.root")) as f:
+        f["mytuple"] = ak.Array(
+            {"variant": ak.Array([{"jet": {"pt": 1.0, "eta": 2.0}}, 2])}
+        )
+
+    with uproot.update(os.path.join(tmp_path, "test.root")) as f:
+        with pytest.raises((ValueError, AssertionError)):
+            f["mytuple"].add_fields({"variant.jet.eta": np.float32})
```
ariostas, line 760: This fails, as expected, but the error is confusing. It says `"This should not have happened. Please report this issue to the Uproot developers."`. It would be good to make it fail in a better way that just says that you can't extend an variant.

ariostas:
This is looking great! I left a comment with a more minor thing, but everything seems to be working well.

tests/test_1687_rntuple_update.py
Yokubas: Fixed — it was actually broader than the add_fields case you flagged, plain access to a variant-field RNTuple via `uproot.update()` was crashing the same way. Now raises a clear `NotImplementedError` telling you to use `uproot.open()` instead for reading

tests/test_1687_rntuple_update.py
ariostas, line 1: let's also add the license header here

ariostas:
I added a commit to resolve conflicts with main.

ariostas (approved):
Thanks for the hard work, @Yokubas! It looks great. This is some really useful functionality to have.

PR #1689, opened 2026-07-23
fix: uproot.update should accept other valid UUID versions

ariostas:
During #1687 it was discovered that `uproot.update` didn't work on ROOT files created with new versions of ROOT. It turned out that ROOT just writes nil UUID (version 0) now, but Uproot expected the version to be 1. So that part was just fixed by removing the constraint.

~~There was another issue while reading the streamers. Claude worked on it for a long time, and the conclusion seems to be that it was a bug in ROOT 6.35. It implemented a workaround that I think is good enough. Maybe the switch to RootFileSpec can help with this.~~

Welp, I spent a lot of time on this, but it turned out there were some stale files on my laptop that were complicating things a lot.

### DIFF
--- a/src/uproot/writing/_cascade.py
+++ b/src/uproot/writing/_cascade.py
@@ -2109,7 +2109,7 @@ class FileHeader(CascadeLeaf):
             compression_code,
             info_location,
             info_num_bytes,
-            uuid_version,
+            _uuid_version,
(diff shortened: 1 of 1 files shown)


ianna (approved):
@ariostas - Indeed -- good call.

PR #1690, opened 2026-07-27
feat: Add in-place TTree branch addition and row extension

Yokubas:
## Summary

Implements in-place modification of existing TTrees:

- `f["tree"].add_branches({"new_x": array1, "new_y": array2, ...})` — add one or more new branches back-filled with provided data
- `f["tree"].extend({"x": array1, "y": array2})` — append new entries to one or more existing branches
- `f["tree"].extend({"x": array1, "new_y": array2}, accept_new_fields=True)` — auto-add new branches back-filled with zeros, then extend

## How it works

Uses uproot's cascade machinery instead of manual byte patching:

- For `extend`: deserializes the existing TTree using uproot's reading side (branch members, cursor positions), reconstructs a `ct.Tree` cascade object, then delegates to the existing cascade write machinery — which appends new baskets and patches `fBasketSeek`, `fBasketBytes`, `fBasketEntry`, `fWriteBasket`, `fEntryNumber`, and `fEntries` in the TTree blob
- For `add_branches`: creates new branch dict via `_branch_np`, calls `write_anew` to rewrite the TTree metadata blob with the new branch included, then writes one basket per new branch. Existing basket data is never touched — the metadata blob just gains new branch headers and the basket seek arrays are updated
- For `accept_new_fields=True`: calls `add_branches` with zeros for existing entries, then extends with the provided data using the updated cascade
- Metadata positions (`metadata_start`, `basket_metadata_start`) are derived structurally, by walking the same layout `write_anew` itself emits (`Tree._build_out()`), rather than by searching the blob for byte patterns. The original byte-pattern-search approach broke whenever the value being searched for was `0` (e.g. a freshly created tree with no baskets yet, or a branch that had never been extended) — a search for zero bytes matches arbitrary unrelated data and silently corrupted the file. Deriving the offsets from the same code path that writes them removes that whole class of bug.

## Known limitations

- Only top-level branches supported — adding subfields (e.g. `particle.phi`) is not yet implemented
- `add_branches`/`extend` for TBranchElement (split-object) files is not yet supported — **this is now enforced with a clean `NotImplementedError` and covered by tests; previously it could crash on plain access to such a file, or silently desynchronize entry counts across branches**
- File-like objects not supported (requires file path for re-reading metadata)
- `add_branches`/`extend` require every branch to agree on basket count and capacity. This is always true for a tree Uproot itself wrote (and hasn't touched with `add_branches` since), but a ROOT-written tree commonly has divergent per-branch basket counts (branches flush baskets at different rates depending on per-entry size) — this case is now detected and rejected with a clear `NotImplementedError` rather than corrupting the file.

## Fixes made during review

Several data-corruption and crash bugs were found and fixed after the initial implementation, all confirmed with reproductions and covered by regression tests:

- Jagged branch `extend` in update mode wrote garbage instead of the real values — `_load_existing_ttree` derived a jagged branch's dtype from the wrong place (the interpretation's own `numpy_dtype`, which is `object` for a jagged array) instead of its content dtype.
- Extending a tree with zero baskets (freshly `mktree`'d, never extended) corrupted the file, for the same "search for zero bytes" reason described above.
- String branches couldn't be extended after reopening a file — `_load_existing_ttree` had no handling for the `AsStrings` interpretation.
- Fixed-size array branches (e.g. `float[3]`) crashed on plain access, not just extend — `_load_existing_ttree` didn't account for a branch's shape.
- `add_branches` on a tree that already had more than one basket wrote an incorrect `fWriteBasket` for the new branch (the whole tree's basket count instead of the new branch's real count of 1), corrupting the new branch's layout.
- `add_branches` leaked a stale cache entry in `WritableFile._trees` on every call after the tree relocated; now cleaned up via the same `_move_tree` path `extend`'s basket-capacity growth already uses.
- `add_branches` unconditionally re-read and re-parsed the whole tree from disk on every call even though the in-memory state was already current; now reuses it directly.
- A jagged/counter-branch could be silently misattached to an unrelated same-named branch (e.g. a string branch `"id"` incorrectly paired with an unrelated int branch `"nid"`); the counter-inference heuristic now excludes string branches.

## Tests (36 passing)
- Basic `add_branches` and `extend` for simple TBranch files
- Multiple branches added or extended in a single call
- Different dtypes (float32, int32)
- Preserves existing data across operations
- Sequential `add_branches` calls across separate sessions
- `add_branches` then `extend` in the same session
- Multiple `extend` calls across separate sessions
- `extend` after `add_branches` in a new session
- `accept_new_fields` behavior
- ROOT verification for simple TBranch files
- Corner cases: nonexistent branch, mismatched lengths, nonexistent tree, missing branch in extend, accept_new_fields error without flag
- Jagged array extend (counter branch not required from user)
- Extend after many extends (fMaxBaskets > 10)
- Jagged branch extend across a fresh `uproot.update()` session (regression test for the dtype-corruption fix)
- Extending a zero-basket tree, and a zero-basket tree with multiple branches (regression tests for the byte-search corruption fix)
- String branch extend, including from zero baskets (regression tests for the missing `AsStrings` handling)
- Fixed-size array branch access (regression test for the shape-handling crash)
- TBranchElement access without crashing, and clean `NotImplementedError` from `extend`/`add_branches` on such files (regression tests)
- Divergent per-branch basket counts correctly rejected, verified against a real PyROOT-written fixture with mismatched basket counts
- `add_branches` after multiple extends (regression test for the `fWriteBasket` corruption fix), including the correct rejection of a follow-up `extend` once basket counts diverge
- `add_branches` does not leak `WritableFile._trees` cache entries (regression test)
- `add_branches` does not re-read the tree from disk (regression test)
- Counter-branch inference does not misattach a string branch to a coincidentally-named sibling (regression test)


src/uproot/writing/writable.py
```diff
@@ -1721,6 +1723,420 @@ def __init__(self, path, file, cascading):
...
+            raise TypeError("'source' must be the name of a TTree")
+
+        tree_key = existing_file.key(source + ";1")
+        key_seek = tree_key.fSeekKey
+        key_len = tree_key.fKeylen
+        compression = existing_file._file.compression
+        file_end = existing_file._file.fEND
+
+        # get directory key info
+        with uproot.update(file_path) as tmp:
```
ariostas: You probably don't need this `uproot.update` since you already have the file open in this mode.

src/uproot/writing/writable.py
```diff
@@ -1721,6 +1723,420 @@ def __init__(self, path, file, cascading):
...
+                f"TTree {source!r} not found in file {file_path}"
+            ) from None
+
+        tree_key = existing_file.key(source + ";1")
+        key_seek = tree_key.fSeekKey
+        key_len = tree_key.fKeylen
+        compression = existing_file._file.compression
+        file_end = existing_file._file.fEND
+
+        with uproot.update(file_path) as tmp:
```
ariostas: Same here.

src/uproot/writing/writable.py
```diff
@@ -1721,6 +1723,420 @@ def __init__(self, path, file, cascading):
...
+        num_branches = len(old_ttree.branches)
+        last_branch = list(old_ttree.branches)[-1]
+        c = last_branch.cursor.copy()
+        c.skip_after(last_branch)
+        insertion_point = c.index
+        existing_file.close()
+
+        # find fBranches TObjArray bcnt
+        tobjarray_bcnt_pos = None
+        for i in range(190, 220):
```
ariostas: where did these numbers come from? They seem file-dependent

src/uproot/writing/writable.py
```diff
@@ -1721,6 +1723,420 @@ def __init__(self, path, file, cascading):
...
+            dir_key_big = dir_key.big
+
+        chunk, cursor = tree_key.get_uncompressed_chunk_cursor()
+        orig_raw = bytearray(chunk.raw_data.tobytes())
+        fEntries = old_ttree.member("fEntries")
+        fMaxBaskets = list(old_ttree.branches)[0].member("fMaxBaskets")
+        existing_file.close()
+
+        # find TTree fEntries position in blob
+        fentries_pos = orig_raw.find(struct.pack(">q", fEntries))
```
ariostas: Seems a bit risky. It would be better to find something more reliable.

src/uproot/writing/writable.py
```diff
@@ -1721,6 +1723,420 @@ def __init__(self, path, file, cascading):
         self._file = file
         self._cascading = cascading
 
+    def add_branches(self, branches):
```
ariostas, line 2176: This function should check that all new branches have the same length as the tree.

src/uproot/writing/writable.py
```diff
@@ -1883,10 +2299,12 @@ def num_baskets(self) -> int:
         """
         return self._cascading.num_baskets
 
-    def extend(self, data):
+    def extend(self, data, *, accept_new_fields=False):
```
ariostas, line 2356: This function should check that the input data contains all existing branches. It would be inconsistent to only extend some branches.

src/uproot/writing/writable.py
```diff
@@ -1721,6 +1723,420 @@ def __init__(self, path, file, cascading):
...
+        struct.pack_into(">i", raw_dir, 6, new_objlen)
+        if dir_key_big:
+            struct.pack_into(">q", raw_dir, 18, file_end)
+        else:
+            struct.pack_into(">i", raw_dir, 18, file_end)
+        self._file.sink.write(dir_key_location, bytes(raw_dir))
+
+        # update fEND
+        new_file_end = file_end + new_nbytes
+        self._file.sink.write(12, struct.pack(">i", new_file_end))
```
ariostas: The size of this depends on whether the file is over 2 GB.

ariostas (commented):
This is looking really promising! Thanks for the hard work!

I left some comments for specific locations.

But one more general comment is that this seems to be doing a lot of manual scanning and patching. It would be better to try to do what you did for the RNTuple one. Deserializing what you need to construct a WritableTree (with `self._cascading` properly built) and then let the existing functionality take care of as much of the serialization as possible.

src/uproot/writing/writable.py
Yokubas: The `range(190, 220)` was a hardcoded byte range based on where the `fBranches` TObjArray bcnt happened to be in the test files. Replaced it with a proper scan of the whole blob looking for a 4-byte value with the `kByteCountMask` bit (`0x40000000`) set, immediately followed by a 2-byte version=3 (which is the TObjArray version). This uniquely identifies the `fBranches TObjArray` header regardless of file layout.

src/uproot/writing/writable.py
Yokubas: Instead of searching for `fEntries` value alone (which could match elsewhere in the blob), now searching for the unique sequence `fEntries + fTotBytes + fZipBytes` — three consecutive 8-byte fields in the TTree header that together form a unique pattern.

src/uproot/writing/writable.py
Yokubas: Fixed — now uses `struct.pack(">q", ...)` (8-byte) for files where `fileheader.big` is True (files >= 2GB), and `struct.pack(">i", ...)` (4-byte) for smaller files, consistent with how uproot's cascade serializes `fEND`.

Yokubas:
> This is looking really promising! Thanks for the hard work!
> 
> I left some comments for specific locations.
> 
> But one more general comment is that this seems to be doing a lot of manual scanning and patching. It would be better to try to do what you did for the RNTuple one. Deserializing what you need to construct a WritableTree (with `self._cascading` properly built) and then let the existing functionality take care of as much of the serialization as possible.

Thanks for the feedback! I understand the concern about manual byte patching — it's fragile and hard to maintain.

For the RNTuple case, we could reconstruct a WritableNTuple because uproot already had full read/write support for the format. For TTree, the challenge is that _cascadetree.py handles writing new trees from scratch, but there's no path to deserialize an existing TTree blob back into a WritableTree with self._cascading set.

Would you be able to point me toward what would need to change in _cascadetree.py to support this? Specifically — how to reconstruct the cascade objects (branches, baskets, etc.) from an existing serialized TTree so we can use the existing extend/write machinery?

eduardo-rodrigues:
Hello @Yokubas, cc @ariostas as reviewer - there are many people watching this important package and I'm sure I'm not the only one noticing that you often get trivial pre-commit updates on your commits. This results in many more email notifications, which could be trivially spared to everyone if you would run pre-commit locally. This standard way of proceding is even mentioned in the [CONTRIBUTING file](https://github.com/scikit-hep/uproot5/blob/main/CONTRIBUTING.md). Could you kindly follow the guidelines and save us all tens of meaningless email notifications? Thank you in advance.

Yokubas:
Apologies for the noise! I've set up pre-commit locally now — won't happen again.

src/uproot/writing/writable.py
```diff
@@ -1721,172 +1873,121 @@ def __init__(self, path, file, cascading):
...
+        Adds new branches to this TTree in-place. Only the new branch data and
+        an updated TTree header are written; existing data is never touched.
+        Works with both simple TBranch and TBranchElement files.
 
-    def __exit__(self, exception_type, exception_value, traceback):
-        self._file.sink.__exit__(exception_type, exception_value, traceback)
+        .. code-block:: python
 
-    @property
-    def compression(self):
```
ariostas, line 1808: Seems like you accidentally deleted a bunch of properties here. We should keep those

src/uproot/writing/writable.py
```diff
@@ -1910,6 +2011,34 @@ def extend(self, data):
...
+        if self._cascading is None:
+            raise RuntimeError(
+                "_cascading is None — this should not happen; please report this bug"
+            )
+        # validate branches
+        if isinstance(data, dict):
+            existing_names = [bd["fName"] for bd in self._cascading._branch_data]
+            new_fields = {k: v for k, v in data.items() if k not in existing_names}
+            missing = [b for b in existing_names if b not in data]
+            if missing:
```
ariostas, line 2519: As you saw from the failed tests, there is one exception here. When you have a jagged array it creates another branch with the same name, but with an `n` prefix that stores how many elements each row has. And they are automatically generated by Uproot/ROOT, so the user is not expected to pass them in.

ariostas: I think also records need to be skipped. So the fix could probably be
```python
existing_names = [bd["fName"] for bd in self._cascading._branch_data if bd.datum["kind"] not in ("counter", "record")]
```

src/uproot/writing/writable.py
Yokubas: Restored the accidentally deleted `WritableTree` properties (`__repr__`, `path`, `object_path`, `file_path`, `file`, `close`, `closed`, `__enter__`, `__exit__`, `compression`) — they got removed when I was cleaning up `_extend_inplace`

src/uproot/writing/writable.py
Yokubas: Counter and record branches are now excluded from the `missing` check in `extend()` — `existing_names` now filters by `bd.datum["kind"] not in ("counter", "record")` as you suggested. Also fixed `_load_existing_ttree` to correctly reconstruct counter branch kind and references so that jagged array extend works in `uproot.update` sessions too. 22 tests passing

Yokubas: One thing to flag — `test_writable_vs_readable_tree` in `test_0406_write_a_ttree.py` expects a `TypeError` when accessing a ROOT-written TTree via `uproot.update`. This was the old behavior when uproot couldn't handle existing TTrees. Since our PR now supports this, the test expectation is outdated — should this test be updated to reflect the new behavior?

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1053,195 @@ def get_chunk(start, stop):
...
+        finally:
+            existing_file.close()
+
+        dir_key = self._cascading.data.get_key(name, 1)
+        freesegments = self._file._cascading.freesegments
+
+        casc = ct.Tree.__new__(ct.Tree)
+        casc._directory = self._file._cascading.rootdirectory
+        casc._name = name
+        casc._title = ""
```
ariostas: You should use the title from the existing tree

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1053,195 @@ def get_chunk(start, stop):
...
+        casc._freesegments = freesegments
+        casc._branch_data = branch_data
+        casc._branch_lookup = branch_lookup
+        casc._basket_capacity = 10
```
ariostas: This should also come from the existing tree. This fails.

```python
import numpy as np, uproot

with uproot.recreate("t.root") as f:
    f.mktree("tree", {"x": np.float32})
    for i in range(12):
        f["tree"].extend({"x": np.full(5, i, np.float32)})

with uproot.update("t.root") as f:
    f["tree"].extend({"x": np.full(5, 99, np.float32)})
```

src/uproot/writing/writable.py
ariostas: Yeah, let's update that test

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1053,195 @@ def get_chunk(start, stop):
...
+            "u8": "Q",
+            "u2": "H",
+            "u1": "B",
+        }
+
+        # flush and read via BytesIO to avoid OS caching issues
+        self._file.sink.flush()
+        _sink_file = self._file.sink._file
+        _sink_file.seek(0)
+        _buf = io.BytesIO(_sink_file.read())
```
ariostas: We shouldn't use this since it ends up loading the entire file into memory, which could be a problem if it's a big file.

I attached below what Claude suggested. I like the suggestion because it follows the way that other objects are handled already, and also we get rid of `uproot.open`, which we had discussed earlier to be a little weird to use when you already had an open file.

So the code should look like this part that's already there:

https://github.com/scikit-hep/uproot5/blob/04d8bf1a7c145f1b85776459352fd6f32c674fe0/src/uproot/writing/writable.py#L1029-L1054

### `_load_existing_ttree` shouldn't read the whole file

```python
_buf = io.BytesIO(_sink_file.read())
existing_file = uproot.open(_buf, minimal_ttree_metadata=False)
```

This materializes the entire ROOT file in memory to read one TTree's metadata blob, and it happens inside `uproot.update`, so it coexists with the file's own write buffers. `add_branches` does it twice (`writable.py:1093` and `:2121`). On a multi-GB file this is fatal.

The instinct behind it is right — naively `uproot.open(path)` on a file that's open for writing gives you a `MemmapSource` that won't see subsequent growth, and on Windows can fail outright. But `_get` already solves this correctly twenty lines below, in its `else` branch: targeted `sink.read(start, n)` + `Chunk.wrap` + `_ReadForUpdate` + `ReadOnlyKey.get()`. That's how `uproot.update` reads every other preexisting object.

Pointing that pattern at a TTree key gives everything `_load_existing_ttree` consumes — the deserialized model, per-branch `cursor.index`, `cursor._refs`, `fMaxBaskets`, `fBasketSeek`, plus the uncompressed blob via `readonlykey.get_uncompressed_chunk_cursor()` for the `raw.find(...)` searches. On a 22,537-byte test file: 3,937 bytes read instead of 22,537, and that figure is fixed metadata, so it stays flat as the file grows.

**Two things needed:**

1. `_ReadForUpdate` (`_cascade.py:1054`) has no `.options`, but `Model_TTree_v20`/`Model_TBranch_v13` `read_members` both consult `file.options["minimal_ttree_metadata"]`. Currently raises `AttributeError`. One-line fix: give it an options dict defaulting to `uproot.reading.open.defaults`.

2. Wrap the chunk at absolute offsets — `Chunk.wrap(..., start=key.seek_location)` and `Cursor(key.seek_location, origin=key.num_bytes)`. The sibling `else` branch wraps at 0, which is fine there because it only calls `.get()`, but `get_uncompressed_chunk_cursor` passes `data_cursor.index` to `_file.chunk()` as a file position.

**Falls out for free:** no second view of a concurrently-written file (which is what the BytesIO copy was working around); the hardcoded `";1"` cycle goes away since you already hold the real `key`; the root-directory name lookup goes away, which fixes `f["sub/tree"]`; and the `file_path is None` guard becomes unnecessary, removing "file-like objects not supported" from Known Limitations.

Separately — is `minimal_ttree_metadata=False` deliberate? I compared both modes on a uproot-written tree and two ROOT-written files: `cursor.index`, `cursor._refs`, and `fLeaves[0].fMaximum` are identical, and flipping both call sites to the default leaves all 1014 tests passing. If there's a case where it matters it'd be worth a comment; otherwise it's extra deserialization (an `fBaskets` TObjArray per branch) on a path that now runs for every `f["tree"]` in update mode.

src/uproot/writing/writable.py
```diff
@@ -1910,6 +2209,66 @@ def extend(self, data):
...
+                    )
+                zeros = {
+                    k: numpy.zeros(
+                        self._cascading._num_entries, dtype=numpy.asarray(v).dtype
+                    )
+                    for k, v in new_fields.items()
+                }
+                self.add_branches(zeros)
+                self._cascading.extend(self._file, self._file.sink, data)
+                return
```
ariostas: You can remove these two lines, since the next line will take care of it

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1053,221 @@ def get_chunk(start, stop):
...
+            refs_list = list(b.cursor._refs.keys())
+            try:
+                dtype = b.interpretation.numpy_dtype.newbyteorder(">")
+            except AttributeError:
+                # TBranchElement or other complex branch — skip
+                continue
+            sc = _dtype_to_struct.get(dtype.kind + str(dtype.itemsize), "f")
+            # detect counter branches (e.g. njets for jagged jets array)
+            _branch_names = [br.name for br in branches]
+            _is_counter = b.name.startswith("n") and b.name[1:] in _branch_names
```
ariostas: is it possible to check if the kind is `"counter"` here instead of doing this?

src/uproot/writing/writable.py
```diff
@@ -1054,6 +1053,221 @@ def get_chunk(start, stop):
...
+                "fMaxEntries",
+                "fMaxEntryLoop",
+                "fMaxVirtualSize",
+                "fAutoSave",
+                "fAutoFlush",
+                "fEstimate",
+            ]
+        }
+
+        dir_key = self._cascading.data.get_key(name, 1)
```
ariostas: The cycle number probably shouldn't be hardcoded

src/uproot/writing/writable.py
```diff
@@ -1883,10 +2097,111 @@ def num_baskets(self) -> int:
...
+                f["tree"].add_branches({"new_branch": np.ones(100, dtype=np.float32)})
+        """
+        if self._file.sink.closed:
+            raise ValueError("cannot modify a TTree in a closed file")
+
+        source = self._path[-1]
+
+        # validate all branches have same length as existing tree
+        key = self._file._cascading.rootdirectory.data.get_key(source, 1)
+        casc = self._file.root_directory._load_existing_ttree(key)._cascading
```
ariostas: Do these two lines work if the tree is in a subdirectory?

Yokubas: Fixed the subdirectory issue — `add_branches` now navigates through `self._path[:-1]` to find the correct directory instead of always using `rootdirectory`. Also fixed `casc._directory` to point to the correct directory so `write_anew` works properly. Tested with a tree in a subdirectory and it works correctly

tests/test_1690_ttree_inplace.py
ariostas, line 1: Let's add the usual license header that the rest of the tests have

tests/test_1690_ttree_inplace.py
```diff
@@ -0,0 +1,874 @@
...
+            f["tree"].add_branches({"new_branch": np.ones(50, dtype=np.float32)})
+
+
+def test_add_branch_nonexistent_tree(tmp_path):
+    with uproot.recreate(os.path.join(tmp_path, "test.root")) as f:
+        f.mktree("tree", {"x": np.float32})
+        f["tree"].extend({"x": np.ones(100, dtype=np.float32)})
+
+    with uproot.update(os.path.join(tmp_path, "test.root")) as f:
+        with pytest.raises(Exception):
```
ariostas: It's better to specify which kind of exception it raises

tests/test_1690_ttree_inplace.py
```diff
@@ -0,0 +1,874 @@
...
+            )
+
+
+def test_extend_nonexistent_branch(tmp_path):
+    with uproot.recreate(os.path.join(tmp_path, "test.root")) as f:
+        f.mktree("tree", {"x": np.float32})
+        f["tree"].extend({"x": np.ones(100, dtype=np.float32)})
+
+    with uproot.update(os.path.join(tmp_path, "test.root")) as f:
+        with pytest.raises(Exception):
```
ariostas: Same here

ariostas (approved):
Thanks for all the hard work, @Yokubas! This is great! It's awesome that not only did you finish the project Zoe was working on, but you also ended up also adding the functionality to add more entries to the TTree, which we had not even planned.

PR #1691, opened 2026-07-28
fix: intermitten issue with fsspec on free-threaded Python

ariostas:
There was an intermittent issue in free-threaded Python where the fsspec source would sometimes not be closed properly, which raised a warning. I had Claude debug this, and it turned out to be a surprisingly simple fix. We don't actually use the `self._fo` object, so we can just get rid of it completely.

### DIFF
--- a/tests/test_0916_read_from_s3.py
+++ b/tests/test_0916_read_from_s3.py
@@ -16,7 +16,7 @@ def test_s3_fail():
         with uproot.source.fsspec.FSSpecSource(
             "s3://pivarski-princeton/does-not-exist", anon=True
         ) as source:
-            uproot._util.tobytes(source.chunk(0, 100).raw_data)
+            source.chunk(0, 100)
 
 
 @pytest.mark.network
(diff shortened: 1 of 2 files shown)


tests/test_0916_read_from_s3.py
```diff
@@ -16,7 +16,7 @@ def test_s3_fail():
         with uproot.source.fsspec.FSSpecSource(
             "s3://pivarski-princeton/does-not-exist", anon=True
         ) as source:
-            uproot._util.tobytes(source.chunk(0, 100).raw_data)
+            source.chunk(0, 100)
```
ariostas, line 19: This was changed because `uproot._util.tobytes` doesn't exist anymore, and the `FileNotFound` error now occurs later on.

ianna (approved):
@ariostas — great! I’ll approve it. Why was it causing a problem if we didn’t use it?

ariostas:
> Why was it causing a problem if we didn’t use it?

Because it held a handle to an open file. And the thing is that garbage collection in 3.14t is not deterministic, so sometimes that handle was staying alive after the Uproot was supposed to close the file.

PR #1695, opened 2026-08-04
test: serve S3 tests from a local moto server

### LINKED ISSUE #1693: Moving pythia_ppZee_run17emb.picoDst.root
I just started getting large data transfer charges on AWS; someone is downloading files from there at a high rate. `https://pivarski-princeton.s3.us-east-1.amazonaws.com/pythia_ppZee_run17emb.picoDst.root` is one of those files, so I need to move it to another server. I created a [Zenodo page for it](https://zenodo.org/records/21777191), so a new download URL is

```bash
wget 'https://zenodo.org/records/21777191/files/pythia_ppZee_run17emb.picoDst.root?download=1' -O pythia_ppZee_run17emb.picoDst.root
```

I will delete the file from AWS at the end of the month (August 31, 2026). If you're using it, please update your sources!

ariostas:
Closes #1693.

The S3 tests were reading `pythia_ppZee_run17emb.picoDst.root` (872 MB) out of the `pivarski-princeton` bucket on every CI run, which is running up data transfer charges for its owner. That file is moving to [Zenodo](https://zenodo.org/records/21777191) and will be deleted from AWS on August 31, 2026.

It was the only object in that bucket, and all three tests using it exist to exercise the `s3://` protocol, so the new Zenodo HTTPS URL isn't a substitute. Instead of repointing them at another remote copy, this runs them against an in-process [moto](https://github.com/getmoto/moto) S3 server.

- New session-scoped `s3_server` fixture in `tests/conftest.py` (alongside the existing `http_server` / `xrootd_server` fixtures): starts `ThreadedMotoServer` on an ephemeral port, creates a bucket, and uploads two copies of `uproot-HZZ.root` from `skhep_testdata`. It yields `(bucket_url, storage_options)`, with fake key/secret passed explicitly so a developer's real AWS credentials are never picked up.
- `test_s3_fail`, `test_read_s3`, `test_open_fsspec_s3`, and `test_fsspec_globbing_s3` now use the fixture. The globbing test asserts 2 matches in a bucket the fixture controls, rather than depending on the remote bucket's contents not changing.
- Dropped `@pytest.mark.network` from these tests — they're local now, and run in ~2.5s.
- Added `moto[s3,server]` to the `test` dependency group.

No reference to the `pivarski-princeton` bucket remains in the repo.

The one trade-off: we no longer read a real remote S3 object anywhere in the test suite. If that's worth keeping, the Zenodo URL could be added as an `https://` fsspec test, but it can't cover the `s3://` path.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ test = [
   "aiohttp",
   "deflate",
   "isal",
+  "moto[s3,server]",
   "paramiko",
   "pytest-rerunfailures",
   "rangehttpserver",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -106,19 +106,19 @@ def s3_server():
         # wheels, so the resolver falls back to a years-old moto
         pytest.skip("moto is too old to report which port it is listening on")
 
+    bucket = "uproot-test"
     server = moto_server.ThreadedMotoServer(ip_address="127.0.0.1", port=0)
     server.start()
-    _, port = server.get_host_and_port()
-
-    bucket = "uproot-test"
-    storage_options = {
-        # moto does not check these, but botocore insists on having them
-        "key": "testing",
-        "secret": "testing",
-        "client_kwargs": {"endpoint_url": f"http://127.0.0.1:{port}"},
-    }
 
     try:
+        _, port = server.get_host_and_port()
+        storage_options = {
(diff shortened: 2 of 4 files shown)


ianna (approved):
@ariostas - Great! Thanks for fixing it.

PR #1702, opened 2026-08-11
docs: add descriptions of reading and writing options

tomeichlersmith:
While working on updating the documentation for coffea (https://github.com/scikit-hep/coffea/pull/1615) I saw a few options that were undocumented, so I'm adding a few lines of documentation to them based off my understanding of the code (mostly other docstrings already existing within uproot).


src/uproot/reading.py
```diff
@@ -526,14 +527,22 @@ class ReadOnlyFile(CommonFileMethods):
...
        The maximum number of elements to be requested in a single vector read, when using XRootD.
     * num_workers (int; 1)
+        Number of tasks to spawn for reading and decompression.
     * use_threads (bool; False on the emscripten platform (i.e. in a web browser), else True)
+        Use multi-threading when spawning workers.
     * num_fallback_workers (int; 10)
     * begin_chunk_size (memory_size; 403, the smallest a ROOT file can be)
+        Size of first chunk that we attempt to read in bytes.
     * minimal_ttree_metadata (bool; True)
+        Defer the reading of metadata until they are accessed.
```
ariostas: Some parts actually get completely skipped. So it's more like "Skip rarely used metadata, and defer reading of embedded TBaskets."

src/uproot/reading.py
```diff
@@ -526,14 +527,22 @@ class ReadOnlyFile(CommonFileMethods):
...
     * handler (:doc:`uproot.source.chunk.Source` class; None)
+        Class implementing reading from the data source.
+        If None, deduced from input file type.
     * timeout (float for HTTP, int for XRootD; 30)
+        The time in seconds to wait before giving up on the connection.
+        Ignored for non-internet sources like local file paths.
     * max_num_elements (None or int; None)
        The maximum number of elements to be requested in a single vector read, when using XRootD.
     * num_workers (int; 1)
+        Number of tasks to spawn for reading and decompression.
```
ariostas: This is only for reading, not decompression. Maybe also mention that only select source types use this.

src/uproot/reading.py
```diff
@@ -526,14 +527,22 @@ class ReadOnlyFile(CommonFileMethods):
     Options (type; default):
 
     * handler (:doc:`uproot.source.chunk.Source` class; None)
+        Class implementing reading from the data source.
+        If None, deduced from input file type.
     * timeout (float for HTTP, int for XRootD; 30)
```
ariostas: 30 is a stale default. Now it uses the default for that backend, so let's also update this.

src/uproot/reading.py
```diff
@@ -526,14 +527,22 @@ class ReadOnlyFile(CommonFileMethods):
...
        The maximum number of elements to be requested in a single vector read, when using XRootD.
     * num_workers (int; 1)
+        Number of tasks to spawn for reading and decompression.
     * use_threads (bool; False on the emscripten platform (i.e. in a web browser), else True)
+        Use multi-threading when spawning workers.
     * num_fallback_workers (int; 10)
     * begin_chunk_size (memory_size; 403, the smallest a ROOT file can be)
+        Size of first chunk that we attempt to read in bytes.
     * minimal_ttree_metadata (bool; True)
+        Defer the reading of metadata until they are accessed.
```
ariostas, line 560: For completeness let's also document `num_fallback_workers` and `http_max_header_bytes` for completeness.

src/uproot/reading.py
```diff
@@ -526,14 +527,22 @@ class ReadOnlyFile(CommonFileMethods):
     Options (type; default):
```
ariostas, line 543: This block is also in other places of the codebase. The most important place is in `open` above, since most people will be using `uproot.open`, but it would also be good to update it in the other ones. (You can check where with `grep -r "Options (type; default)" src`)

tomeichlersmith:
Thanks @ariostas , I'll get to the last few broader comments this afternoon

tomeichlersmith:
Alright, I realized that instead of copying the same text everywhere, we can `..include` it to make it easier to update in the future. I can drop the `..include` and manually copy the same descriptions everywhere if you prefer.

ariostas:
@tomeichlersmith sorry for the late reply. I like the idea of the `..include` to avoid repetition, but in this case it doesn't work well since doing something like `help(uproot.open)` won't show the actual content.

A couple of other things:
- `uproot.update` doesn't accept the same options. `compression` and `initial_streamer_bytes` are not accepted.
- `behaviors/RNTuple.py` also needs to be updated.

ariostas:
By the way, don't worry about the CI errors. They're obviously unrelated.

The one that is real the for the PR title. We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for the titles. So in this case the PR title should start with `docs: `

tomeichlersmith (commented):
Ok, I believe it is ready to be merged again

src/uproot/writing/writable.py
```diff
@@ -154,14 +166,22 @@ def update(file_path: str | Path | IO, **options):
...
     Options (type; default):
 
-    * initial_directory_bytes (int; 256)
-    * uuid_function (callable; ``uuid.uuid1``)
+    * initial_directory_bytes (int; 256): The number of bytes to allocate for new directories so that
+      TKeys can be added to them without needing immediate rewriting of the block.
+    * initial_streamers_bytes (int; 1024): Ignored by ``update``.
+    * uuid_function (callable; ``uuid.uuid1``): Function to create the file's UUID and/or any directory's UUID.
+    * compression (:doc:`uproot.compression.Compression` or None; ``uproot.ZLIB(1)``):
+      Ignored by ``update``.
```
ariostas: Last thing. `initial_streamers_bytes` and `compression` are not just ignored, they actually end up raising an exception. So let's just remove those two here.

tomeichlersmith: That makes sense, I did not try to trace them through the code - thank you for the correction, they have been removed.

ariostas (approved):
Thank you, @tomeichlersmith!

ariostas:
@all-contributors please add @tomeichlersmith for docs

PR #1704, opened 2026-08-14
ci: automate release-note categorization with type labels

TaiSakuma:
## What this does

Makes "Generate release notes" produce the categorized format directly, eliminating the manual reorganization the release manager does today (e.g. for v5.7.5):

1. **`.github/workflows/pr-type-label.yml`** (new) — labels every PR `type/<type>` from its conventional-commit title, parsed by the same SHA-pinned `amannn/action-semantic-pull-request` already used by the title check. One label per commit type; no release-note semantics in the labels.
2. **`.github/release.yml`** (new) — groups those labels into the sections uproot's notes already use: `type/feat` → **New features**; `type/fix`, `type/perf` → **Bug-fixes and performance**; everything else (incl. unlabeled) → **Other** via the `*` catch-all.

This is a port of scikit-hep/awkward#4125, which has been running in awkward since 2026-06-12.

## Design notes for review

- **`pull_request_target`**: required so labeling works on fork PRs (plain `pull_request` gets a read-only token). Mitigations: the workflow never checks out or executes PR code; the parsed type reaches the script only via `env:`; top-level `permissions: {}` with job-level `pull-requests: write` only. The zizmor ignore is justified inline, and `pre-commit run` — which runs zizmor with `--persona=pedantic` over `^\.github` — passes.
- The existing `semantic-pr-title.yml` required check is untouched, and its action pin (`48f2562…`, v6.1.1) is reused rather than bumped.
- `synchronize` is deliberately not a trigger: pushes can't change a title, so it could never change the label.
- **`cancel-in-progress: false`**, which differs from awkward. Workflow-level concurrency cancels a run before the job-level `if` is evaluated, so a body-only edit shortly after a title edit cancels the labeling run and then skips itself. That happened on awkward#4301, which still carries no `type/fix` label. Queueing is safe here because the reconcile step reads live labels rather than the event payload, and is idempotent.
- Reconciliation only ever touches `type/*` labels — existing labels (`bug`, `feature`, `next-release`, …) are never modified.
- Bot PRs are covered: in awkward, Dependabot and pre-commit.ci PRs receive `type/chore` alongside their existing labels (awkward#4302, awkward#4295, awkward#4285).
- **Heading shape changes slightly**: GitHub's generator emits `## What's Changed` with `###` category subsections, vs. the bare `##` sections of past hand-written notes.
- This workflow does not run on its own PR: `pull_request_target` loads the workflow file from the base branch, so nothing gets labeled until this merges.

## Completing the rollout

Two steps finish this change but can't live in a diff — one is repository metadata, the other a one-time pass over existing PRs:

- **Create the 11 labels** (repo metadata) — *once this is approved, before it merges*, so the workflow never runs against labels that don't exist: `type/feat` `type/fix` `type/perf` `type/docs` `type/test` `type/build` `type/ci` `type/chore` `type/refactor` `type/style` `type/revert`. Planned with the same colors and descriptions as awkward's, so the two repos match. Naming feedback welcome before they're created.
- **One-time backfill** — *after merge* — labeling the 22 PRs merged since v5.7.5 and the 17 currently open, so the next release's notes are complete from day one.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,18 @@
+# Configuration for GitHub's automatically generated release notes
+# ("Generate release notes" button / the generate-notes REST API).
+# https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes
+#
+# The type/* labels are applied automatically from the conventional-commit
+# type in each PR title by .github/workflows/pr-type-label.yml.
+changelog:
+  categories:
+    - title: New features
+      labels:
+        - type/feat
+    - title: Bug-fixes and performance
+      labels:
+        - type/fix
+        - type/perf
+    - title: Other
+      labels:
+        - '*'
--- a/.github/workflows/pr-type-label.yml
+++ b/.github/workflows/pr-type-label.yml
@@ -1,9 +1,10 @@
 name: PR Type Label
 
 # SECURITY: pull_request_target grants a write token, so this workflow must
-# never check out or execute code from the PR. It only parses the PR title
-# (via the same pinned action used by semantic-pr-title.yml) and calls the
-# GitHub REST API to reconcile labels.
+# never check out or execute code from the PR. It runs no third-party
+# actions: the title is parsed by sed against a fixed allowlist, reaches the
+# script only through an environment variable, and labels are reconciled via
+# the GitHub REST API.
 on: # zizmor: ignore[dangerous-triggers] This privileged workflow never checks out or runs PR code; it only parses the PR title and edits labels via the REST API.
   pull_request_target:
     types:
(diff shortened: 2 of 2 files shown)


ianna (approved):
@TaiSakuma - Fantastic! Thank you.

ariostas (approved):
Looks good, thank you!

ariostas:
Actually, @TaiSakuma. Claude suggested two tweaks, which make sense to me.

## Two things worth changing

**1. An unknown type creates a stray label.** `validatePrTitle.js` calls `core.setOutput('type', result.type)` *before* it validates the type against the allowed list, and step outputs survive a failed step. So a title like `foo: bar` yields `TYPE=foo`, and the `POST .../labels` call then either creates a stray `type/foo` label in the repo or errors out. The required title check blocks such a PR from merging, but the label damage happens at open/edit time. Gating on the parse outcome fixes it:

```yaml
TYPE: ${{ steps.parse.outcome == 'success' && steps.parse.outputs.type || '' }}
```

Empty type → the reconcile step just strips stale labels, which is the desired behavior for an invalid title.

**2. Consider dropping the third-party action from the privileged job.** It's the one real blast-radius increase in this PR: `amannn/action-semantic-pull-request` currently runs with `pull-requests: read` in an unprivileged context, and this workflow hands the same action a write token in a privileged one. It's SHA-pinned, so this isn't urgent, but the parsing you need is one line:

```yaml
env:
  PR_TITLE: ${{ github.event.pull_request.title }}
run: |
  type=$(printf '%s' "$PR_TITLE" | sed -nE 's/^([a-z]+)(\([^)]*\))?!?: .+/\1/p')
  case "$type" in
    feat|fix|perf|docs|test|build|ci|chore|refactor|style|revert) ;;
    *) type='' ;;
  esac
```

That removes third-party code from the write-token job and solves finding #1 via the allowlist. The cost is losing "same parser as the required check," but the required check still enforces the canonical rule — this job only needs the type prefix. Your call on whether the consistency argument outweighs it.

TaiSakuma:
@ariostas, thanks. I pushed a commit with the two suggested changes.

ariostas:
Thank you @TaiSakuma! The failures are unrelated, so I'll merge it

Issue #1673, opened 2026-07-02
Implement expression support for RNTuples


ariostas:
The `arrays` method for TTrees supports expressions, so you can do things like `tree.arrays(["x+y", "x**2"])` and it will pre-compute those operations. We should implement the same functionality for RNTuples.

However, the expression support for TTrees is handled fully internally (with code like [this](https://github.com/scikit-hep/uproot5/blob/f5c0cded2f6ca46b791d2ff3e6549f92f9c4ceb0/src/uproot/language/python.py) and [this](https://github.com/scikit-hep/uproot5/blob/f5c0cded2f6ca46b791d2ff3e6549f92f9c4ceb0/src/uproot/behaviors/TBranch.py#L3257)). We want to fully delegate expression parsing to [Formulate](https://github.com/scikit-hep/formulate), which is specifically designed for these kind of things. Formulate is still not very widely used, so mostly likely this will bring up limitations, and we will have to adapt on that side as well.

After the RNTuple implementation is done, the plan is to refactor the expression support for TTrees to use Formulate as well.

ariostas:
@KaranSinghDev this could be a good project for you, if you're interested.

KaranSinghDev:
@ariostas Thanks, I am currently doing some testing to understand it . I will let you know .

KaranSinghDev:
Hi @ariostas,
I have spent some days on it and here is my understanding.

Looking at how functions like `arrays` interact, we need a approach that covers especially the `cut, aliases, expressions` toghether while bypassing the legacy `uproot.language.python` engine for formulate.

My thinking is that we can use `formulate.from_root()` to generate AST for the expressions, and extract `ast.variables` to ensure we only pass the minimally required `RFields`  to awkward reading layer. Once the raw array is materialized we convert AST and computer them using `ak.numexpr.evaluatet()` , then evaluate the cut AST into boolean mask and apply it to final arrays .

The approach seems safe and wold not interece GPU or virutal layers as math  is post materlization. But there will be memeory footprint for the larger chunks while iteration.

Let me know your thoughts on this .

ariostas:
@KaranSinghDev yeah, that's exactly what I was thinking. And since initially this would be only for RNTuple, we don't have to worry about the legacy `uproot.language.python`.

> But there will be memeory footprint for the larger chunks while iteration.

[...]

PR #1705, opened 2026-08-17
feat: implement expressions, cut, and aliases for RNTuple.arrays via formulate 

KaranSinghDev:
**Description**
This PR implements initial support for `expressions`, `cut`, and `aliases` in `RNTuple.arrays()` by delegating parsing to `formulate` and evaluation to `ak.numexpr`, as discussed in #1673.

**Key Changes**
1.formulate.from_root() parses expressions and extracts ast.variables to ensure minimal I/O (we only fetch required fields).
2.Evaluation happens post-materialization via ak.numexpr.evaluate().
3.cut expressions generate a boolean mask applied to the resulting Awkward array.
4.Added regression test in tests/test_1673_rntuple_expressions.py.

@ariostas Let me know your thoughts on this.


src/uproot/behaviors/RNTuple.py
```diff
@@ -733,14 +733,64 @@ def arrays(
...
+        parsed_expressions = {}
+        parsed_cut = None
+
+        if expressions is not None or cut is not None:
+            try:
+                import formulate
+
+                has_formulate = True
+            except ModuleNotFoundError:
+                has_formulate = False
```
ariostas: Let's add formulate as a strict dependency. We'll also have to change some dependencies for compression, so we'll release a minor version soon.

KaranSinghDev: @ariostas have added.

KaranSinghDev:
@ariostas I was thinking if we should put the virtual arr args first on function, the invalid ones would fail before loading dependencies and parsing

ariostas:
> I was thinking if we should put the virtual arr args first on function, the invalid ones would fail before loading dependencies and parsing

Yeah, the virtual check should come checks. Expressions won't be supported for virtual arrays (there's already a check there), so it should fail right away.

ariostas:
Also, I see that the tests are failing because numexpr is not installed. I forgot that numexpr is not a dependency of formulate. So we should also add it as a dependency as well.

What you have so far looks good to me. I think the one thing it might be worth thinking a bit about is if there should be some simple check that decides whether the cut should be applied before of after evaluation. If the cut throws away most of the data then it's worth applying it before, but if it let's most things through, then it's better to apply it at the end.

ariostas:
Hmm I guess Pyodide might be a problem. I'll see what we can do about that

KaranSinghDev:
Hi @ariostas I tried to lok in it, since numexpr does not have a WebAssembly wheel, so making it a strict dependency seems to be breaking the Uproot installation for browser users. I think this can be fixed by using an environment marker in pyproject.toml ,`"numexpr; sys_platform != 'emscripten'"`

Also working on the codecov coverage. Let me know your thoughts on this.

ariostas:
Yeah, I think for now we can just add the marker to the pyproject.toml and skip the tests in Pyodide. I'm gonna see how feasible it is to have a numexpr wheel for pyodide.

As for codecov. Don't worry too much about it. If the tests ends up needing to be really convoluted just to hit coverage, then they might not be worth it. We might need to reconsider that threshold since we haven't been hitting it for most PRs.

ariostas:
I opened https://github.com/pydata/numexpr/pull/568. We'll see what they say.

KaranSinghDev:
Hi @ariostas so far pydata has not responded to our issue, I was thinking to continue with the marker for now.

ariostas:
> Hi @ariostas so far pydata has not responded to our issue, I was thinking to continue with the marker for now.

Yeah, that sounds good. Seems like they can be a little slow to review, but hopefully it won't take too long.

PR #1718, opened 2026-09-10
fix: PyArrow instead of deprecated awkward-pandas

### LINKED ISSUE #1706: awkward_pandas is outdated and AwkwardExtensionArray breaks HDF5 export
## Summary

`AwkwardExtensionArray` breaks Pandas' HDF5 export as shown below, resulting in errors of type `TypeError: objects of type ``AwkwardExtensionArray`` are not supported in this context, sorry; supported objects are: NumPy array, record or scalar; homogeneous list or tuple, integer, float, complex or bytes` .

That `AwkwardExtensionArray` comes from `awkward_pandas`, the pypi package of which was not updated since 3 years ( https://pypi.org/project/awkward-pandas/ ) and the GitHub page of which ( https://github.com/intake/awkward-pandas ) now redirects to https://github.com/intake/akimbo , which no longer provides `AwkwardExtensionArray`.

I did not study in detail why AwkwardExtensionArray was used in the first place and why the library is discontinued, but I believe it should no longer be used as it introduces this issue.

---

## Reproducer

Step 1: Generate a test ROOT file using the ROOT bindings:

```
import ROOT

f = ROOT.TFile("test.root", "RECREATE")
t = ROOT.TTree("ResultTree", "")

s = ROOT.std.string()
t.Branch("run_name", s)

s.assign("test123")
t.Fill()

t.Write()
f.Close()
```

Step 2: Try to export it as hdf5 using uproot:

```python
import uproot
print("Uproot:", uproot.__version__)

import pandas as pd
print("Pandas:", pd.__version__)

filePath="test.root"
file = uproot.open(filePath)["ResultTree"]

branches = list(file.keys())
dataframe = file.arrays(branches, library="pd")

h5cache = '.result-cache.h5'
cache   = pd.HDFStore(h5cache)
dataframe.to_hdf(h5cache, key=filePath.replace('/','_'))
```

It fails with this output:

```
Uproot: 5.7.5
Pandas: 3.0.5
/home/<user>/test/venv/lib/python3.14/site-packages/tables/path.py:146: NaturalNameWarning: object name is not a valid Python identifier: 'test.root'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
  check_attribute_name(name)
  File "/home/<user>/test/Test.py", line 17, in <module>
    dataframe.to_hdf(h5cache, key=filePath.replace('/','_'))
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/<user>/test/venv/lib/python3.14/site-packages/pandas/core/generic.py", line 2776, in to_hdf
    pytables.to_hdf(
    ~~~~~~~~~~~~~~~^
        path_or_buf,
        ^^^^^^^^^^^^
    ...<13 lines>...
        encoding=encoding,
        ^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/<user>/test/venv/lib/python3.14/site-packages/pandas/io/pytables.py", line 316, in to_hdf
    f(store)
    ~^^^^^^^
  File "/home/<user>/test/venv/lib/python3.14/site-packages/pandas/io/pytables.py", line 296, in <lambda>
    f = lambda store: store.put(
                      ~~~~~~~~~^
        key,
        ^^^^
    ...<8 lines>...
        dropna=dropna,
        ^^^^^^^^^^^^^^
    )
    ^
  File "/home/<user>/test/venv/lib/python3.14/site-packages/pandas/io/pytables.py", line 1218, in put
    self._write_to_group(
    ~~~~~~~~~~~~~~~~~~~~^
        key,
        ^^^^
    ...<12 lines>...
        dropna=dropna,
        ^^^^^^^^^^^^^^
    )
    ^
  File "/home/<user>/test/venv/lib/python3.14/site-packages/pandas/io/pytables.py", line 1957, in _write_to_group
    s.write(
    ~~~~~~~^
        obj=value,
        ^^^^^^^^^^
    ...<11 lines>...
        track_times=track_times,
        ^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/<user>/test/venv/lib/python3.14/site-packages/pandas/io/pytables.py", line 3501, in write
    self.write_array(f"block{i}_values", blk.values, items=blk_items)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/<user>/test/venv/lib/python3.14/site-packages/pandas/io/pytables.py", line 3356, in write_array
    self._handle.create_array(self.group, key, value)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/<user>/test/venv/lib/python3.14/site-packages/tables/file.py", line 1214, in create_array
    flavor = flavor_of(obj)
  File "/home/<user>/test/venv/lib/python3.14/site-packages/tables/flavor.py", line 200, in flavor_of
    raise TypeError(
    ...<2 lines>...
    )
TypeError: objects of type ``AwkwardExtensionArray`` are not supported in this context, sorry; supported objects are: NumPy array, record or scalar; homogeneous list or tuple, integer, float, complex or bytes
/home/<user>/test/venv/lib/python3.14/site-packages/tables/file.py:130: UnclosedFileWarning: Closing remaining open file: .result-cache.h5
  warnings.warn(UnclosedFileWarning(msg))
```

2xB:
awkward-pandas is deprecated. This PR solves that by using PyArrow instead as described in https://awkward-array.org/doc/main/user-guide/how-to-convert-pandas.html . This does change the output data type slightly, see tests affected by this commit.

Some arrays in the test are no longer plain nested Python arrays but Python arrays of NumPy arrays, using `awkward.array_equal` to compare them across types.

I tried to reflect the changed behavior in the documentation. This does change the dependency on awkward-pandas to a dependency on PyArrow, which I however believe to be justified since the above linked Pandas guide itself suggests doing so.

Fixes #1706

[edit] Side note: I thought about whether to try and replicate the old behavior of awkward-pandas, but I believe it should be better to use the documented approaches. It is worth noting that having NumPy arrays instead of plain Python arrays or no longer having the `.ak` accessor but instead a different code to get Awkward arrays does affect the usage of this library, so this commit could also be tagged `BREAKING CHANGE:`. Btw. Pandas nowadays also has a native ArrowDtype https://pandas.pydata.org/docs/reference/api/pandas.ArrowDtype.html, but that is still written as experimental. But going from the Awkward dtype of awkward-pandas to the ArrowDtype provided by Pandas would rely on experimental features and I do not know if one would gain much.


2xB:
I just noticed I should check if updating from awkward-pandas to https://github.com/intake/akimbo is possible, since that maintains more features of the current implementation.

2xB:
Note: The predecessor of awkward-pandas is akimbo (https://github.com/intake/awkward-pandas even forwards to https://github.com/intake/akimbo). That is a library with the feature that as soon as you write `import akimbo.pandas`, all Pandas arrays automatically can be accessed with `.ak` like those gotten by awkward-pandas before. One could document that as well or discuss if akimbo.pandas should be imported automatically so uproots Pandas arrays automatically behave more like they did before. I would guess the second option might be more natural, but it also adds another dependency.

2xB:
Just to remark this: The codecov error stems from the fact that the tests do not test what happens when a module is not installed. However I believe this to be usual for these tests.

ariostas:
Hi @2xB, thank you for this PR and sorry for the delay!

I'm a bit concerned with some of the changes since dtype info is completely lost, and also since there are intermediate steps as Python lists, which could have really bad performance when working with large datasets.

From what I understand, `akimbo` is the spiritual successor, instead of the predecessor. So I think it would be worth exploring swapping the `awkward-pandas` dependency for `akimbo`. I did some quick exploration with Claude, and it seems like there would still be some breaking changes, but it is much closer to the way `awkward-pandas` worked, so I think it makes more sense to go that route.

2xB:
Hej @ariostas, oh, I used the wrong word. `successor` is of course right. It still shares the same git history with the old project though (tag `2023.8.0` is the latest `awkward-pandas` commit).

Anyways, https://github.com/intake/akimbo/pull/52 removed `AwkwardExtensionArray` and internally replaced its use with its new `PandasAwkwardAccessor.to_output`, which either uses `pd.arrays.ArrowExtensionArray(ak.to_arrow(data, extensionarray=False))` or `pd.Series(pd.arrays.ArrowExtensionArray(ak.to_arrow(data, extensionarray=False)), index=...)`.

We can leave out the `pd.Series` wrapper they use since uproot already does this wrapping in https://github.com/2xB/uproot5/blob/4ac028a275f66de741f35bf4ab7173ef892167ae/src/uproot/interpretation/library.py#L880 .

I just added the explicit `pd.arrays.ArrowExtensionArray` wrapper to this Pull Request as well so this now behaves more like it did before - I honestly just expected `pd.Series` to automatically wrap the PyArrow array into an `ArrowExtensionArray`. Now we again have reasonable dtypes.

Also, I added a dependency on `akimbo` and imported `akimbo.pandas`, which automatically decorates dataframes with the `.ak` accessor that was available through `awkward-pandas` earlier. One could argue that this dependency is not really needed for uproot itself, one can also argue that this recovers the old behavior. I find it a bit weird since this decorator is automatically added to *all* Pandas dataframes, also ones that users of uproot may create themselves. This kind of spooky action at a distance is something I would rather try to avoid, but here we are.

2xB:
Note: The HDF5 export issue from #1706 is no longer fixed since using `ArrowExtensionArray` again, but this is a missing Pandas feature now and no longer a problem for Uproot to solve. So this will close #1706 although the complete solution will require follow-up changes to Pandas. It is an important step though since it removes an outdated dependency and afterwards the issue is just purely one of Pandas.

2xB:
Ah, but before https://github.com/scikit-hep/uproot5/pull/734 there were no extension arrays and HDF5 export was not an issue. Back in the days, no `ExtensionArray` was used so the HDF export worked as expected. If instead of using `ArrowExtensionArray`, one would go back to the old implementation, there would not be any issue with the HDF export. However I have no idea if that is desirable in any way.

ariostas (commented):
Thank you, @2xB! I like this a lot better than the previous version. Unfortunately, there would still be some behavior changes, but I think this is probably the best option that we have since `awkward_pandas` is deprecated. I think it should be fine since I don't think as many people use the pandas side, but I want to discuss this a bit with other people to see what they think.

I've attached some comments from Claude below.

Thanks for the update! Moving to `ArrowExtensionArray` + akimbo is a big improvement over the earlier object-dtype version: columns stay columnar, and `.ak` works again. I ran the pandas-related tests locally on this head (108 passed, 1 skipped for needing ROOT; network/xrootd/distributed not run) and left inline comments. The points that don't belong on a single line:

**This is a breaking change.** The title is `fix:` and the akimbo commit is `feat:` without `!`, so release tooling would treat it as a patch/minor bump. For users:
- the column dtype changes from `awkward` to an Arrow dtype;
- cells (`s[i]`) are now Python lists instead of `ak.Array` (the updated `test_0912` asserts this);
- arithmetic directly on the Series (`s * 2`) no longer works (`s.ak * 2` does);
- `uproot.extras.awkward_pandas()` is removed.

Could you retitle to `feat!:` (or add a `BREAKING CHANGE:` footer) and include a short migration note?

**PR description is out of date.** It still mentions "Python arrays of NumPy arrays" and `awkward.array_equal`, which no longer match the code. It also says this no longer fixes #1706, so that issue should stay open.

src/uproot/interpretation/library.py
```diff
@@ -870,6 +874,9 @@ def imported(self):
 
     def finalize(self, array, branch, interpretation, entry_start, entry_stop, options):
         pandas = self.imported
+        uproot.extras.akimbo_pandas()  # Automatically adds .ak accessors to Pandas DataFrames
```
ariostas: :robot: _AI text below_ :robot:

This makes akimbo a hard requirement for **every** `library="pd"` call, including flat numeric data. With akimbo not installed:

```python
events.arrays(["NMuon", "MET_px"], library="pd")
# ModuleNotFoundError: install the 'akimbo' package with: ...
```

On `main`, flat branches need only pandas. Could this import be optional (try to import `akimbo.pandas`, and continue without it if it's missing), or moved into the jagged branch of `_process_array_for_pandas`? It also runs once for every branch that gets finalized.

Making it optional also helps with two other points:
- akimbo's last PyPI release is 2025.3.0 (conda-forge only has 2024.12.0), so it's better not to make it a hard dependency.
- awkward-pandas and akimbo both register `.ak`, and whichever is imported last wins. If a user still has awkward-pandas installed and imports it after uproot, `s.ak` fails with "ak accessor called on incompatible data". It's worth telling users in the docs to uninstall awkward-pandas.

src/uproot/interpretation/library.py
```diff
@@ -832,7 +832,11 @@ def _process_array_for_pandas(
                 ) and array.layout.minmax_depth == (1, 1):
                     array = array.to_numpy()
                 else:
-                    array = uproot.extras.awkward_pandas().AwkwardExtensionArray(array)
+                    pandas = uproot.extras.pandas()
+                    uproot.extras.pyarrow()  # The next line requires PyArrow
+                    array = pandas.arrays.ArrowExtensionArray(
+                        awkward.to_arrow(array, extensionarray=False)
```
ariostas: :robot: _AI text below_ :robot:

With `extensionarray=False`, Awkward-specific types are lost when converting back with `s.ak.array`:

- `uproot-HZZ-objects.root:events/muonp4`: `var * TLorentzVector[fP: TVector3[...], fE: float64]` becomes `var * {fP: {fX, fY, fZ}, fE}`. The record names, and the behaviors attached to them, are gone.
- `uproot-stl_containers.root:tree/map_int32_int16`: `var * tuple[[int32, int16], parameters={"__array__": "sorted_map"}]` becomes `var * {"0": int32, "1": int16}`, and cells print as `[{'0': 1, '1': 1}, ...]`.

With `extensionarray=True`, both convert back to exactly the original type, and so do float32 and nullable ints. The trade-off is that pandas' own `.list` accessor doesn't recognize the extension type, but akimbo's `.ak` covers that. I'd suggest `extensionarray=True`, since ROOT data is full of named records and maps.

src/uproot/interpretation/library.py
```diff
@@ -832,7 +832,11 @@ def _process_array_for_pandas(
                 ) and array.layout.minmax_depth == (1, 1):
                     array = array.to_numpy()
                 else:
-                    array = uproot.extras.awkward_pandas().AwkwardExtensionArray(array)
+                    pandas = uproot.extras.pandas()
+                    uproot.extras.pyarrow()  # The next line requires PyArrow
```
ariostas: :robot: _AI text below_ :robot:

Minor: `awkward.to_arrow` already raises an ImportError with install instructions when pyarrow is missing, so this bare call (whose result is thrown away) can probably be removed.

src/uproot/extras.py
```diff
@@ -284,19 +284,34 @@ def graphed_executors():
         return graphed_executors
 
 
-def awkward_pandas():
+def pyarrow():
```
ariostas: :robot: _AI text below_ :robot:

`uproot.extras.awkward_pandas()` is public and is removed here. Downstream code that calls it would get `AttributeError`. Could it stay for a release, emitting a `DeprecationWarning` that points to pyarrow/akimbo?

docs-sphinx/basic.rst
```diff
@@ -657,7 +657,7 @@ These "nested" NumPy arrays are not slicable as multidimensional arrays because
     # File "<stdin>", line 1, in <module>
     # IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
 
-The Pandas form for this type of data is a DataFrame with Awkward Dtype, provided by the `awkward-pandas <https://github.com/intake/awkward-pandas>`__ package.
+The Pandas form for this type of data is a DataFrame with PyArrow dtype, provided by the `Pandas ArrowExtensionArray <https://pandas.pydata.org/docs/reference/api/pandas.arrays.ArrowExtensionArray.html>`__ type.
```
ariostas: :robot: _AI text below_ :robot:

The example output right below is now stale. With this PR, cells print as NumPy-style float32 values, e.g. `[-52.899456  37.73778 ]`, not full-precision Python lists like `[-8.16079330444336, -11.307581901550293]`. It would be good to regenerate it.

docs-sphinx/basic.rst
```diff
@@ -677,7 +677,7 @@ The Pandas form for this type of data is a DataFrame with Awkward Dtype, provide
 
     [2421 rows x 6 columns]
 
-You can operate on Awkward Array data in Pandas using the ``.ak`` accessor; see the [awkward-pandas documentation](https://awkward-pandas.readthedocs.io/en/latest/quickstart.html).
+You can operate on Awkward Array data in Pandas using the ``.ak`` accessor; see the [akimbo documentation](https://akimbo.readthedocs.io/en/latest/quickstart.html).
```
ariostas: :robot: _AI text below_ :robot:

It would help to add a short migration note here, because cells are now Python lists rather than `ak.Array`. Arithmetic on a raw cell now fails **silently** instead of raising:

```python
s[0] * 2      # before: <Array [3, 5]>   now: [1.5, 2.5, 1.5, 2.5]
s[0] + s[0]   # before: <Array [3, 5]>   now: [1.5, 2.5, 1.5, 2.5]
```

Suggested wording: to get an Awkward array for a row, use `s.ak.array[i]`. In loops, convert the column once (`arr = s.ak.array`, then `arr[i]`), because each `s.ak.array` call converts the whole column (about 0.3 ms on 1M rows).

tests/test_0018_array_fetching_interface.py
```diff
@@ -744,8 +744,10 @@ def test_jagged_awkward():
 
 
 def test_jagged_pandas():
+    awkward = pytest.importorskip("awkward")
```
ariostas: :robot: _AI text below_ :robot:

Nit: `awkward` is imported here but never used in this test.

src/uproot/interpretation/library.py
2xB: As said in the comment above, I don't really know if I want akimbo a dependency at all. Yes, it brings back the `.ak` accessor, but it does so by altering *every* Pandas dataframe that ever exists in that session. I would definitely not make it optional and automatically import if it is available since that means code using the `.ak` accessor would fail with confusing error messages that point to everything but not that the users should just install this random Python module.

src/uproot/extras.py
2xB: If downstream code depended on the uproot importer, that would be a very, very confusing design decision from my point of view. But even if it did because of Hyrum's law ( https://xkcd.com/1172/ ), I believe it is trivial to find this PR. I would prefer to not introduce stale code covering such a weird use case.

tests/test_0018_array_fetching_interface.py
2xB: Sure. The test only works if `awkward` is available due to the nature of this Pull Request, but I can skip assigning it to a value.

src/uproot/interpretation/library.py
2xB: Quoting from https://awkward-array.org/doc/main/user-guide/how-to-convert-pandas.html#conversion-through-apache-arrow :

> The extensionarray flag is normally True, and enables Awkward to preserve metadata through Arrow transformations. However, tools like Arrow’s Pandas conversion do not recognise Awkward’s special extension type, so we must take care to provide Arrow with native types: [...]

So either the documentation is wrong or outdated or this is not supported to my understanding.

2xB: The documentation being outdated can btw. very well be, it does not use `ArrowExtensionArray`s at all. Or one is not supposed to use them in this conversion...

docs-sphinx/basic.rst
2xB: I think if behavior changed, this makes much sense. If you have a good wording, feel free to add this here, ideally mentioning Uproot version numbers - that is great. Thanks!

docs-sphinx/basic.rst
2xB: Ah so there is a precision loss. That's quite bad, I believe.

2xB:
Before we continue, I have to admit that the Claude comment on updating the quickstart guide is actually quite worrying to me. That looks like we have significant precision loss when using PyArrow. I think in a scientific environment, that is quite a bad regression. Intuitively I have no idea why this is, but if that is an issue occurring due to PyArrow, I would like to bring up again the option to revert #734 instead, I don't think that had such a precision loss.

PR #1720, opened 2026-09-14
feat: add hooks for working with graphed-org/graphed

lgray:
Add hooks to uproot that allow it to be used with the delayed execution engine "graphed", it has stabilized enough to where it makes sense to support the code upstream and ease installation.

**Added 2026-09-19 — `uproot.graphed` takes the same form-mapping seam `uproot.dask` does.** Three parameters, each mirroring its `uproot.dask` namesake so a schema library (coffea's `NanoEventsFactory` is the archetype) can drive either arm from one mapping:

- `form_mapping=` — the `ImplementsFormMapping` object `uproot.dask` takes. The recorded array has the MAPPED form, the mapping info's `behavior` is registered on the backend, and the source declares its read list in `TBranch` names, so projection still fetches only the branches the recorded graph touches.
- `known_base_form=` — use this form instead of opening a file to learn it; no file is opened at record time, and the `TTree` name comes from the object path in `files`.
- `backend=` — a graphed backend instance to record on, in place of the default `graphed.awkward.AwkwardBackend` (the route for a caller's own `Array` subclass); `behavior=` is refused beside it.

Along the way: every partition is opened through the source, so the caller's file options reach both read and write; an empty declared read list is honoured; an RNTuple's base form is built the way a TTree's is; and `uproot.extras.graphed()` registers `uproot` as graphed-internal, so a recorded op's provenance (and a run-time error) points at the analyst's line rather than an Uproot frame. That last call needs `graphed>=0.0.3` (on PyPI since 2026-09-19), so the `test` group floor moves to 0.0.3 and `extras.graphed()` tells a user on an older graphed how to upgrade. Tests: `tests/test_1720_graphed_form_mapping.py`, `…_mapped_projection.py`, `…_mapped_write.py`, `…_mapped_edge_cases.py` over a small NanoAOD-shaped helper in `tests/graphed/mini_nanoaod.py`; `src/uproot/_graphed.py` stays at 100 % line+branch under the suite.

The graphed hooks ride uproot's own CI: `graphed`, `graphed-executors` and `graphed-histogram` (all on PyPI; `graphed` at 0.0.3, the other two at 0.0.2 or later) are entries in the `test` dependency group behind `python_version >= "3.11"` (graphed requires 3.11+), so the `tests/test_1720_graphed_*.py` suite runs on the 3.11+ jobs and is skipped on 3.10. The optional imports go through `uproot.extras.graphed()` / `uproot.extras.graphed_executors()`; the suite's non-test helpers live in the `tests/graphed` package and are imported as `tests.graphed.<name>`. The three packages are opted out of `[tool.uv] exclude-newer` the way awkward already is, because the 7-day window would otherwise hide fresh releases of the three; those keys can go once the window has rolled past.

The pre-commit fixes are committed by hand because pre-commit.ci cannot push to a fork owned by an organization. `pyproject.toml` keeps the layout pre-commit.ci's cached `toml-sort` 0.24 produces; a fresh local hook env pulls toml-sort 0.25.0 (2026-09-12), which reformats inline tables and would fail the same hook on pre-commit.ci. The "GPU test with CUDA 13" failure is the self-hosted runner's `cuda-bindings` pre-release version parse in `tests-cuda/`, which this PR does not touch and which `main` hits too.


lgray:
@ariostas this one is ready to go now.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,430 @@
...
+    This is the ``graphed`` analogue of :doc:`uproot._dask.dask`.
+    """
+    library = uproot.interpretation.library._regularize_library(library)
+    if library.name != "ak":
+        raise NotImplementedError(
+            f"uproot.graphed currently supports only library='ak', not {library.name!r}"
+        )
+
+    import awkward
+    from graphed import Session
```
ariostas: Since this is an optional depencency it should be imported through `extras`.

tests/test_1720_graphed_behaviors.py
ariostas, line 1: The test filenames should be `test_1720_*`

tests/graphed/analysis.py
ariostas, line 1: Maybe we should put the non-test files into a `graphed` submodule?

ariostas (commented):
Hi @lgray, I left a few comments. I'll take a closer look when I have some time.

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:
Done in 93e8aab: `uproot.extras.graphed()` and `uproot.extras.graphed_executors()` follow the `dask_awkward()` pattern, and every graphed import in `_graphed.py` / `_graphed_write.py` goes through them.

tests/test_1720_graphed_behaviors.py
lgray: :robot: _AI text below_ :robot:
Renamed to `test_1720_graphed_*.py` in b3e32e4.

tests/graphed/analysis.py
lgray: :robot: _AI text below_ :robot:
Moved into a `tests/graphed` package (`analysis`, `report`, `vector_backend_ref`) in b3e32e4, imported as `tests.graphed.<name>`; that also retired the `sys.path.insert` hacks since `tests/__init__.py` already puts the repo root on `sys.path`. One consequence: running `python -m pytest` with `tests/` as the cwd would resolve `graphed` to that package and skip the graphed tests; the console-script `pytest`, repo-root runs and CI are unaffected.

lgray:
@ariostas took care of your comments - but it looks like whatever serves the root files in the tests is 403'ing hard. I'll try kicking the tests again tomorrow.

lgray:
@ariostas something's really wrong with the scikit-hep hosting all the CI jobs keep timing out.

ariostas:
> @ariostas something's really wrong with the scikit-hep hosting all the CI jobs keep timing out.

Hmm I'll have a look

ariostas:
@lgray should work now. Thanks for letting me know! I restarted the CI

lgray:
@ariostas cool - all green here now except CUDA-13 as expected. Lemme know if you have any comments :-)

lgray:

Seven commits added (`73b479f..6bdc948`, a fast-forward — nothing already reviewed was rewritten): `uproot.graphed` now takes `form_mapping=`, `known_base_form=` and `backend=`, mirroring `uproot.dask`, so a schema library such as coffea's NanoEvents can drive the graphed arm from the same mapping it gives the dask arm. It was a follow-up branch; folding it in here makes this PR the complete hook set rather than two serialized reviews. The description has a section on it. The `test` group floor for `graphed` moves to 0.0.3 (on PyPI today), which `uproot.extras.graphed()` now needs for `graphed.provenance.register_internal`.

lgray:
@ariostas a gentle ping on this one when you have time.

src/uproot/writing/_graphed_write.py
```diff
@@ -0,0 +1,204 @@
...
+    resolved = partition.resolve(tree.num_entries)
+    if resolved.entry_stop <= resolved.entry_start:
+        return []  # fewer entries than steps: skip, never write an empty part file
+    chunk = source.read_range(tree, columns, resolved.entry_start, resolved.entry_stop)
+    evaluated: Any  # a backend array (awkward/numpy); evaluate_ir is typed list[object]
+    (evaluated,) = graphed.evaluate_ir(compiled, backend, {source_name: chunk})
+    # a record graph yields named fields (the derived columns); a bare (non-record) expression
+    # yields a fieldless array with no branch name to write it under — fall back to the source
+    # columns it reads
+    out_rec = evaluated if evaluated.fields else chunk
```
ariostas: :robot: _AI text below_ :robot:

**Bug — a bare expression writes the wrong data.** `graphed_write(ev.px1 * 2, dest)` succeeds and writes unscaled `px1`: for a fieldless result this falls back to the raw input chunk. Through a `form_mapping` the same call crashes with `AttributeError: 'PlaceholderArray' object has no attribute 'reshape'`, because the chunk holds placeholders for unread branches. Please raise a `TypeError` asking for a record (or write under an explicit default branch name) instead of substituting the inputs. `test_write_of_a_projected_array_writes_only_its_branches` currently pins the wrong behaviour (`g.x + g.y` → branches `x`, `y`).

src/uproot/writing/_graphed_write.py
```diff
@@ -0,0 +1,204 @@
...
+    record = {name: out_rec[name] for name in out_rec.fields}
+    idx = graphed.write.blind_part_index(partition, dict(bases))
+    path = graphed.write.part_path(
+        destination, idx, prefix=prefix or "part", suffix=".root"
+    )
+    with uproot.recreate(
+        path, **_recreate_kwargs(compression, compression_level)
+    ) as out:
+        out[tree_name] = record
```
ariostas: :robot: _AI text below_ :robot:

**Output type differs from `dask_write`.** Dict assignment writes an `RNTuple` here, while `dask_write` calls `mktree` and writes a `TTree` — under a parameter called `tree_name`. For a drop-in replacement I'd default to `mktree` and make RNTuple an explicit opt-in.

src/uproot/writing/_graphed_write.py
```diff
@@ -0,0 +1,204 @@
...
+        "lzma": uproot.LZMA,
+        "lz4": uproot.LZ4,
+        "zstd": uproot.ZSTD,
+    }
+    resolved = (
+        codes[compression](compression_level)
+        if isinstance(compression, str)
+        else compression
+    )
+    return {"compression": resolved}
```
ariostas: :robot: _AI text below_ :robot:

`compression="ZLIB"` (accepted by `dask_write`) and unknown names both raise a bare `KeyError`. `ak_to_root` in `_dask_write.py` already has this mapping with a proper error message — could the two share one helper?

src/uproot/writing/_graphed_write.py
```diff
@@ -0,0 +1,204 @@
...
+
+def _select_executor(executor):
+    graphed_executors = uproot.extras.graphed_executors()
+
+    if isinstance(executor, str):
+        return {
+            "process": graphed_executors.local.ProcessPoolExecutor,
+            "thread": graphed_executors.local.ThreadExecutor,
+        }[executor]
+    return executor  # an executor class passed directly (instantiated by the caller)
```
ariostas: :robot: _AI text below_ :robot:

`executor="threads"` is a bare `KeyError`, and an executor *instance* fails with `'ThreadExecutor' object is not callable`. A `ValueError` listing the accepted strings, and accepting an instance (use it as-is when it has `.run`), would be friendlier.

src/uproot/writing/_graphed_write.py
```diff
@@ -0,0 +1,204 @@
...
+    *,
+    steps_per_file=1,
+    prefix=None,
+    tree_name="tree",
+    compute=True,
+    executor="process",
+    max_workers=None,
+    compression="zlib",
+    compression_level=1,
+):
```
ariostas, line 122: :robot: _AI text below_ :robot:

Missing relative to `dask_write`: `storage_options` / remote destinations (`os.makedirs` makes this local-only, `dask_write` goes through `url_to_fs`), `title`, `field_name`, `counter_name`, `initial_basket_capacity`, `resize_factor`. Fine to defer, but then please list them as not-yet-supported in the docstring.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+    full_paths=False,
+    library="ak",
+    ak_add_doc=False,
+    custom_classes=None,
+    allow_missing=False,
+    behavior=None,
+    form_mapping=None,
+    known_base_form=None,
+    backend=None,
+    **options,
```
ariostas, line 289: :robot: _AI text below_ :robot:

**Ported `uproot.dask` kwargs are silently ignored.** `uproot.graphed(f, steps_per_file=2)`, `step_size="1 MB"` and `open_files=False` are all swallowed by `**options` and do nothing. Please either support them or reject them by name with a pointer to where partitioning lives (`graphed_write(steps_per_file=...)`, `aggregate_plan(...)`). Same for `allow_read_errors_with_report`, `decompression_executor`, `interpretation_executor`.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+):
+    """
+    Args:
```
ariostas, line 292: :robot: _AI text below_ :robot:

Please add an **experimental** note here (and on `graphed_write`): the API tracks a pre-1.0 `graphed` and may change without a deprecation cycle.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+
+    import awkward
+
+    graphed = uproot.extras.graphed()
+
+    real_options = options.copy()
+    real_options.setdefault("num_workers", 1)
+    filter_branch = uproot._util.regularize_filter(filter_branch)
+
+    resolved = uproot._util.regularize_files(files, steps_allowed=False, **options)
```
ariostas: :robot: _AI text below_ :robot:

`steps_allowed=False` rejects the `{"f.root": {"object_path": ..., "steps": [...]}}` form that `uproot.dask` accepts ("'steps' are only allowed in uproot.dask").

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+            file_tree.append((file_path, object_path))
+            continue
+        obj = uproot._util.regularize_object_path(
+            file_path, object_path, custom_classes, allow_missing, real_options
+        )
+        if obj is None:
+            continue
+        # TTrees take filter_branch; RNTuples (HasFields) take filter_field (cf. uproot._dask)
+        filter_kw = "filter_field" if isinstance(obj, HasFields) else "filter_branch"
+        keys = obj.keys(
```
ariostas: :robot: _AI text below_ :robot:

Two `uproot.dask` behaviours are missing from this loop: (a) a `TBranch` as the object path (`"f.root:events/px1"`, dask's `is_self` handling) — here it raises "no TBranches in common"; (b) the `AsGrouped` parent/child key normalisation that `_get_dak_array` does after collecting `common_keys`. Sharing that key-collection code with `_dask.py` would fix both and keep them from drifting again.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+            **{filter_kw: filter_branch},
+        )
+        if common_keys is None:
+            common_keys = list(keys)
+        else:
+            keyset = set(keys)
+            common_keys = [k for k in common_keys if k in keyset]
+        if first_ttree is None:
+            first_ttree = obj
+        file_tree.append((file_path, object_path))
```
ariostas: :robot: _AI text below_ :robot:

**Bug — no object path works at construction, fails at read.** `uproot.graphed("file.root")` succeeds (single TTree, as in dask), but every read/write then dies with `TypeError: argument of type 'NoneType' is not a container or iterable` because `partition.tree` is `None`. Store the resolved `obj.object_path` here instead of the caller's `object_path`.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+
+    def open_tree(self, partition, resources):
+        """This partition's ``TTree``, from a file opened once per worker (via ``resources``) with
+        the options ``uproot.graphed`` was given, so ``decompression_executor`` /
+        ``interpretation_executor`` reach the read the way they do on any other uproot read.
+        Every partition read — this source's and ``uproot.graphed_write``'s — opens through here.
+        """
+        return resources.open_once(
+            partition.uri, lambda uri: uproot.open(uri, **self._options)
+        )[partition.tree]
```
ariostas: :robot: _AI text below_ :robot:

**`custom_classes` and `allow_missing` never reach partition reads.** This bypasses `uproot._util.regularize_object_path` (which `_resolve` uses). With `known_base_form=..., allow_missing=True` and a file lacking the tree, the run dies with `KeyInFileError`; the eager `__call__` path would hit `None.num_entries` in the same situation. Suggest opening through `_resolve` semantics and skipping a partition whose object is missing.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+        )
+
+    def _read_columns(self, columns):
+        """The branches one partition reads. This source declares no read list, so an empty one
+        means the driver's syntactic walk named no branch (a whole-record read), not a request to
+        read nothing: fall back to every selected branch."""
+        return list(columns) if columns else list(self._common_keys)
+
+
+class _MappedGraphedTTreeSource(_GraphedTTreeSource):
```
ariostas: :robot: _AI text below_ :robot:

**Simplification — one read path.** `uproot.dask` always reads through a form mapping (`TrivialFormMapping` by default). Doing the same here removes `_evaluation_columns`, `_task_columns`, both `_read_columns` variants and this subclass split. It also fixes two inconsistencies of the unmapped path: `ak_add_doc` reaches the recorded form but not the data (`read_range` never passes it to `tree.arrays`), and `necessary_columns` reports `frozenset()` for a count-only analysis while `Jet_Px` is actually read (the trivially-mapped source reports `{'Jet_Px'}`).

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+    ``TBranch`` names."""
+
+    def __init__(self, *args, name, expected_form, form_mapping_info, interp_options):
+        super().__init__(*args)
+        self._name = name  # the graphed source name, which the projection answers under
+        self._expected_form = expected_form
+        self._form_mapping_info = form_mapping_info
+        self._interp_options = interp_options
+
+    def read_range(self, tree, columns, start, stop):
```
ariostas: :robot: _AI text below_ :robot:

This is nearly a copy of `UprootReadMixin.read_tree` in `_dask.py`, minus the `buffer_replacements` hook — so `FormMappingInfoWithVirtualArrays` isn't honoured here. Could both call one shared helper?

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+    branches the analysis touches; ``None`` reads every selected branch."""
+
+    def __init__(self, file_tree, common_keys, custom_classes, allow_missing, options):
+        self._file_tree = file_tree  # list of (file_path, object_path)
+        self._common_keys = list(common_keys)
+        self._custom_classes = custom_classes
+        self._allow_missing = allow_missing
+        self._options = options
+        self.columns = None  # None -> all common_keys; otherwise the projected subset
+        self.last_columns_read = None  # set on each read, for inspection/tests
```
ariostas: :robot: _AI text below_ :robot:

`columns` is documented as "set by projection", but nothing in uproot or graphed 0.0.3 ever sets it, and `last_columns_read` exists only so tests can assert `__call__` didn't run. Please drop both (a test can spy on `__call__` directly).

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+    ``DATA`` needs read the branch itself. An ``OFFSETS``-only need reads the jagged ``TBranch``'s
+    **counter branch** when the file provides one (``TBranch.count_branch``) — the list lengths
+    without the payload baskets; where no counter exists (or for ``RNTuple``, whose index column is
+    not independently addressable through the public API) it falls back to the branch itself.
+    Returns ``{branch_to_read: requested_path}``."""
+    graphed = uproot.extras.graphed()
+
+    out = {}
+    for path, need in needs.items():
+        if need is graphed.BufferNeed.DATA or str(need) == "data":
```
ariostas: :robot: _AI text below_ :robot:

`str(need) == "data"` — the enum identity check should be enough given the `graphed>=0.0.3` floor.

ariostas (commented):
I did some skimming, but since it's a big PR, my review was mostly AI-driven. I left some comments below. This was with Fable. I'll see if a GPT model has any extra recommendations.

Thanks for this — the projection story and the form-mapping seam are nicely done, and the suite is green and lint-clean locally. Since the goal is for this to eventually stand in for `uproot.dask`, I checked it against `uproot.dask` / `dask_write` behaviour and found a few correctness problems and parity gaps (inline). Three overall requests:

1. **Mark it experimental.** The top-level API depends on a 0.0.x package whose required release predates this PR update by hours. Please say so in the `uproot.graphed` / `uproot.graphed_write` docstrings (API may change without a deprecation cycle), and keep the public surface to those two names (+ maybe `graphed_head`) until it settles.
2. **Collapse the two read paths into one** by defaulting to `TrivialFormMapping`, as `uproot.dask` does. I tried it: 128/135 tests still pass, and the 7 that don't either assert unmapped-only reporting or depend on the bare-expression write fallback, which is a bug.
3. **Trim the tests.** On the ubuntu 3.11 CI job the `test_1720_*` block is 42 s of a 160 s pytest run (26 % of the time for 13 % of the tests); locally it is 23 s wall but 43 s CPU because of process pools, which is what a single-core runner pays. Almost half of that is four files that exercise only `graphed_partitions` / `read_graphed_partition`.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+    Mirrors :doc:`uproot._dask.dask`'s chunking knobs:
+
+    - ``steps_per_file`` (default 1): split each ``TTree`` into that many contiguous chunks.
+    - ``step_size`` (int entries, or a memory string like ``"100 MB"``): cap each chunk's size.
+      Mutually exclusive with ``steps_per_file``, and **incompatible with** ``open_files=False``.
+    - ``open_files`` (default ``True``): open every file to read its entry count and emit exact entry
+      ranges. ``open_files=False`` is the **blind-steps** mode — files are *not* opened here; each
+      chunk records ``(step_index, n_steps)`` instead (encoded as ``entry_start=step_index``,
+      ``entry_stop=-n_steps``) and the real entry range is resolved against the file's own count when
+      the chunk is read (:doc:`uproot._graphed.read_graphed_partition`).
```
ariostas: :robot: _AI text below_ :robot:

Stale docstring: blind chunks are no longer encoded as `entry_start=step_index, entry_stop=-n_steps`; the code uses `Partition.blind`.

src/uproot/_graphed.py
```diff
@@ -0,0 +1,671 @@
...
+    """A partition's concrete ``(entry_start, entry_stop)`` against a tree's entry count."""
+    if getattr(partition, "is_blind", False):
+        partition = partition.resolve(n_entries)
+    start, stop = partition.entry_start, partition.entry_stop
+    if (
+        stop < 0
+    ):  # legacy blind sentinel in older serialized plans: step `start` of `-stop` steps
+        n_steps, step = -stop, start
+        start = (step * n_entries) // n_steps
+        stop = ((step + 1) * n_entries) // n_steps
```
ariostas: :robot: _AI text below_ :robot:

This "legacy blind sentinel in older serialized plans" branch can't be reached for a feature that has never shipped; with the 0.0.3 floor the `getattr(partition, "is_blind", False)` above can also be a plain attribute access.

src/uproot/__init__.py
```diff
@@ -191,6 +192,15 @@
 
 from uproot._util import no_filter
 from uproot._dask import dask, ImplementsFormMapping, ImplementsFormMappingInfo
+from uproot._graphed import (
```
ariostas: :robot: _AI text below_ :robot:

Eight new top-level names (dask added two), three of them with nothing `graphed` in the name (`uproot.necessary_columns`, `uproot.necessary_buffers`, `uproot.resolve_read_branches`). `resolve_read_branches`, `graphed_partitions` and `read_graphed_partition` aren't used by any uproot code path — they support hand-built plans that never go through `uproot.graphed`. While this is experimental I'd export only `graphed` and `graphed_write` (maybe `graphed_head`) and leave the rest importable from `uproot._graphed`.

src/uproot/extras.py
```diff
@@ -239,6 +239,45 @@ def dask_awkward():
...
+        import graphed
+        import graphed.awkward
+        import graphed.awkward.projection
+        import graphed.core
+        import graphed.provenance
+        import graphed.write
+
+        # a recorded op's provenance is the analyst's line, not the Uproot frame that recorded it
+        graphed.provenance.register_internal("uproot")
+    except (ModuleNotFoundError, AttributeError) as err:
```
ariostas: :robot: _AI text below_ :robot:

Catching `AttributeError` to detect an old `graphed` will also turn a genuine bug in these imports into an "install graphed" message. An explicit version check (as done elsewhere in `extras.py`) is clearer.

pyproject.toml
```diff
@@ -21,13 +21,17 @@ extra = [
 test = [
   "aiohttp",
   "deflate",
+  "graphed[awkward,parquet]>=0.0.3; python_version >= '3.11'",
```
ariostas: :robot: _AI text below_ :robot:

Putting these in the main `test` group runs the graphed suite on every 3.11+ job. Given the cost below, would a separate `test-graphed` group with one dedicated CI job (like `test-xrootd` / `test-pyodide`) work for you while this is experimental?

tests/test_1720_graphed_blind_steps.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def _run(tasks):
+    plan = Plan(process=gu.process, combine=gu.combine, empty=gu.empty, tasks=tasks)
+    return ProcessPoolExecutor(max_workers=4).run(plan).value
+
+
+@pytest.mark.parametrize("step_size", ["50 kB", uproot._util.unset])
+@pytest.mark.parametrize("steps_per_file", [1, 2, 5, 13, uproot._util.unset])
+@pytest.mark.parametrize("open_files", [False, True])
```
ariostas: :robot: _AI text below_ :robot:

This is the most expensive file (9 s on CI): 20 parametrisations, the valid ones each spawning a 4-worker process pool, to cover the ~20-line range arithmetic in `graphed_partitions`. Asserting on the returned entry ranges directly (no executor) covers the same uproot code in milliseconds; one executor run is enough as an integration check. More generally, `max_workers=4` process pools on a single-core runner mostly measure worker start-up — `executor="thread"` or a sequential runner would do for everything except one process-pool smoke test.

tests/test_1720_graphed_preserve.py
```diff
@@ -0,0 +1,174 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot5/blob/main/LICENSE
```
ariostas: :robot: _AI text below_ :robot:

In-process, this file, `test_1720_graphed_report.py` and `test_1720_graphed_executor.py` execute only `graphed_partitions`, `read_graphed_partition` and `_partition_range` from uproot; the assertions are about graphed's Bundle / DurablePlan / dead-letter / executor machinery. Those belong in graphed's own test suite — here they tie uproot CI to a 0.0.x package's behaviour. I'd drop the three files (and `tests/graphed/report.py`).

tests/test_1720_graphed_projection_and_compiled_reads.py
```diff
@@ -0,0 +1,171 @@
...
+    with uproot.recreate(src_path) as f:
+        f["events"] = {
+            "x": np.arange(12, dtype="f8"),
+            "y": np.arange(12, dtype="f8") * 2.0,
+            "z": np.arange(12, dtype="f8") * 3.0,
+        }
+    g = uproot.graphed(src_path + ":events", library="ak")
+    outdir = os.path.join(tmp_path, "out")
+    paths = uproot.graphed_write(
+        g.x + g.y, outdir, steps_per_file=2, tree_name="events"
```
ariostas: :robot: _AI text below_ :robot:

This pins the fallback bug noted in `_graphed_write.py`: writing `g.x + g.y` is expected to produce branches `x` and `y` rather than the sum.

tests/graphed_helpers/__init__.py
```diff
@@ -0,0 +1 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot5/blob/main/LICENSE
```
ariostas, line 1: :robot: _AI text below_ :robot:

As you noted, this package shadows the real `graphed` when pytest runs from `tests/`. Renaming it (e.g. `tests/graphed_helpers`) removes the hazard for free. Also, 20 test files is a lot for one feature — after the cuts above, consolidating into a handful (`read`, `form_mapping`, `write`, `projection`) would make this much easier to maintain.

ariostas (commented):

A GPT second pass found these additional, non-duplicate issues:

1. **P1: nested RNTuples get the wrong schema** (`src/uproot/_graphed.py:316-341`). `HasFields.keys(full_paths=False)` flattens `rec.x` to `x`, then `_get_ttree_form` advertises `x` as a top-level field even though `RNTuple.arrays(["x"])` returns it under `rec`. In a local `{rec: {x, y}, z}` RNTuple, `g.x` records but fails at execution, and `graphed_head(g.rec.x, 3)` also fails. Please use `to_akform(filter_name=common_keys)` as `_dask.py:1767-1769` does.

2. **P1: repeated inputs overwrite output parts** (`src/uproot/writing/_graphed_write.py:181-199`). `file_bases` is a dict keyed only by `(uri, tree)`, so `graphed([spec, spec])` with two steps reports `part-00002.root`, `part-00003.root` twice and writes only one copy of the rows. The occurrence needs to be part of the partition identity/index, or duplicate inputs need to be rejected.

3. **P1: writes silently change axis-0 semantics with partition count** (`src/uproot/writing/_graphed_write.py:80-82`). The complete expression is replayed independently on each chunk. For `g.x[2:8]` over `range(10)`, one step writes `[2,3,4,5,6,7]`, two steps write `[2,3,4,7,8,9]`, and three write `[2,5,8,9]`; `dask_write` preserves dataset-global slicing. Either global operations need partition-aware handling or this path must reject/document non-row-local graphs.

4. **P1: `graphed_head` applies `n` before the expression** (`src/uproot/_graphed.py:665-670`). It reads the first `n` source rows and only then filters/slices. Thus `graphed_head(g.x[10:20], 5)` is empty instead of `[10,11,12,13,14]`, and `head(g[g.x % 2 == 0], 2)` returns one row instead of two. A dask-like head must collect the first `n` output rows.

5. **P1: mapped projection can under-supply compiled evaluation** (`src/uproot/_graphed.py:632-638`). The comment above `_evaluation_columns` correctly notes that compiled IR replays dead intermediate nodes, but mapped sources switch back to output-buffer projection. With `zip({"jet": g.Jet.pt, "muon": g.Muon.pt})[["jet"]]`, projection omits `Muon_pt` while IR still evaluates that node; both `graphed_head` and `graphed_write` fail. Mapped reads need the same syntactic-evaluation safeguard translated through the mapping.

6. **P2: the default process writer cannot ship common behaviors** (`src/uproot/writing/_graphed_write.py:185-195`). The partial serializes `session.backend`; `vector.backends.awkward.behavior` contains lambdas, so even the returned `compute=False` plan raises `PicklingError` under `pickle.dumps`. The generic Parquet writer already solves this with an importable behavior/backend reference; ROOT writing needs the same seam or should avoid serializing the backend instance.

7. **P2: external evaluators are dropped by `graphed_head` and `graphed_write`** (`src/uproot/_graphed.py:663-670`, `src/uproot/writing/_graphed_write.py:82`). A `graphed.apply(...)` expression materializes and runs through `aggregate_plan`, but these helpers call `evaluate_ir` without the session evaluators and fail with “External payload ... needs an evaluator.” `aggregate_plan` already wires these through `_external_evaluators`.

8. **P2: already-open TTrees are not partition-compatible** (`src/uproot/_graphed.py:82-100`). They are accepted by the documented dask-compatible `files` API and whole-dataset materialization works, but the `Model_TTree` is stored as `Partition.uri`; `resources.open_once` then raises because it is unhashable. Normalize it to a reusable file/tree reference or special-case open objects in `open_tree`.

The existing Graphed-focused suite still passes (`135 passed`); these are missing edge/parity cases rather than current test failures.

lgray:
@ariostas thanks for spending the tokens - I will do so similarly once my limits reset :-)

lgray:

Pushed as four commits (dd14af0, 8565546, 5aac34e, bbb884b), one per class of finding. Three of the round-3 items are defects in graphed's own partition-wise drivers, so they are fixed there: graphed-org/graphed#38, and this PR's floor moves to `graphed>=0.0.4`. Until 0.0.4 is on PyPI the new `graphed-build` CI job here will fail to resolve the dependency; everything else is unchanged for non-graphed jobs.

Item-by-item (the test named is the one that fails without the change):

**Round 2 — overall**

| # | Request | Result |
|---|---|---|
| 1 | Mark experimental, shrink the surface | Done. Notes in the module and both function docstrings. Exported names: `uproot.graphed`, `uproot.graphed_head`, `uproot.graphed_write` (the last via `uproot.writing`, like `dask_write`). `necessary_columns`, `graphed_partitions`, `read_graphed_partition` stay importable from `uproot._graphed`. `necessary_buffers` and `resolve_read_branches` are deleted. |
| 2 | One read path via `TrivialFormMapping` | Done. Every read goes through the mapping and a module-level `_read_tree` shared with `uproot.dask` (`buffer_replacements` honoured: `test_a_virtual_array_mapping_fills_unread_buffers_lazily`). The two inconsistencies you found are gone: `ak_add_doc` reaches the data (`test_dict_ak_add_doc_and_typenames_reach_the_mapping`), a count-only analysis reports its counter branch (`test_count_only_output_declares_the_counter_branch`). |
| 3 | Trim the tests | Done. 20 files → 5 (`read`, `form_mapping`, `projection`, `write`, `partitions`), 54 tests, ~5 s wall locally on thread/sequential executors, one process-pool smoke. Tests of graphed's own machinery (preserve, report, executor) are dropped. |

**Round 2 — inline**

| Item | Result |
|---|---|
| W1 bare expression writes inputs | Refused with `TypeError` before anything is opened (`test_derived_record_is_written_and_a_fieldless_expression_is_refused`); the test that pinned the fallback is gone. |
| W2 RNTuple instead of TTree | Writes go through `_dask_write.ak_to_root`, so `mktree` → TTree (`test_writes_one_ttree_file_per_partition`). No RNTuple opt-in added; `dask_write` has none either. |
| W3 compression `KeyError` | Shared with `ak_to_root`: `"ZLIB"` accepted, unknown names get its message (`test_compression_names_match_dask_write`). |
| W4 executor `KeyError` / instance | `ValueError` listing the accepted names; an instance is used as-is, a class is instantiated with `max_workers` (`test_executor_is_a_name_an_instance_or_a_class`). |
| W5 missing `dask_write` kwargs | All added: `storage_options` + remote destinations through `url_to_fs` (`test_remote_destinations_go_through_fsspec`), `title`, `field_name`, `counter_name`, `initial_basket_capacity`, `resize_factor` (`test_tree_options_reach_mktree`). |
| G1 dask kwargs swallowed | `step_size`, `steps_per_file`, `open_files`, `allow_read_errors_with_report` are rejected by name with a pointer to where partitioning lives (`test_dask_only_keywords_are_refused_by_name`). Partly refuted: `decompression_executor` / `interpretation_executor` were never dropped — `**options` went to `uproot.open`, which applies them to reads. They are explicit parameters now, as in `uproot.dask` (`test_decompression_executor_reaches_every_read`). |
| G2 experimental note | Done (see overall 1). |
| G3 `steps_allowed=False` | `regularize_files(steps_allowed=True)`; explicit steps become the partitions (`test_explicit_steps_are_the_partitions`). `graphed_write` refuses explicit steps (it derives the part index from a blind partition) and says to use `steps_per_file`. |
| G4 TBranch path, AsGrouped | Resolution is `_dask.py`'s own (`_resolve_trees_and_keys`, `_normalize_grouped_keys`), used by both dask array paths and by `uproot.graphed` (`test_a_leaf_branch_as_object_path_selects_that_branch`; AsGrouped covered by test_1642 on the shared code). |
| G5 no object path fails at read | Partitions carry the resolved `object_path` (`test_a_single_tree_file_needs_no_object_path`). |
| G6 `custom_classes` / `allow_missing` | Workers open through `object_in_directory` (the body of `regularize_object_path`), so both reach every read; a missing tree yields an empty chunk / no part (`test_known_base_form_with_allow_missing_reads_a_missing_tree_as_empty`, `test_a_missing_tree_under_allow_missing_writes_no_part`). Partly refuted: at construction `allow_missing` already skipped the file; only the `known_base_form` path was affected. |
| G7 one read path | Done (overall 2). |
| G8 copy of `read_tree` | `UprootReadMixin.read_tree` and the graphed source both call `_read_tree`. |
| G9 `columns` / `last_columns_read` | Removed. |
| G10 `str(need)` | Enum identity. |
| G11 stale blind docstring | Rewritten. |
| G12 legacy sentinel, `getattr(..., "is_blind")` | Both gone. |
| I1 eight exports | Three (overall 1). |
| E1 `AttributeError` version detection | Explicit `parse_version` floor (`test_a_too_old_graphed_reports_the_floor`). |
| P1 `test` group | New `test-graphed` group and a single `graphed-build` job. |
| T1 blind_steps cost | Range arithmetic asserted directly, no executor (`test_steps_per_file_splits_into_contiguous_ranges`, `test_step_size_caps_each_chunk`); one process-pool test remains. |
| T2 preserve/report/executor tests | Deleted, with `tests/graphed/report.py`. |
| T3 pinned fallback | Gone with W1. |
| T4 `tests/graphed` shadowing | Renamed `tests/graphed_helpers`; consolidated into the five files. |

**Round 3**

| Item | Result |
|---|---|
| 1 nested RNTuple schema | Fixed: base form from `to_akform(filter_name=common_keys)`, shared with `_dask.py` (`test_a_nested_rntuple_keeps_its_nesting`: form `{rec: {x, y}, z}`, `g.rec.x` records). Reading a nested field still fails, and it fails identically in `uproot.dask` — `uproot.dask(f).rec.x.compute()` raises `ValueError: zip() argument 2 is longer than argument 1` from the RNTuple top-level-fields workaround in `TrivialFormMappingInfo.load_buffers` (the `TODO` there). Top-level RNTuple fields read fine on both paths. That is dask's RNTuple gap, not introduced here. |
| 2 repeated inputs overwrite parts | Refused: `graphed.write.file_bases` raises `ValueError("duplicate input ...")` (graphed#38), surfaced before any file is opened (`test_refusals_happen_before_anything_is_written`). |
| 3 axis-0 semantics change with partition count | Refused: `graphed.refuse_chunk_partials` (graphed#38) rejects any graph whose reduction of the partitioned axis (slice, index, `axis=None` reduction) is an output or feeds another node, with `GraphedError: 'slice' reduces the partitioned axis ...`. The same guard now sits in graphed's own `to_parquet`, which had the same defect. Partition-aware global slicing is not implemented; the rejection is explicit rather than documented-only. |
| 4 `graphed_head` applies `n` first | Collects the first `n` output rows by growing a prefix read (`test_head_returns_the_first_rows_of_the_output`: `g.x[3:8]` → `[3, 4, 5]`, `g[g.x % 2 == 0]` head 2 → two rows). |
| 5 mapped projection under-supplies | Refuted. `zip({"jet": g.Jet.pt, "muon": g.Muon.pt})` is not a valid awkward expression: with both branches read, in eager awkward, `uproot.dask` and here, `ak.zip` raises "cannot broadcast nested list" (jets and muons have different per-event multiplicities). The same shape with `depth_limit=1` and `[["jet"]]` projects to `{nJet, Jet_pt, nMuon}` and evaluates: the replayed `Muon.pt` node runs on the mapping's placeholder buffers and is then dropped (`test_unread_branches_are_placeholders_the_graph_can_pass_through`, `test_behavior_properties_project_to_their_branches_and_evaluate`). |
| 6 process writer cannot ship behaviors | The worker receives the backend as a class or `"module:attr"` import reference (the `aggregate_plan` seam), default = the session backend's class; a behavior travels by reference (`test_process_pool_with_a_behavior_backend_by_import_ref`, which also pins that a lambda-bearing instance still cannot pickle). |
| 7 externals dropped | `graphed_head` and `graphed_write` carry the session's evaluators (`test_externals_reach_the_write_workers`; head checked with `graphed.apply`). graphed's own writers had the same defect (graphed#38). |
| 8 open TTrees not partition-compatible | Normalised to `(file_path, object_path)`; workers reopen by path (`test_an_open_tree_is_reopened_by_path_in_the_workers`). |

Also run: uproot's offline suite and the dask tests against the shared helpers (all green locally).

src/uproot/writing/_graphed_write.py
lgray: :robot: _AI text below_ :robot:

Fixed (W1). A fieldless result is refused with `TypeError` before any file is opened, telling the user to wrap it (`gak.zip({"x2": g.x * 2})`); nothing substitutes the inputs any more. Witness: `test_derived_record_is_written_and_a_fieldless_expression_is_refused`. The test that pinned the fallback is gone with the old suite.

src/uproot/writing/_graphed_write.py
lgray: :robot: _AI text below_ :robot:

Fixed (W2). Each partition is written by `_dask_write.ak_to_root`, so the writer is `mktree` and the output is a `TTree` under `tree_name`, as `dask_write` does (`test_writes_one_ttree_file_per_partition`). I did not add an RNTuple opt-in: `dask_write` has none either, so there is no parity to keep and nobody has asked for it yet.

src/uproot/writing/_graphed_write.py
lgray: :robot: _AI text below_ :robot:

Fixed (W3). Compression goes through the same `ak_to_root` lookup: `"ZLIB"` is accepted and an unknown name gets its error message (`test_compression_names_match_dask_write`).

src/uproot/writing/_graphed_write.py
lgray: :robot: _AI text below_ :robot:

Fixed (W4). `executor=` takes `"process"`, `"thread"` or `"sequential"` (unknown string → `ValueError` listing them), an executor *instance* used as-is, or an executor *class* instantiated with `max_workers` (`test_executor_is_a_name_an_instance_or_a_class`).

src/uproot/writing/_graphed_write.py
lgray: :robot: _AI text below_ :robot:

Fixed (W5), not deferred. `graphed_write` now takes `title`, `field_name`, `counter_name`, `initial_basket_capacity`, `resize_factor`, `compression_level` and `storage_options`, all forwarded to `ak_to_root`; the destination is created and addressed through `fsspec.core.url_to_fs`, so remote destinations work and return full URLs while local ones return plain paths (`test_tree_options_reach_mktree`, `test_remote_destinations_go_through_fsspec`).

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Fixed for the four partitioning/reporting kwargs (G1): `step_size`, `steps_per_file`, `open_files` and `allow_read_errors_with_report` are rejected by name with a pointer to where that lives in graphed (`test_dask_only_keywords_are_refused_by_name`).

Partly refuted for the two executors: `decompression_executor` / `interpretation_executor` were never dropped — `**options` reached `uproot.open`, which applies them to every read from that file. They are explicit parameters now, mirroring `uproot.dask` (`test_decompression_executor_reaches_every_read`).

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Done (G2). The module docstring and both `uproot.graphed` and `uproot.graphed_write` docstrings carry an **Experimental** note: the API tracks a pre-1.0 `graphed` and may change without a deprecation cycle.

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Fixed (G3). `regularize_files(..., steps_allowed=True)`; explicit `steps` become the partitions when reading (`test_explicit_steps_are_the_partitions`). `graphed_write` refuses explicit steps with a `TypeError` naming `steps_per_file` instead, because a write worker derives its part index from a blind partition alone.

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Fixed by sharing (G4). The key collection is now `_dask.py`'s own: `_resolve_trees_and_keys` (TBranch object paths via `is_self`) and `_normalize_grouped_keys` (the `AsGrouped` parent/child normalisation), used by both `uproot.dask` array paths and by `uproot.graphed`. Witness for the TBranch path: `test_a_leaf_branch_as_object_path_selects_that_branch`; the `AsGrouped` case is covered by `test_1642_read_AsGrouped_with_dask_or_virtual_array.py` on the shared code (133 dask tests pass).

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Fixed (G5). Partitions carry the *resolved* `(file_path, object_path)` — `TTree.object_path`, e.g. `/events;1` — so a single-TTree file with no object path reads and writes (`test_a_single_tree_file_needs_no_object_path`).

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Fixed (G6). Workers open a partition's tree through `uproot._util.object_in_directory`, which is now the body of `regularize_object_path`, so `custom_classes` and `allow_missing` reach every read; a missing tree yields an empty chunk / no output part (`test_known_base_form_with_allow_missing_reads_a_missing_tree_as_empty`, `test_a_missing_tree_under_allow_missing_writes_no_part`).

One correction to the report: at construction time `allow_missing=True` already skipped the file (the `__call__` path did not reach `None.num_entries`); only the `known_base_form` path, which opens nothing up front, was affected.

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Done (G7). There is one read path: every read goes through a form mapping, `TrivialFormMapping` by default. `_evaluation_columns`, `_task_columns`, both `_read_columns` variants and the subclass split are gone. Both inconsistencies you found are closed: `ak_add_doc` reaches the data (`test_dict_ak_add_doc_and_typenames_reach_the_mapping`), and a count-only analysis reports its counter branch (`test_count_only_output_declares_the_counter_branch`).

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Done (G8). `read_tree` is now a module-level `_read_tree` in `_dask.py`; `UprootReadMixin.read_tree` and the graphed source both call it, so `buffer_replacements` (and `FormMappingInfoWithVirtualArrays`) is honoured on the graphed path (`test_a_virtual_array_mapping_fills_unread_buffers_lazily`).

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Done (G9). `columns` and `last_columns_read` are removed; the surviving test spies on the read instead.

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Done (G10). Enum identity (`need is BufferNeed.DATA`), given the floor.

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Done (G11). The docstring describes `Partition.blind` / `resolve`; the `entry_stop=-n_steps` wording is gone.

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:

Done (G12). The legacy-sentinel branch and the `getattr(partition, "is_blind", False)` are gone; blind partitions are the only encoding and `partition.is_blind` is a plain attribute.

src/uproot/__init__.py
lgray: :robot: _AI text below_ :robot:

Done (I1). Top-level exports are `uproot.graphed`, `uproot.graphed_head` and `uproot.graphed_write` (the last via `uproot.writing`, like `dask_write`). `necessary_buffers` and `resolve_read_branches` are deleted outright — nothing reached their OFFSETS→counter trick through the mapped read path. `necessary_columns`, `graphed_partitions` (returns `graphed.core.Partition`s for `aggregate_plan(partitions=...)`) and `read_graphed_partition` stay importable from `uproot._graphed`.

src/uproot/extras.py
lgray: :robot: _AI text below_ :robot:

Done (E1). `extras.graphed()` does an explicit `parse_version` check against `_GRAPHED_MIN = "0.0.4"`, as the other extras do; a genuine import error propagates (`test_a_too_old_graphed_reports_the_floor`).

pyproject.toml
lgray: :robot: _AI text below_ :robot:

Done (P1). The graphed test dependencies live in a `test-graphed` group (`graphed[awkward,parquet]>=0.0.4`, `graphed-executors>=0.0.3`, `vector`, all `python_version >= '3.11'`), and a single `graphed-build` job installs it. That job is red on this head only because the 0.0.4 floor is not on PyPI yet; graphed 0.0.4 (graphed-org/graphed#39) is being released now, after which the job gets re-run.

tests/test_1720_graphed_blind_steps.py
lgray: :robot: _AI text below_ :robot:

Done (T1). The range arithmetic is asserted directly on the returned partitions, no executor (`test_steps_per_file_splits_into_contiguous_ranges`, `test_step_size_caps_each_chunk`). Everything else runs on the thread or sequential executor; one process-pool smoke test remains. The five files run in ~5 s wall locally.

tests/test_1720_graphed_preserve.py
lgray: :robot: _AI text below_ :robot:

Done (T2). `test_1720_graphed_preserve.py`, `test_1720_graphed_report.py`, `test_1720_graphed_executor.py` and `tests/graphed/report.py` are deleted; graphed's own machinery is tested in graphed.

tests/test_1720_graphed_projection_and_compiled_reads.py
lgray: :robot: _AI text below_ :robot:

Done (T3). This test is gone; the behaviour it pinned is now a `TypeError` (see the `_graphed_write.py` thread).

tests/graphed_helpers/__init__.py
lgray: :robot: _AI text below_ :robot:

Done (T4). The package is `tests/graphed_helpers` and the twenty files are consolidated into `test_1720_graphed_read.py`, `_form_mapping.py`, `_projection.py`, `_write.py` and `_partitions.py` (54 tests).

lgray:

> **Mark it experimental.** … keep the public surface to those two names (+ maybe `graphed_head`) until it settles.

Done. Experimental notes on the module and on `uproot.graphed` / `uproot.graphed_write`; the exported names are `uproot.graphed`, `uproot.graphed_head`, `uproot.graphed_write`. The rest is importable from `uproot._graphed` only (details in the `__init__.py` thread).

lgray:

> **Collapse the two read paths into one** by defaulting to `TrivialFormMapping`, as `uproot.dask` does.

Done, and further: reads share `_dask.py`'s resolver, base-form construction and `_read_tree` with `uproot.dask`, so the two front ends cannot drift apart again (details in the `_graphed.py` L118 / L132 threads).

lgray:

> **Trim the tests.** … Almost half of that is four files that exercise only `graphed_partitions` / `read_graphed_partition`.

Done. 20 files → 5 (read / form_mapping / projection / write / partitions), 54 tests, ~5 s wall locally; thread and sequential executors plus one process-pool smoke; graphed-machinery tests dropped; the graphed dependencies moved to a `test-graphed` group with one CI job.

lgray:

> **P1: nested RNTuples get the wrong schema** … Please use `to_akform(filter_name=common_keys)` as `_dask.py:1767-1769` does.

Fixed as suggested: the base form comes from `to_akform(filter_name=common_keys)` through the helper `uproot.dask` uses, so `{rec: {x, y}, z}` keeps its nesting and `g.rec.x` records (`test_a_nested_rntuple_keeps_its_nesting`).

One thing this does not fix, because it is not this PR's: *reading* a nested RNTuple field still fails, identically on `uproot.dask` — `uproot.dask(f).rec.x.compute()` raises `ValueError: zip() argument 2 is longer than argument 1` from the RNTuple top-level-fields workaround in `TrivialFormMappingInfo.load_buffers` (the `TODO` there). Top-level RNTuple fields read fine on both paths.

lgray:

> **P1: repeated inputs overwrite output parts** … The occurrence needs to be part of the partition identity/index, or duplicate inputs need to be rejected.

Rejected explicitly. `graphed.write.file_bases` (graphed-org/graphed#38) raises `ValueError("duplicate input …: list each file (or file, tree) once")`, and `graphed_write` surfaces it before any file is opened (`test_refusals_happen_before_anything_is_written`).

lgray:

> **P1: writes silently change axis-0 semantics with partition count** … Either global operations need partition-aware handling or this path must reject/document non-row-local graphs.

Rejected explicitly, at the cause. A partition-wise driver evaluates the compiled IR once per chunk, so any reduction of the partitioned axis (`x[2:8]`, `x[0]`, `gak.sum(x)`) is a per-chunk partial. `graphed.refuse_chunk_partials` (graphed-org/graphed#38) refuses one that is a write output or that feeds another node, with `GraphedError: 'slice' reduces the partitioned axis: a partitioned write has no combine step …`; `graphed_write` calls it before opening anything (`test_refusals_happen_before_anything_is_written`). graphed's own `to_parquet` writers had the same defect and carry the same guard now. Dataset-global slicing is not emulated.

lgray:

> **P1: `graphed_head` applies `n` before the expression** … A dask-like head must collect the first `n` output rows.

Fixed. `graphed_head` collects the first `n` rows of the *output*, growing a prefix read until the evaluated result has `n` rows or the first tree is exhausted: `graphed_head(g.x[3:8], 3) == [3, 4, 5]`, `graphed_head(g[g.x % 2 == 0], 2)` returns two rows (`test_head_returns_the_first_rows_of_the_output`).

lgray:

> **P1: mapped projection can under-supply compiled evaluation** … With `zip({"jet": g.Jet.pt, "muon": g.Muon.pt})[["jet"]]`, projection omits `Muon_pt` while IR still evaluates that node; both `graphed_head` and `graphed_write` fail.

Refuted. `zip({"jet": g.Jet.pt, "muon": g.Muon.pt})` is not a valid awkward expression: with *both* branches read — eagerly, through `uproot.dask`, or here — `ak.zip` raises `cannot broadcast nested list` because jets and muons have different per-event multiplicities. The failure you saw is that, not under-supply.

The well-formed version of the same shape, `zip({...}, depth_limit=1)[["jet"]]`, projects to `{nJet, Jet_pt, nMuon}` and evaluates on both `graphed_head` and `graphed_write`: the replayed `Muon.pt` node runs on the mapping's placeholder buffers (offsets read, data not) and is then dropped (`test_unread_branches_are_placeholders_the_graph_can_pass_through`, `test_behavior_properties_project_to_their_branches_and_evaluate`).

lgray:

> **P2: the default process writer cannot ship common behaviors** … ROOT writing needs the same seam or should avoid serializing the backend instance.

Fixed with the same seam as the Parquet writer. The worker receives the backend as a class or a `"module:attr"` import reference (the `aggregate_plan` convention), default = the session backend's class; a behavior travels by reference (`test_process_pool_with_a_behavior_backend_by_import_ref`, which also pins that a lambda-bearing backend *instance* still cannot pickle — that is inherent, and no longer the default).

lgray:

> **P2: external evaluators are dropped by `graphed_head` and `graphed_write`** … `aggregate_plan` already wires these through `_external_evaluators`.

Fixed. Both carry the session's evaluators, through the now-public `graphed.aggregate.external_evaluators` (graphed-org/graphed#38, which also wired them into graphed's own writers, which had the same gap). `test_externals_reach_the_write_workers`; `graphed_head(graphed.apply(f, g.px1), 2)` evaluates.

lgray:

> **P2: already-open TTrees are not partition-compatible** … Normalize it to a reusable file/tree reference or special-case open objects in `open_tree`.

Fixed by normalising: an open `TTree` becomes its `(file_path, object_path)` pair and workers reopen it by path (`test_an_open_tree_is_reopened_by_path_in_the_workers`).

lgray:
@ariostas OK - bot responses done and everything is addressed. Re-review or let me know what to do otherwise! Thanks!

lgray:
@ariostas I decided to fix up the RNtuple reading problem for dask/graphed in this PR as well. Review when ready!

src/uproot/_graphed.py
```diff
@@ -0,0 +1,564 @@
...
+
+
+def read_graphed_partition(
+    partition, columns, *, tree=None, library="ak", **open_options
+):
+    """Read a ``graphed.core.Partition``'s chunk of ``columns`` from its ROOT file, resolving a
+    **blind** partition against the file's actual entry count. Pass an already-open ``tree`` to
+    reuse a per-worker ``open_once`` handle."""
+    if tree is None:
+        tree = uproot.open(partition.uri, **open_options)[partition.tree]
```
ariostas: :robot: _AI text below_ :robot:

A blind partition from `graphed_partitions(open_files=False)` over a file given *without* an object path carries `tree=""`, and `uproot.open(uri)[""]` raises `KeyInFileError` here (`open_files=True` is fine, since it stores the resolved `obj.object_path`). `open_tree` already handles this with `object_path or None`; this helper could go through `uproot._util.regularize_object_path(partition.uri, partition.tree or None, ...)` for the same single-TTree fallback. Low priority — it's a private helper.

src/uproot/writing/_graphed_write.py
```diff
@@ -0,0 +1,223 @@
...
+    if isinstance(executor, str):
+        graphed_executors = uproot.extras.graphed_executors()
+
+        classes = {
+            "process": graphed_executors.local.ProcessPoolExecutor,
+            "thread": graphed_executors.local.ThreadExecutor,
+        }
+        if executor not in classes:
+            raise ValueError(
+                f"executor must be 'process' or 'thread', an executor instance, or an executor "
```
ariostas, line 94: :robot: _AI text below_ :robot:

The reply on the earlier thread says `executor="sequential"` is accepted, but the code and docstring only take `"process"` / `"thread"`. Just confirming the docstring is the intended truth (which seems fine — a sequential run is one `max_workers=1` thread executor away).

ariostas (commented):

Second pass at `3b5d6d7`. Everything from the first round is addressed — I re-ran each probe against this head rather than going by the replies — and the shared-code refactor holds up: the full suite passes locally on this branch (1149 passed, network-free), the graphed suite is 62 tests in ~8 s wall at ~93 % CPU, and all CI jobs including the new `graphed-build` are green.

On the RNTuple fix: I diffed `keys()` / `filter_name` results between `main` and this branch on the RNTuple files in `skhep_testdata`; the only change is that exact dotted keys under a collection now match (e.g. `_collection0.Muon_pt` went from `[]` to `['_collection0.Muon_pt']`), so it's a strict improvement and fine to keep here.

Two small nits inline; nothing blocking.

lgray:
@ariostas lemme have the bot address the nits and then I think we can merge!

src/uproot/_graphed.py
lgray: :robot: _AI text below_ :robot:
Fixed in 11185f5: `read_graphed_partition` now resolves the tree with `uproot._util.object_in_directory(uproot.open(partition.uri, **open_options), partition.tree or None, False)` — the same fallback `open_tree` already uses, so a blind partition with `tree=""` lands on the single TTree. I grepped `src/uproot` for every place a partition's tree indexes an opened file and found two members: `open_tree` (already correct, untouched) and this one. Added `tests/test_1720_graphed_partitions.py::test_a_blind_partition_without_an_object_path_reads_the_single_tree`, which fails on the pre-fix source with `KeyInFileError: not found: ''` and passes after.

src/uproot/writing/_graphed_write.py
lgray: :robot: _AI text below_ :robot:
The docstring is the truth — my earlier reply overstated it; `executor=` accepts only `"process"` and `"thread"`, and the `ValueError` already lists exactly those. No `"sequential"` string was added: `graphed_executors.local` has no sequential/synchronous executor class to map it onto, so mapping it would mean inventing a feature. A sequential run is available today by passing an instance instead, e.g. `uproot.graphed_write(g, out, executor=graphed.core.SequentialRunner())`, which is what I should have said the first time.

ariostas:
Thank you @lgray. I think this is good to go. For now, I think it's fine to have a separate CI job for graphed, but I think later on we should make it part of the standard test dependencies and have it tested on all platforms. Let me know if it's also ready on your side.

I'll aim to have a new uproot release this week so that people can start using this.

ariostas (approved):

lgray:
Yep - sounds good to me! Thanks!

Issue #1721, opened 2026-09-14
Support RNTuple attributes


ariostas:
The RNTuple spec added "attributes" which are basically user-defined metadata. We should implement proper support to read them. It is still to be determined how we should use this data.

PR #1725, opened 2026-09-21
feat: add support for reading linked RNTuple attribute sets

aashirvad08:
This PR adds support for reading linked attribute sets from RNTuples

RNTuple v1.0.1.0 introduced attribute sets as separate linked RNTuples. Uproot was already reading the corresponding footer records but it was not  following the locators to actually load the attribute data

And PR exposes them through RNTuple.attributes and adds RNTupleAttributeSet for accessing the user data, ranges and entry-based lookups

It also handles linked RNTuple cache isolation and lazy loading so unsupported attribute schemas only fail when that particular attribute set is accessed 

limitation: test_ROOT has not been run because it requires ROOT 6.40 which introduced CreateAttributeSet

closes - #1721


src/uproot/models/RNTuple.py
```diff
@@ -2061,6 +2085,281 @@ def array(
         return arrays
 
 
+class RNTupleAttributeSet:
```
ariostas: Maybe it would be better to have this be a subclass of `HasFields`.

